PR #3, opened 2021-02-11
Monitoring System End To End Tests

mig4:
This is an RFC for some end-to-end testing ideas for the monitoring
system we're building that we talked about with @TheoBrigitte and
@QuentinBisson.


mig4:
Let's get this ball rolling. Opinions, critiques, and ideas welcome.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
```diff
@@ -0,0 +1,79 @@
...
+    Prometheus instances; this is to catch issues with configuration we
+    generate and catch regressions when monitored components change
+
+  - ensure that provisioned Prometheus instances ingest the metrics we need for
+    alerting; this is to catch issues with configuration and regressions when
+    monitored components change
+
+## Test Infrastructure
+
+Given the above goals, we'd ideally need a system that is able to provision
```
JosephSalisbury, line 56: Yeah, if this is a hard requirement, we're probably still a few months away from that :D - are there any smaller steps we can take that start us in that direction? For example, does only installing the operator on a small cluster start flushing out a lot of the Helm issues, for example?

QuentinBisson: We discussed it with @TheoBrigitte a little and we thought using "production" test installations like gaia and godsmack could be a good starting point

JosephSalisbury: Oh sure, we integrate w/ all the fancy tekton stuff. I hadn't thought about that, nice!

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
```diff
@@ -0,0 +1,79 @@
...
+    correctly; this is to catch issues with configuration for Prometheus that
+    we generate
+
+3. Monitoring system operation:
+
+  - ensure that all built-in/configured targets are configured in provisioned
+    Prometheus instances; this is to catch issues with configuration we
+    generate and catch regressions when monitored components change
+
+  - ensure that provisioned Prometheus instances ingest the metrics we need for
```
JosephSalisbury, line 50: How do you think we could determine differences between changes introduced by Atlas, and changes introduced by other teams? Does this interact w/ platform releases somehow?

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
JosephSalisbury: Okay, given that, it'd be cool to see some smaller initial steps (e.g: 80/20 rule) - what's the first small thing we can do?

QuentinBisson: TBH, prometheus-meta-operator is an app so I think we should check with ManagedApp teams what is their solution to test app across the different providers, including how they do e2e tests and maybe use what they use?

QuentinBisson: If we only want the easy solution, we can just deploy the app across all 3 providers and then add tests over time

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
```diff
@@ -0,0 +1,79 @@
...
+## Open Questions
+
+1. Testing everything against all supported providers may not be necessary,
+   could we split the tests into provider-independent tests which could even be
+   run on a light-weight infra (like KinD or similar) and provider-dependent
+   ones that need real infra?
+
+2. I think we can leave the question of triggering the tests for later once we
+   have something more concrete, for now even being able to trigger such tests
+   manually before a release would be OK. Right?
```
TheoBrigitte, line 70: Yes that's right, let's not start with everything. Having anyway to run a test for now would be better than what we have (no tests), let's think about how we trigger that later on.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
```diff
@@ -0,0 +1,79 @@
...
+clean _Management Cluster_ environments into which we could deploy the system
+and which is then able to provision specifically configured
+_Workload Clusters_ to test the monitoring system operation.
+
+## Open Questions
+
+1. Testing everything against all supported providers may not be necessary,
+   could we split the tests into provider-independent tests which could even be
+   run on a light-weight infra (like KinD or similar) and provider-dependent
+   ones that need real infra?
```
TheoBrigitte, line 66: Yes and no, things like `1. Deployment` and `2. Provisioning` could be tested on light-weight infra, but my gut feelings says let's test things on real infra because there are always funny cases that we might forgot to replicate on the light-weight infra.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
TheoBrigitte: Yes we could test on small cluster, but is anything preventing us to test on testing installations ? My thinking is the closer we test on things that match production environment the better.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
mig4: Yeah so eventually it'd be good to separate out the infrastructure bits (PMO, promxy, etc.) and mechanisms we provide from the rules. I don't know if there are any rules/targets we'd want to have "built-in", but to me the line of responsibility would be there.

The testing of ingested metrics is something to validate alerts, so I'd say this would just be a test useful for other teams to know that they're not introducing a regression to their own alerting, etc.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
mig4: > TBH, prometheus-meta-operator is an app so I think we should check with ManagedApp teams what is their solution to test app across the different providers, including how they do e2e tests and maybe use what they use?

I like that. For deployability let's not re-invent stuff, other teams would have the same issues with their components.

TheoBrigitte (approved):
Fine with me overall

anvddriesch (approved):
Not as deep into this topic as atlas but sounds very good.

0003-monitoring-system-end-to-end-tests/0003-monitoring-system-end-to-end-tests.md
JosephSalisbury: yo @cokiengchiara / @piontec  do y'all have any thoughts on testing apps like this?

piontec: We're not there yet - I mean testing across different providers. We have this on our map, and I hope we can do it soon, but so far we only have a sketch of the test framework we might be using for that. If you're interested, have a look here https://github.com/giantswarm/app-build-suite/tree/master/examples/apps/hello-world-app/tests/abs

The end goal is to have something like that: https://docs.google.com/presentation/d/1EP1TweMtMg-YkTCSgKUU5ec6yP2L2ZteaqD7WC24WH8/edit#slide=id.gafe28231d1_0_165

teemow:
Please make sure to pull the changes before making new modifications. I've renamed the folder and markdown as we agreed on a slightly different structure for RFCs.

tuommaki:
Overall this document seems like a good start. Considering how many steps are involved in the monitoring pipeline e2e, would it make sense to have a some kind of table that lays out different parts of it and draws some kind of intermediate targets for test targets?

- Install all monitoring components and ensure they work together.
- Ensure that PMO generates derived configuration resources correctly from a "trigger" CR (in most cases this is `Cluster` CR, rigth?).
- Ensure that metrics are ingested correctly.
  - Which metrics?
  - What's the criteria for correctness?
 - Ensure that alerting by metrics works.
   - How to ensure that test mechanism itself works reliably?
- What about OpsGenie part - is it included?

JosephSalisbury:
@mig4 i reckon we have enough consensus here - why don't we merge, and start working on roadmap issues for implementation?

whites11:
What should we do about this PR? I get pinged by it every day :blush:

mig4:
I'll go ahead and merge this, we'll definitely need to add details once we take that into backlog, split the tasks and refine them.

Thanks for the feedback and smashing ideas everyone! That'll keep us busy for a while 😉

PR #4, opened 2021-02-15
Adds RFC 0004 - RFCs over ADRs and PDRs

JosephSalisbury:
Towards https://github.com/giantswarm/roadmap/issues/260

### DIFF
--- a/0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
+++ b/0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
@@ -1,0 +1,9 @@
+# RFC 0004 - RFCs over ADRs and PDRs
+
+This RFC describes the relationship between RFCs, ADRs, and PDRs.
+
+'ADR' refers to Architecture Decision Record, and 'PDR' refers to Product Decision Record. These records, as well as their processes, are currently located in the private Giant Swarm Intranet.
+
+ADRs and PDRs should be superseded by RFCs. This allows for one public process for discussing and recording ideas.
+
+The current ADRs and PDRs should be left as is. If they need to be reconsidered or extended, they can be brought up as new RFCs.


0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
```diff
@@ -0,0 +1,9 @@
+# RFC 0004 - RFCs over ADRs and PDRs
+
+This RFC describes the relationship between RFCs, ADRs, and PDRs.
+
+'ADR' refers to Architecture Decision Record, and 'PDR' refers to Product Decision Record. These records, as well as their processes, are currently located in the private Giant Swarm Intranet.
+
+ADRs and PDRs should be superseded by RFCs. This allows for one public process for discussing and recording ideas.
```
cokiengchiara, line 7:
```suggestion
ADRs and PDRs should be replaced by RFCs. This allows for one public process for discussing and recording ideas.
```

I had to look up what "superseded" means... 🙈

0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
```diff
@@ -0,0 +1,9 @@
...
+
```
snizhana-dynnyk, line 8: not sure about this. IMO ADRs/PDRs and RFCs serve different purposes. IF ADRs/PDRs are the records about decisions we made and committed to, RFCs can contain any ideas and there is no guarantee (and also no process atm) that they will become reality at some point.

I see RFCs as a 'start of conversation', to ask for input, to get feedback. 
If am wrong and don’t understand the meaning of RFCs then we need to establish a process for RFC lifecycle. Because getting few approvals is definitely not enough to take a decision.

0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
cokiengchiara: Hm. But PDRs and ADRs merely document that a certain decision was made and why. Doesn't the RFC complement the ADR and PDR, but not replace them? i.e.

1. Open an RFC
2. People comment
3. We make a decision
4. Write PDR or ADR

snizhana-dynnyk: that is my understanding as well

puja108: I like having the decisions public, but I would not like to change the format and mush them all together. The structure of the DRs serves its purpose and the non-structure of RFCs serves another purpose. RFCs might not even necessarily develop into DRs IMHO as we are not documenting every decision there is in architecture and product.

0004-rfcs-over-adr-pdr/0004-rfcs-over-adr-pdr.md
```diff
@@ -0,0 +1,9 @@
+# RFC 0004 - RFCs over ADRs and PDRs
+
+This RFC describes the relationship between RFCs, ADRs, and PDRs.
+
+'ADR' refers to Architecture Decision Record, and 'PDR' refers to Product Decision Record. These records, as well as their processes, are currently located in the private Giant Swarm Intranet.
```
anvddriesch, line 5: We explain what ADR and PDR stands for so I'd suggest to also explain what RFC is.

JosephSalisbury:
okay, i don't think people generally like this. let's see how we feel in the future

PR #5, opened 2021-02-15
Manage apps and configuration on third-party clusters

teemow:


JosephSalisbury:
Can you move this to RFC 0005 or 0006? It's clashing with the earlier https://github.com/giantswarm/rfc/pull/4

teemow:
I knew the numbers are a bad idea. I'll vote to remove the numbers. They do not make any sense to me. Things will get merged in a different order. Even some PRs will not even make it.

teemow:
Added #13 to remove the numbering

PR #7, opened 2021-02-15
Adopt existing clusters for KaaS management

teemow:


0006-adopt-existing-clusters-for-kaas-management/0006-adopt-existing-clusters-for-kaas-management.md
```diff
@@ -0,0 +1,7 @@
+# Adopt existing clusters for KaaS management
+
+As a user I can adopt a Kubernetes cluster and have it managed by Giant Swarm 24/7. Clusters can get upgraded to a full Giant Swarm release.
```
fiunchinho, line 3: Is there value in adopting a cluster even though is not using a full Giant Swarm release? Should we split those two things in different RFCs?

0006-adopt-existing-clusters-for-kaas-management/0006-adopt-existing-clusters-for-kaas-management.md
```diff
@@ -0,0 +1,7 @@
+# Adopt existing clusters for KaaS management
```
fiunchinho, line 1: [There is an issue in CAPZ upstream](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1173) to track the development for adopting `AKS` cluster specifically. Maybe we can link it to add context.

0006-adopt-existing-clusters-for-kaas-management/0006-adopt-existing-clusters-for-kaas-management.md
teemow: This is about adopting clusters that aren't created by us. What else would you want to split into a separate RFC? FYI there is also #5.

fiunchinho: One RFC for adopting clusters. Another one to upgrade an adopted clusters to a full Giant Swarm release.

teemow: I didn't see the need for that yet. Sure we can make this more granular. But even this one here is barely specified. So I'll keep this in mind in case it makes sense to separate this. The RFCs are not epics. I treated this just as a brainstorming document.

puja108: only possible split I could think of was splitting it into "adopting a managed cluster" (e.g. EKS/AKS) vs. "adopting a random cluster". Former is planned and partly implemented upstream in CAPI and also a bit more standardized, latter is much more complicated as the variety of cluster creation methods might create different migration problems. In the end, most probably it will anyway be limited to support certain setups in the beginning and could then be extended to support more.

That said, not seeing a big reason to split right now.

PR #9, opened 2021-02-15
Managed K8s in our KaaS product

teemow:


0008-managed-k8s-via-cluster-api/0008-managed-k8s-via-cluster-api.md
```diff
@@ -0,0 +1,7 @@
+# Managed K8s in our KaaS product
+
+As a User I would like to provision EKS or AKS clusters via Giant Swarm.
+
+Benefits: We use the building block we get from the cloud providers and save effort on our side. ClusterAPI allows this as well.
+
+Challenges: Do we also automate migration of clusters? Do we still offer a non-eks non-aks control plane?
```
puja108, line 7: We have tried this out for AKS (and also looked into the EKS implementation) with CAPZ and CAPA. A challenge IIRC @nprokopic mentioned was that currently a lot of the options one can set for AKS/EKS are not wired through, yet, but that we could work on.

0008-managed-k8s-via-cluster-api/0008-managed-k8s-via-cluster-api.md
```diff
@@ -0,0 +1,7 @@
+# Managed K8s in our KaaS product
+
+As a User I would like to provision EKS or AKS clusters via Giant Swarm.
+
+Benefits: We use the building block we get from the cloud providers and save effort on our side. ClusterAPI allows this as well.
```
puja108, line 5: Another benefit that we would get here with CAPI is that theoretically cluster adoption, i.e. taking over existing AKS/EKS clusters into our management, also should work.

0008-managed-k8s-via-cluster-api/0008-managed-k8s-via-cluster-api.md
nprokopic: > currently a lot of the options one can set for AKS/EKS are not wired through, yet, but that we could work on.

I can confirm this for AKS. Working on it to wire in more options that Azure provides should be relatively easy.

PR #11, opened 2021-02-15
Self Service

teemow:


0010-self-service/0010-self-service.md
```diff
@@ -0,0 +1,7 @@
+# Self service
+
+As a potential user I would like to create k8s clusters and/or my own control plane without having to talk to Giant Swarm.
+
+Benefits: Lower barrier to experience us
```
cokiengchiara, line 5: What is the benefit to the user of creating k8s clusters using Giant Swarm's product?

Without talking to us, what is the "us" that they are experiencing with this?

0010-self-service/0010-self-service.md
```diff
@@ -0,0 +1,7 @@
+# Self service
+
+As a potential user I would like to create k8s clusters and/or my own control plane without having to talk to Giant Swarm.
```
puja108, line 3: I was a bit confused first, cause we have also talked about widening our user-base within customers to have more people and especailly also end-user developers have access to GS as a product and interface, but this is for non-GS customers, right? So more of a trial or entry approach.

0010-self-service/0010-self-service.md
puja108: We might need to think about the value of the product alone here, but this should maybe also not be seen as the end state, but as a quicker entry, from where people can get in touch with us, but without having to do it before they've even tried the product. Assuming or asserting that our product alone does offer some kind of value by itself that would motivate to get in touch?

0010-self-service/0010-self-service.md
teemow: On a high level not necessarily. Could also be the engineering teams not having to talk to the platform team that has access to happa.

puja108: Sure, I understand where you are coming from on a high abstraction level. I would most probably however separate those two opportunities to give room to more ideas and solutions towards them. E.g. the solutions one could apply to the "engineering teams not having to talk to the platform team that has access to GS" opportunity do not necessarily also apply to or solve all the requirements of the wider non-GS opportunity (vice verse might apply though).

PR #12, opened 2021-02-15
Cloud Integration

teemow:


0011-cloud-integration/0011-cloud-integration.md
```diff
@@ -0,0 +1,10 @@
+# Cloud Integration
+
+As a user I would like to manage other cloud resources from my kubernetes clusters.
+
+Tools for that:
+ * AWS Kubernetes controllers (ACK) - https://github.com/aws-controllers-k8s/community
+ * Azure Service Operator - https://github.com/Azure/azure-service-operator
+ * Cross Plane - https://crossplane.io/
+
+Benefits: Manage a bigger landscape of cloud infrastructure
```
fiunchinho, line 10: What can GS do to bring value here other than offering the app in our catalog?

pipo02mix: IMO make easy to install (adding in the catalog) plus giving docs and guides to explain how this work. If we offer as managed app then we maintain the operator running too :)

PR #13, opened 2021-02-15
Remove rfc numbers

teemow:
Manually created numbers aren't really helpful. My very first RFC already collided. Things will get merged in a different order. Some PRs will not even make it. So we will end up with lots of holes in the numbering anyway. Let's try this without numbers. It makes creating RFCs even easier and this was the main goal.

### DIFF
--- a/0001-rfc-process/0001-rfc-process.md
+++ b/0001-rfc-process/0001-rfc-process.md
@@ -12,9 +12,8 @@ All ideas, thoughts, comments, suggestions, and eldritch ramblings, are valid an
 
 The following lists some basic formatting and housekeeping decisions concerning RFCs:
 
-- RFC numbers must be unique and increasing (i.e: take the next available number).
-- Each RFC must be in its own directory, with the directory name in the form `XXXX-$title`.
-- RFC filenames must be in the form `XXXX-$title.md`.
+- Each RFC must be in its own directory, with the directory name in the form `$title`.
+- RFC filenames must be in the form `$title`.
 - RFCs must be written in Markdown.
 
 The following lists some ideas towards RFC that are explicitly not presented:
--- a/rfc-process/rfc-process.md
+++ b/rfc-process/rfc-process.md
@@ -1,4 +1,4 @@
-# RFC 0001 - RFC Process
+# RFC Process
 
 This RFC describes the RFC process for Giant Swarm.
 


cokiengchiara (approved):
makes sense to me

puja108 (commented):
I don't care much about the numbers, I think in original RFC processes it is mainly for easier referencing of RFC between each other (similar to issues numbers cause titles can change), but we might not have that much use for it or maybe I don't see it and @JosephSalisbury does

teemow:
But atm we have both title and number in the folder name. So links will change either way. I'd love to have numbers in here like issues that can get easily referenced. But with a repo of markdown files this isn't as easy. I think this is why we've had the former roadmap rfcs in long running PRs. It is easier to work with the PR than the merged markdown :smile:.

webwurst:
Kubernetes KEP process [switched to this](https://github.com/kubernetes/enhancements/tree/master/keps#what-is-the-number-at-the-beginning-of-the-kep-name):
> KEPs are now prefixed with their associated tracking issue number. This gives both the KEP a unique identifier and provides an easy breadcrumb for people to find the issue where the current state of the KEP is being updated.

JosephSalisbury:
I think RFCs having an _identifier_ is useful. I kind of like the monotonically increasing IDs for that, but not married (i'm not too stressed about gaps).

@teemow do you still want numbers, but not to have them in the filename? i don't get that so much from the changes in the PR itself

teemow:
My points here:
We need to think about the whole company. There are people that don't use git on their laptops. And they shouldn't need to. To write an RFC. I've created an RFC that collides with the number. This is nearly impossible to change in the webinterface now. So imo this is a no go. So let's go with a title only for now. Imo this won't hurt us much atm. As said before links would contain title and number anyway and having folders with just numbers is also not usable at all.

teemow:
Let's align here before we have to change a whole lot of RFCs :smile:

marians:
Each RFC needs a unique identifier that is easy to use.

In our internal roadmap repo, [two to three years back](https://github.com/giantswarm/giantswarm/tree/6c601bec2d473054fb680ce76f99085b76de175d/roadmap), we had a convention where each document (one markdown file in that case) had to have a unique name that also represented a speaking working title. Examples:

image

That wasn't bad. However it also wasn't ideal, because there is a lot in a name, and if the name (working title) does not exactly fit the spec 100%, it can already cause misinterpretations. As an example: when reading `EVENT-STREAM`, many will immediately think about some streaming API, websockets, etc., while that wasn't necessarily the point of the spec.

So I would not vote for "speaking identifiers any more". Which leaves us with systematic, non speaking titles.

### Date stamps

**Use a format `YYYY-MM-DD`, representing the day when the RFC actually gets merged.**

Resolving conflicts will be done before merging. In the unlikely case that there are several RFCs to be merged on the same day, one could for example simply wait for another day. Or we start appending letters, like `2021-04-21a`.

The beautiful side effect: There is a chronological order when browsing RFCs in the repo. And we automatically know when the RFC has been published.

### Numbers

Numbers are fine, IMO. I don't mind the gaps that will occur. I think identifier conflicts can be addressed before merging. When opening a new PR, it's not even necessary to know the final number.

teemow:
I disagree. Mainly RFCs should be easy to add. So far there is no way to add automatic numbering and my RFCs are already blocked since weeks because of a collision and this discussion. Changing a folder name is impossible from the webinterface which means non-tech people can't work with this. Using dates is highly impractical and imo also rather unfortunate. There is enough metadata to find out the date of creation or the latest change. No need to put this into the naming.

I'm definitely against numbers and even more against dates. I'd be fine if we find another workflow to use issues or PRs and their numbers. But let's not create an artifical numbering system.

teemow:
Anyway this seems to be ignored widely. So to me the whole RFC setup is broken and blocked and doesn't have priority enough to get fixed. Shall we rethink RFCs as a whole? Maybe this was a nice try but will not find wide spread adoption anyway. WDYT @JosephSalisbury?

marians:
I do not agree with the conclusion that RFCs are broken and need rethinking in general, just because of different opinions regarding the identifier. Just my 2ct.

teemow:
They are because the workflow so far is not accessible for non-tech people in the company.

puja108:
I don't want to derail the numbering discussion too much, but I have to agree and would maybe go even one step further and say that even without the numbering RFCs as markdown PRs are not the most accessible. The idea at least from my side was that anyone that has an idea can jot it down and get early feedback and then we can have some lightweight process to be able to evaluate and move ideas to a next stage of either more research needed or actual first pocs or MVPs and inclusion in roadmap. To some degree RFCs have been quite successful as not a small number got created and discussions look good. However, they are definitely not very inclusive, like even Oli would usually not put his ideas in here and we've recently seen even engineers using the discussions feature of github for an idea/proposal and others are using google docs even others hackmd.
Discussions seem nice and there's sections like ideas or show and tell and there's voting/plus1s integrated, but one thing they miss is being able to directly comment on the text itself, which made me already hesitate using them before. 

That said, do we need to move the discussion around this somewhere else? Is this still really about numbers or should we reevaluate accesibility in a more general sense?

JosephSalisbury:
sorry for the delay, i've been wanting to get back to this.

i think we need to find and agree on a legit way to create RFCs via the web ui, and also keep them easy to identify. that should be straightforward. happy for that to be a call, or i might have a little poke around beforehand

i'm not super convinced about abandoning the whole thing just yet, the problems seems fairly tractable (my time has just been focused on hive sprint and then next steps for oncall)

JosephSalisbury:
actually i ate some rice and did some other things and i've completely changed my mind

rfcs are gonna be identifiable whatever - we're either gonna come up with use names for them, or we can just use the url - at worst, we can always find something else in future. it's more important to get people to use them

tl;dr - let's roll

JosephSalisbury (approved):

teemow:
Great :tada:.

I'll update all merged RFCs in this PR and then push updates and inform people about the unmerged RFCs.

teemow:
Done :rocket:

JosephSalisbury:
I've got a todo to add some rough CI to this repo, that should help in future too

PR #15, opened 2021-02-19
Add automatic cluster upgrades RFC

pipo02mix:
Let's automate the cluster upgrades!


0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,61 @@
+# RFC 0015 - Automatic Cluster Upgrades
+
+As a user, and also as KaaS team provider, I would like to have the clusters running in the latest version automatically, and not worry about CVEs or EOL of the cluster components.
+
+As a user I would like to define maintenance windows to ensure cluster are not upgraded out of these hours.
```
puja108, line 5:
```suggestion
As a user, I would like to define maintenance windows to ensure clusters are not upgraded outside of these hours.
```

puja108: In a way, you could think of either giving maintenance windows or do it the other way around and say "these times are not good". I know enterprise is rather used to former, but we could evaluate if latter might not also be feasible, to go towards "it should always be ok, and only exceptions to that are marked". Might also be sth for a future iteration though.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,61 @@
...
+3. The operator bump the version for the given set of clusters.
+
+## Open questions
+
+- Do we enable automatic upgrades by default? and then customer need to mark cluster(s) as frozen 
+
+- Do we setup a default upgrade window for cluster that does not have one defined/attached?
+
+- If we let customer to freeze the cluster we can apply for all type of upgrades? at some point it can be used for chart operator (or a new app upgrade operator) too to disable app upgrades?
+
```
puja108, line 51: we should also find out if settings should be per cluster only, or if we could have global/MC level settings, or per cluster type/label settings, sth to elaborate with customers on use cases I'd say

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,61 @@
...
+As a user, and also as KaaS team provider, I would like to have the clusters running in the latest version automatically, and not worry about CVEs or EOL of the cluster components.
+
+As a user I would like to define maintenance windows to ensure cluster are not upgraded out of these hours.
+
+As a user I would like to mark a cluster as frozen and block upgrades for certain period of time.
+
+## Reasoning
+
+Our customers need to plan every upgrade in their side and manage the upgrades for every GS version. At the same time, Cloud Native projects and Giant Swarm are realising more versions every time. It provokes a lot of time preparing and executing these upgrades which is not usually given much value to any of the stakeholders.
```
fiunchinho, line 11:
```suggestion
Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. At the same time, Cloud Native projects and Giant Swarm are releasing more versions every time. It provokes a lot of time preparing and executing these upgrades which is not usually given much value to any of the stakeholders. 
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: yeah we could come with a format the could include or exclude dates/times

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: yea something like `Upgrade policy`, where we have a default one by installation which has the good defaults, and then customers can create other policies and bind clusters to these ones

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+
+- The user can freeze (automatic) upgrades for the new cluster
+
+### Operator workflow
+
+1. Giant Swarm runs in each management cluster an operator that watches releases. 
+
+2. As part of the reconciliation loop, the operator checks if there is any cluster eligible for upgrade.
+ - There is a new patch/minor version active available. 
+ - There is clusters not in the latest version of a major release.
```
pipo02mix, line 37:
```suggestion
 - There are clusters not in the latest version of a major release. 
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,60 @@
...
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. At the same time, Cloud Native projects and Giant Swarm are releasing more versions every time. It provokes a lot of time preparing and executing these upgrades which is not usually given much value to any of the stakeholders. 
+
+Giant Swarm designed a versioning schema (semver), which includes all components on the stack, to ensure idempotency and expose the level of impact that changes bring to the customer. Patches and Minor releases should have no impact on the customer workloads, which mean can be applied in the environments at any time. 
+
+## Current state
+
+Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selection. The idea is to automate all this behavior and information under a Kubernetes operator and Custom Resource(s).
+
+The Solution engineer plan cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
```
pipo02mix, line 19:
```suggestion
The Solution engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+
+- As a user I want to schedule a upgrade for a certain day/time.
+
+- As a user, and also as KaaS team provider, I want to capture the result of upgrade.
+
+## Reasoning
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. Every customer has their own requirement when they need to schedule upgrades (freeze periods, maintenance windows,...). At the same time, Cloud Native projects and Giant Swarm are releasing more versions every time. It provokes a lot of time preparing and executing these upgrades which is not usually given much value to any of the stakeholders. 
+
+Giant Swarm designed a versioning schema (semver), which includes all components on the stack, to ensure idempotency and expose the level of impact that changes bring to the customer. Patches and Minor releases should have no impact on the customer workloads, which mean can be applied in the environments at any time.
```
fiunchinho, line 21: at any time? do you mean "automatically" instead? "any time" sounds like it wouldn't respect the maintenance / freeze windows.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+
+## User stories
+
+- As a user, and also as KaaS team provider, I would like to have the clusters running in the latest version automatically, and not worry about CVEs or EOL of the cluster components.
+
+- As a user, I would like to define maintenance windows to ensure clusters are not upgraded outside of these hours.
+
+- As a user I would like to mark a cluster as frozen and block upgrades for certain period of time.
+
+- As a user I want to define a policy that enforces some environments are upgraded before than others to an agreed version.
```
mig4, line 11:
```suggestion
- As a user I want to define a policy that enforces some environments are upgraded before others to an agreed version.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+
+- As a user I want to schedule a upgrade for a certain day/time.
```
mig4, line 13:
```suggestion
- As a user I want to schedule an upgrade on a certain day/time.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+
+- As a user, and also as KaaS team provider, I want to capture the result of upgrade.
+
+## Reasoning
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. Every customer has their own requirement when they need to schedule upgrades (freeze periods, maintenance windows,...). At the same time, Cloud Native projects and Giant Swarm are releasing more versions every time. It provokes a lot of time preparing and executing these upgrades which is not usually given much value to any of the stakeholders.
```
mig4, line 19:
```suggestion
Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. Every customer has their own requirement when they need to schedule upgrades (freeze periods, maintenance windows,...). At the same time, Cloud Native projects and Giant Swarm are continuously releasing new versions. This means a lot of time is consumed preparing and executing these upgrades which is not usually given much value by any of the stakeholders. 
```

Hope I got this right, you meant that the stakeholders are not giving much value to upgrades, right?

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+
+In cluster upgrade hive we have discuss around entities and ideas how this would look like.
+
+### Cluster Upgrade entity
+
+There are different reason why we are interested in having a object to define the upgrade intent. 
+
+- As we define in the user stories, we/customers would like to schedule upgrades on certain date from running version to a specific one. That way customer can plan upgrades ahead and so do we.
+
+- We have a registry of upgrades done to a cluster for auditing and tracking purposes.
```
mig4, line 41: Let's work out how this could work. Say I have a `Cluster` CR with `name: foo01` and an associated KCP with `version: 1.18.16`.

I then create a hypothetical `ClusterUpgrade` with a reference to the `foo01` instance of `Cluster` and a `version: 1.19.8` to express the desire to upgrade.

An _upgrade-operator_ then processes that `ClusterUpgrade`, orchestrates the upgrades where needed, sets the `status` on the `ClusterUpgrade` and we're done. But if we want to keep that `ClusterUpgrade` resource around as a record of upgrade history, how do we make sure it's not processed again?

A simple answer could be adding a label like `upgrade-operator.giantswarm.io/upgrade-done: true` or something, but is that going to be performant enough?

The average number of clusters per installation is probably below 10, so not too bad. What about how many upgrades do clusters go through in their lifetime, on average?

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,95 @@
...
+The upgrade policy would contain
+
+- A way to define dates or times where clusters (or relatives) can be modified.
+
+- A way to define dates or times where clusters (or relatives) cannot be modified.
+
+## Technical aspects
+
+We envision aforementioned concepts would become custom resource definitions and there will be additional controller that would enhance and ensure the correct behavior.
+
```
mig4, line 62: I'd like to also start hashing out what the resources we're talking about would look like. Let's say something like this for starters:

```suggestion

### Resources

#### Cluster Upgrade

\`\`\` yaml
apiVersion: 
kind: ClusterUpgrade
metadata:
  name: upgrade-to-release-v20
  namespace: default
spec:
  clusterRef:
    apiVersion: cluster.x-k8s.io/v1alpha3
    kind: Cluster
    name: foo01
  kubernetesVersion: 1.19.8
\`\`\`
```

(WIP)

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: well I can rephrase it but I wanted to emphasize `patches` and `minors` should be harmless. I mention the use case of customer need to define constraints later, is that enough?

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: you got it right

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
+# RFC 0015 - Automatic Cluster Upgrades
+
+## User stories
+
+- As a user, and also as KaaS team provider, I would like to have the clusters running in the latest version automatically, and not worry about CVEs or EOL of the cluster components.
```
marians, line 5: I'm not sure what `KaaS team provider` means. If it means a Giant Swarm KaaS provider team, I'm not sure it is necessary to encode that in an RFC as a stakeholder.

Also I'd try to avoid abbreviations in favour of simplicity and accessibility of the content.

```suggestion
- As a Giant Swarm customer, and also as Giant Swarm, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As a user, I would like to define maintenance windows to ensure clusters are not upgraded outside of these hours.
```
marians, line 7:
```suggestion
- As a customer admin, I would like to define maintenance windows to ensure workload clusters are only upgraded within these hours.
```

user -> customer admin, to be more specific.

Changed into positive language instead of negative.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
+# RFC 0015 - Automatic Cluster Upgrades
```
marians, line 1:
```suggestion
# RFC 0015 - Automatic workload cluster upgrades
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+## User stories
+
+- As a user, and also as KaaS team provider, I would like to have the clusters running in the latest version automatically, and not worry about CVEs or EOL of the cluster components.
+
+- As a user, I would like to define maintenance windows to ensure clusters are not upgraded outside of these hours.
+
+- As a user I would like to mark a cluster as frozen and block upgrades for certain period of time.
```
marians, line 9:
```suggestion
- As a customer admin, I would like to mark a cluster as "frozen" in order to block upgrades for a certain time period.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As a user I want to define a policy that enforces some environments are upgraded before others to an agreed version.
```
marians, line 11:
```suggestion
- As a customer admin, I want to define the upgrade order of clusters, so that certain workload clusters are upgraded to a distinct workload cluster release before others are.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As a user I want to schedule an upgrade on a certain day/time.
```
marians, line 13:
```suggestion
- As a customer admin, I want to schedule an upgrade on a certain day/time.
```

Does this mean the upgrade of one distinct workload cluster to one distinct workload cluster release? If yes, we may clarify this.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As a user, and also as KaaS team provider, I want to capture the result of upgrade.
```
marians, line 15:
```suggestion
- As a customer, and also as Giant Swarm, we want to capture the result of an upgrade.
```

I'd like to see this ending in `, so that <whatever the goal is>` to explain why we should capture the result.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Reasoning
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. Every customer has their own requirement when they need to schedule upgrades (freeze periods, maintenance windows,...). At the same time, Cloud Native projects and Giant Swarm are continuously releasing new versions. This means a lot of time is consumed preparing and executing these upgrades which is not usually given much value by any of the stakeholders.
```
marians, line 19:
```suggestion
Our customers need to plan every upgrade on their side and manage the upgrades for every workload cluster release. Each customer has their own requirements regarding when they need to schedule upgrades (freeze periods, maintenance windows, etc.). At the same time, cloud native projects and Giant Swarm are continuously providing new releases. The preparation and execution of upgrades is time and labor intensive, while the fact that they are executed manually does not add much value to stakeholders.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As a user I want to schedule an upgrade on a certain day/time.
+
+- As a user, and also as KaaS team provider, I want to capture the result of upgrade.
+
+## Reasoning
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every GS version. Every customer has their own requirement when they need to schedule upgrades (freeze periods, maintenance windows,...). At the same time, Cloud Native projects and Giant Swarm are continuously releasing new versions. This means a lot of time is consumed preparing and executing these upgrades which is not usually given much value by any of the stakeholders. 
+
+Giant Swarm designed a versioning schema (semver), which includes all components on the stack, to ensure idempotency and expose the level of impact that changes bring to the customer. Patches and Minor releases should have no impact on the customer workloads, which mean can be applied in the environments at any time.
```
marians, line 21:
```suggestion
Giant Swarm designed a versioning scheme (based on Semver), which includes all components on the stack, to ensure idempotence and expose the level of impact that changes bring to the customer. Patches and Minor releases should have no impact on the customer workloads, which means that they can be applied in the environments at any time.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selection. The idea is to automate all this behavior and information under a Kubernetes operator(s) and Custom Resource(s).
+
+The Solution engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
+
+Based on the fact Giant Swarm is moving to Cluster API this is story let us define a new greenfield scenario where we can design a model that can be more generic and not focus only in Giant Swarm platform. The idea would be contribute as much as possible with upstream due the fact most of this functionality would be useful for everyone that manages clusters.
+
+## Concepts
+
+In cluster upgrade hive we have discuss around entities and ideas how this would look like.
```
marians, line 33: I would remove this sentence

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Cluster Upgrade entity
+
+There are different reason why we are interested in having a object to define the upgrade intent.
```
marians, line 37:
```suggestion
There are different reason why we are interested in having an object to define the upgrade intent. 
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+The Solution engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
+
+Based on the fact Giant Swarm is moving to Cluster API this is story let us define a new greenfield scenario where we can design a model that can be more generic and not focus only in Giant Swarm platform. The idea would be contribute as much as possible with upstream due the fact most of this functionality would be useful for everyone that manages clusters.
+
+## Concepts
+
+In cluster upgrade hive we have discuss around entities and ideas how this would look like.
+
+### Cluster Upgrade entity
```
marians, line 35:
```suggestion
### Cluster upgrade entity
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As we define in the user stories, we/customers would like to schedule upgrades on certain date from running version to a specific one. That way customer can plan upgrades ahead and so do we.
+
+- We have a registry of upgrades done to a cluster for auditing and tracking purposes.
+
+- We can save cluster upgrade status displaying if something goes wrong or not.
+
+- In theory we can include apps upgrades together or separately.
+
+### Upgrade Policy entity
```
marians, line 47:
```suggestion
### Upgrade policy entity
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- A way to define dates or times where clusters (or relatives) cannot be modified.
+
+## Technical aspects
+
+We envision aforementioned concepts would become custom resource definitions and there will be additional controller that would enhance and ensure the correct behavior.
+
+### Resources
+
+#### Cluster Upgrade
```
marians, line 65:
```suggestion
#### ClusterUpgrade
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+  namespace: default
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  kubernetesVersion: 1.19.8
+\```
+
+#### Upgrade Policy
```
marians, line 83:
```suggestion
#### UpgradePolicy
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+metadata:
+  name: upgrade-to-release-v20
+  namespace: default
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  kubernetesVersion: 1.19.8
+\```
```
marians, line 81:
```suggestion
```yaml
apiVersion: 
kind: ClusterUpgrade
metadata:
  name: upgrade-to-release-v20
  namespace: default
spec:
  clusterRef:
    apiVersion: cluster.x-k8s.io/v1alpha3
    kind: Cluster
    name: foo01
  kubernetesVersion: 1.19.8
```
```

File: 0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md [hunk 28]
```diff
@@ -0,0 +1,139 @@
...
+  - name: working-hours
+    startTime: 2021-03-10T07:00:00Z
+    endTime: 2021-03-10T17:00:00Z
+    recurrence: FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR
+  maintenanceExclusions:
+  - name: month-start
+    startTime: 2021-03-01T00:00:00Z
+    endTime: 2021-03-10T23:59:59Z
+    recurrence: FREQ=MONTHLY
+\```
```
marians, line 104:
```suggestion
```yaml
apiVersion: 
kind: UpgradePolicy
metadata:
  name: working-hours-except-month-start
  namespace: default
spec:
  maintenanceWindows:
  - name: working-hours
    startTime: 2021-03-10T07:00:00Z
    endTime: 2021-03-10T17:00:00Z
    recurrence: FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR
  maintenanceExclusions:
  - name: month-start
    startTime: 2021-03-01T00:00:00Z
    endTime: 2021-03-10T23:59:59Z
    recurrence: FREQ=MONTHLY
```
```

File: 0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md [hunk 29]
```diff
@@ -0,0 +1,139 @@
...
+
+This would allow upgrades to happen between 07:00 UTC and 17:00 UTC on Monday-Friday, except for days between 1st and 10th of a month.
+### Controllers
```
marians, line 107:
```suggestion
This would allow upgrades to happen between 07:00 UTC and 17:00 UTC on Monday-Friday, except for days between 1st and 10th of a month.

### Controllers
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: The idea is customer want to plan upgrades for the clusters. So my dev cluster `abcde` will be upgraded 14/04/21 at 2:00 am to version `14.0.0`

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,137 @@
...
+
+- As a customer admin, I would like to define maintenance windows to ensure workload clusters are only upgraded within these hours.
+
+- As a customer admin, I would like to mark a workload cluster as "frozen" in order to block upgrades for a certain time period.
+
+- As a customer admin, I want to define the upgrade order of clusters, so that certain workload clusters are upgraded to a distinct workload cluster release before others are.
+
+- As a user I want to schedule an upgrade on a certain day/time.
+
+- As a user, and also as KaaS team provider, I want to capture the result of upgrade.
```
pipo02mix, line 15:
```suggestion
- As a customer, and also as Giant Swarm, we want to capture the result of an upgrade. Based on the result we can alert on a certain condition, track the history for certain cluster(s) or automate response based on the end state.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,137 @@
...
+
+Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selection. The idea is to automate all this behavior and information under a Kubernetes operator(s) and Custom Resource(s).
+
+The Solution engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
+
+Based on the fact Giant Swarm is moving to Cluster API this is story let us define a new greenfield scenario where we can design a model that can be more generic and not focus only in Giant Swarm platform. The idea would be contribute as much as possible with upstream due the fact most of this functionality would be useful for everyone that manages clusters.
+
+## Concepts
+
+In cluster upgrade hive we have discuss around entities and ideas how this would look like.
```
pipo02mix, line 33:
```suggestion
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
mig4: That's a good question, that should be enough to have a cleanup for older resources.

I'm not really sure if metrics are the right thing, but let's first narrow down what information about upgrades we care about keeping:

- that an upgrade happened: for sure 😉 
- how it was triggered: auto / manual
- outcome: success / fail (rollback?)
- from GS release?
- to GS release
- do we want to have a list of all components that were actually changed? like upgrade from `v13.0.0` to `v13.0.1` would only have changed the version of `kvm-operator`, do we need that level of detail?
- anything else?

pipo02mix: > do we want to have a list of all components that were actually changed? like upgrade from v13.0.0 to v13.0.1 would only have changed the version of kvm-operator, do we need that level of detail?

I guess knowing from which GS version we started and which we ended, will give us all info

> anything else?

When it was triggered and when it was done.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+kind: ClusterUpgrade
+metadata:
+  name: upgrade-to-release-v20
+  namespace: default
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  kubernetesVersion: 1.19.8
```
mig4, line 79: Hmm, actually just thinking about it now, our _upgrade intent_ resource should only specify the Giant Swarm release we want to upgrade to, right? And _upgrade-operator_ will figure out the versions of components to set based on `Release` CR?

mig4, line 79: First stab at what `status` might look like:

```suggestion
  kubernetesVersion: 1.19.8
status:
  triggeredBy: auto / release
  triggeredAt: 2021-03-10T10:00:00Z
  outcome: success
  finishedAt: 2021-03-10T10:31:10Z
  fromRelease: v13.0.0
  toRelease: v13.0.1
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+- As we define in the user stories, we/customers would like to schedule upgrades on certain date from running version to a specific one. That way customer can plan upgrades ahead and so do we.
+
+- We have a registry of upgrades done to a cluster for auditing and tracking purposes.
+
+- We can save cluster upgrade status displaying if something goes wrong or not.
+
+- In theory we can include apps upgrades together or separately.
+
+### Upgrade policy entity
```
mig4, line 46: Would you mind if we renamed this to _"maintenance schedule"_? This feels like a lower level thing that we could later on separate and push upstream. 

I'm also thinking whether we may want to later introduce a higher-level _"upgrade policy"_ to specify things like transitions that can happen automatically (e.g. auto-minor version upgrades, or auto-patch version upgrades), or somehow to encode things like in which order node pools should be upgraded.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
marians: ACK. Suggestion updated.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
mig4: From the chat we had:

- storing that information in a metric doesn't really fit the concept, with `from` and `to` versions which would have to be stored as labels you'd have unbounded sets of values and to represent start time and end time you'd need to do something like emitting the metric with value 1 when an upgrade is in progress and value 0 when it's not and then rely on history in cortex/grafana cloud or something
- even a couple of hundred of objects on large installations shouldn't be an issue, in some cases we're already dealing with thousands of objects and things work fine

TL;DR I'm happy enough with this approach.

I've also taken the list of bits of information we want to preserve and added to suggestion of how the status of `ClusterUpgrade` could look like.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
mig4: We had a chat and decided that yes, this should only reference the `Release` that should be upgraded to, in order to:

- avoid duplicating the information about components between `Release` and `ClusterUpgrade`
- avoid introducing a way to side-step the linking between component versions encoded in a `Release` (as someone could create a `ClusterUpgrade` with versions whose combination is not encoded in a `Release` and so is untested)

I've updated my suggestion below.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+kind: UpgradePolicy
+metadata:
+  name: working-hours-except-month-start
+  namespace: default
+spec:
+  maintenanceWindows:
+  - name: working-hours
+    startTime: 2021-03-10T07:00:00Z
+    endTime: 2021-03-10T17:00:00Z
+    recurrence: FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR
```
pipo02mix, line 104: @mig4 is this any kind of standard? just wondering

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,139 @@
...
+
+This would allow upgrades to happen between 07:00 UTC and 17:00 UTC on Monday-Friday, except for days between 1st and 10th of a month.
+
+### Controllers
+
+#### Cluster upgrade executor
+
+Thee will be a cluster that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps. 
+
+This operator can also check on the `Upgrade Policy` if the cluster can be done or not, and in case it cannot happen alert or set an appropriate status on the object status.
```
pipo02mix, line 120:
```suggestion
There will be an operator that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps at some point though ideally would detach apps. totally from the cluster and. apps metadata and App Platform enforce the requirements. 

This operator can also check on the `Upgrade Policy` to verify the upgrade fulfil or not the criteria, and in case it does not, alert or set an appropriate status on the object status.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
mig4: Yes, it's an `RRULE` (recurrence rule) described in [RFC-5545](https://tools.ietf.org/html/rfc5545), there are libraries in Go (like [rrule-go](https://pkg.go.dev/github.com/teambition/rrule-go)) to parse and handle those so I thought it would be useful. This is also what [Google Cloud uses](https://cloud.google.com/kubernetes-engine/docs/how-to/maintenance-windows-and-exclusions#maintenance-window) for maintenance windows. But we can use something else if there's something better sure.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+
+- We can save cluster upgrade status displaying if something goes wrong or not.
+
+- In theory we can include apps upgrades together or separately.
+
+### Maintenance schedule entity
+
+Our customer can have a different set of constraints when it comes to allow changes in their clusters. The maintenance schedule would define those and then the clusters can rely on a policy to enforce the desired behavior.
+
+The maintenance schedule would be possible to be referenced from different cluster resources (Cluster,Machine Pool, KubeadmControlPlane, Apps,...) letting controllers to decide if the resource would be or not reconciled.
```
fiunchinho, line 49: If I understand correctly, the resources would be still reconciled, but they wouldn't be upgraded.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+metadata:
+  name: upgrade-to-release-v20
+  namespace: default
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  release: v13.0.1
+  startTime: 2021-03-10T07:00:00Z
```
fiunchinho, line 79: I was thinking that we may use a reference to the `Release` object in a similar fashion to what we do with the `Cluster`. And also `startTime` seems to be similar to what upstream is doing with the field `upgradeAfter`, so maybe we can use the same field name.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
fiunchinho: I think we can link to the Google Cloud page somewhere in the RFC, as it seems we are following a similar approach, and it shows other providers are doing similar things.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+    recurrence: FREQ=MONTHLY
+```
+
+This would allow upgrades to happen between 07:00 UTC and 17:00 UTC on Monday-Friday, except for days between 1st and 10th of a month.
+
+### Controllers
+
+#### Cluster upgrade executor
+
+There will be an operator that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps at some point though ideally would detach apps. totally from the cluster and. apps metadata and App Platform enforce the requirements.
```
fiunchinho, line 118:
```suggestion
There will be an operator that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps at some point though ideally would detach apps totally from the cluster and apps metadata and App Platform enforce the requirements. 
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+
+There will be an operator that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps at some point though ideally would detach apps. totally from the cluster and. apps metadata and App Platform enforce the requirements. 
+
+This operator can also check on the `maintenance schedule` to verify the upgrade fulfil or not the criteria, and in case it does not, alert or set an appropriate status on the object status.
+
+Also the operator will handle the mechanics of the upgrade like which `MachinePool` is upgraded first or wait till one `MachinePool` is done successfully to continue with next one. There is already an [RFC](https://github.com/giantswarm/rfc/pull/17/files) up for this specific controller.
+
+#### Cluster upgrade scheduler
+
+Ideally to automate the whole process there will be an operator that creates all the `Cluster Upgrade` CRs when a new release of Giant Swarm is created (and/or possibly other events).
```
fiunchinho, line 126: Would this automation create `Cluster Upgrade` CR's for all new GiantSwarm releases? only for minor / patches? is it configurable by cluster?

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Existing CAPI controllers
+
+There was [a proposal in CAPI upstream](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md) for the KubeadmControlPlane (later included in CAPZ components) to add `upgradeAfter` parameter to influence on the upgrades. The idea would be to create a new proposal that allows to point a new entity (`UpgradePolicy`) to extend the possibilities more than a single timestamp. Later controllers could leverage on that to allow changes or not to their reconciled resources.
```
fiunchinho, line 130: If I understand correctly, the `upgradeAfter` field was added to be able to trigger upgrades even when there is nothing to change (it's useful to re-create infrastructure and re-generate certs, for example). So I believe this wouldn't replace the `upgradeAfter` field, but it would be a new field. When the controller reconciling the CR's wants to upgrade something due to the `upgradeAfter` field, it would need to check the `UpgradePolicy`.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: yes it is the case. I guess at the beginning only the upgrade operator would reconcile it but maybe in the future upstream operators want to rely on it

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,148 @@
...
+kind: ClusterUpgrade
+metadata:
+  name: upgrade-to-release-v20
+  namespace: default
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  release: v13.0.1
```
pipo02mix, line 78:
```suggestion
  clusterRef:
    apiVersion: cluster.x-k8s.io/v1alpha3
    kind: Cluster
    name: foo01
  releaseRef:
    apiVersion: release.giantswarm.io/v1alpha1
    kind: Release
    name: v13.0.1
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,151 @@
...
+spec:
+  clusterRef:
+    apiVersion: cluster.x-k8s.io/v1alpha3
+    kind: Cluster
+    name: foo01
+  releaseRef:
+    apiVersion: release.giantswarm.io/v1alpha1
+    kind: Release
+    name: v13.0.1
+  startTime: 2021-03-10T07:00:00Z
```
pipo02mix, line 82: Thanks to @mig4 to clarify `upgradeAfter` has other meaning https://github.com/kubernetes-sigs/cluster-api/pull/4285/files. 

Anyway we could think in two different approaches that can live together. One is to add a field like `startTime` to define exactly when cluster must be upgraded (even not upgrade the cluster if there has been a problem as X minutes has passed the startTime), and/or add something like `upgradeAfter` which just means you want to upgrade when the time goes over the `upgradeAfter` field. I tend to think to avoid later, first, the field already exist upstream with a different purpose and second, this sounds like a policy/maintananceWindow scope, WDYT?

 @mig4 @fiunchinho

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: I guess only for minor/patches but at some point we can do it for major to ensure K8s EOL (give the option to customers)

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: yes it is the case

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,151 @@
...
+
+Also the operator will handle the mechanics of the upgrade like which `MachinePool` is upgraded first or wait till one `MachinePool` is done successfully to continue with next one. There is already an [RFC](https://github.com/giantswarm/rfc/pull/17/files) up for this specific controller.
+
+#### Cluster upgrade scheduler
+
+Ideally to automate the whole process there will be an operator that creates all the `Cluster Upgrade` CRs when a new release of Giant Swarm is created (and/or possibly other events).
+
+#### Existing CAPI controllers
+
+There was [a proposal in CAPI upstream](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md) for the KubeadmControlPlane (later included in CAPZ components) to add `upgradeAfter` parameter to influence on the upgrades. The idea would be to create a new proposal that allows to point a new entity (`UpgradePolicy`) to extend the possibilities more than a single timestamp. Later controllers could leverage on that to allow changes or not to their reconciled resources.
```
pipo02mix, line 133:
```suggestion
There was [a proposal in CAPI upstream](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md) for the KubeadmControlPlane (later included in CAPZ components) to add `upgradeAfter` parameter to influence the upgrades. Though this field only means a cluster upgrade will be triggered after that date, to force an upgrade, it is related to the goal of controlling upgrades. The idea would be to create a new proposal that allows pointing a new entity (`MaintenanceSchedule `/`UpgradePolicy`)  to extend the possibilities more than a single timestamp. Later controllers could leverage on that to allow changes or not to their reconciled resources.
```

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
fiunchinho: If we have the `MaintenanceSchedule` CRD to define the maintenance window, I guess the `ClusterUpgrade` would hold a reference to a `MaintenanceSchedule`?

puja108: hmm, maybe that is in the status then?

fiunchinho: How come? wouldn't a `ClusterUpgrade` need to select which `MaintenanceSchedule` to use?

puja108: hmm, seems more complex now that I think of it, the schedule only gives permission to upgrade right? then an upgrade scheduler can use this to create upgrade CRs automatically, so there should not be a need for any reference I think. But then my question would be, why even need a startTime here? If you want to trigger an upgrade just create the CR, start time is the creation time of the CR, the creator of the CR (here the cluster upgrade scheduler) has to adhere to maintenanceschedule or it could be validated in admission. In the end it depends a lot on how you want to implement this in the end, but as it is described below to me it makes more sense if the scheduler knows about maintenance and then the executor just executes.

From what we tried to come up with in batman back then for auto-updates of apps, we thought to align it more with the concepts that people are used to from autoscaling in HPA as in the end the upgrade is just a scale operation on a version field, so you could set an autoupgrade policy like upgrade all patch or all patch and minor (that can be expressed with semver notation even) and reference the maintenance schedule in or from there, then with simple CRs like the App CR the upgrade controller can just increment the version field based on a new version being present and in more complex situations like the cluster upgrade you might wanna delegate to another controller that can handle upgrade strategies, and delegation could be through sth like the upgrade CR here.

The above is not very far from what you describe here. It only is missing autoupgrade policy being configurable and upgrade strategies, but that's ok if you wanna define an MVP scope here and add stuff like that later.

othylmann (approved):

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
pipo02mix: > But then my question would be, why even need a startTime here? 

I guess we will enable teams to plan upgrades. So they say next Monday we will upgrade cluster X and cluster Y at 9:00 (and they prepare the objects), on Tuesday afternoon we will upgrade cluster Z...

> From what we tried to come up with in batman back then for auto-updates of apps, we thought to align it more...

I started with that idea but in reality customers (at least as it is today) dont like to lose the control of an upgrade. It could be interesting to get all reasons why and maybe build functionality on the upgrade mechanism to avoid them

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,151 @@
+# RFC 0015 - Automatic workload cluster upgrades
+
+## User stories
+
+- As a Giant Swarm customer, and also as Giant Swarm, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
+
+- As a customer admin, I would like to define maintenance windows to ensure workload clusters are only upgraded within these hours.
```
marians, line 7: I'd like to explicitly call it "upgrade window". Calling it "maintenance window" suggests that the window is defined for other purposes, such as manual maintenance, while in fact this RFC should be focused on upgrades.

```suggestion
- As a customer admin, I would like to define upgrade windows, to ensure that upgrades happen when disruption is most permittable and impact on production traffic is minimal.
```

If you accept this suggestion, please also update other occurrences of "maintenance window".

marians (commented):
As commented in the text, I am wondering about the specifity of this RFC. In the title and use cases, it speaks explicitly about upgrades and upgrade scheduling. Then, in the proposed solution, it gets more general and becomes about "maintenance".

If the goal here is to solve automated upgrades only, then I would make the terminology more specific. This will help in the future to interpret the meaning of the artefacts we create (e. g. CRDs and CRs) correctly.

If the goal is to solve configuring maintenance windows that are applicable to all sorts of maintenance and manual ops work, then I think the title and use case list need adjustment.

teemow:
Please make sure to pull the changes before making new modifications. I've renamed the files as we agreed on a slightly different structure for RFCs.

0015-automatic-cluster-upgrades/0015-automatic-cluster-upgrades.md
```diff
@@ -0,0 +1,151 @@
...
+
+- As we define in the user stories, we/customers would like to schedule upgrades on certain date from running version to a specific one. That way customer can plan upgrades ahead and so do we.
+
+- We have a registry of upgrades done to a cluster for auditing and tracking purposes.
+
+- We can save cluster upgrade status displaying if something goes wrong or not.
+
+- In theory we can include apps upgrades together or separately.
+
+### Maintenance schedule entity
```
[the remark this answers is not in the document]
pipo02mix: I get your point. My first idea was taking profit from this implementation to enforce clusters (like production clusters) are not modified out of these windows. In theory, it does only happen when you press upgrade but in reality, we have seen other changes in the Mgmt clusters that have provoked changes on workload clusters (rolling nodes). Depend on the implementation we do, we can block operators to touch WCs out of these windows.

IMO it is a good idea, though it is not the main purpose of this RFC.  If we do all naming only thinking into upgrades it will really hard to change later. Maintainance window is a more general term that expresses the intention more accurate manner. WDYT?

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
+# RFC 0015 - Automatic workload cluster upgrades
+
+## User stories
+
+- As a Giant Swarm customer, and also as Giant Swarm, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
```
pipo02mix, line 5:
```suggestion
- As a Giant Swarm customer, and also as a Giant Swarm maintainer, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Reasoning
+
+Our customers need to plan every upgrade on their side and manage the upgrades for every workload cluster release. Each customer has their own requirements regarding when they need to schedule upgrades (freeze periods, maintenance windows, etc.). At the same time, cloud native projects and Giant Swarm are continuously providing new releases. The preparation and execution of upgrades is time and labor intensive, while the fact that they are executed manually does not add much value to stakeholders.
+
+Giant Swarm designed a versioning scheme (based on Semver), which includes all components on the stack, to ensure idempotence and expose the level of impact that changes bring to the customer. Patches and Minor releases should have no impact on the customer workloads, which means that they can be applied in the environments at any time.
+
+## Current state
+
+Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selection. The idea is to automate all this behavior and information under a Kubernetes operator(s) and Custom Resource(s).
```
pipo02mix, line 25:
```suggestion
Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selected. The idea is to automate all this behaviour and information under a Kubernetes operator(s) and Custom Resource(s).
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+The Solution engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
```
pipo02mix, line 27:
```suggestion
The Solution Engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
+# RFC 0015 - Automatic workload cluster upgrades
+
+## User stories
+
+- As a Giant Swarm customer, and also as a Giant Swarm maintainer, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
+
+- As a customer admin, I would like to define maintenance windows to ensure workload clusters are only upgraded within these hours.
```
pipo02mix, line 7:
```suggestion
- As a customer admin, I would like to define upgrade windows, to ensure that upgrades happen when disruption is most permittable and impact on production traffic is minimal.
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+Ideally to automate the whole process there will be an operator that creates all the `Cluster Upgrade` CRs when a new release of Giant Swarm is created (and/or possibly other events).
+
+#### Existing CAPI controllers
+
+There was [a proposal in CAPI upstream](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md) for the KubeadmControlPlane (later included in CAPZ components) to add `upgradeAfter` parameter to influence the upgrades. Though this field only means a cluster upgrade will be triggered after that date, to force an upgrade, it is related to the goal of controlling upgrades. The idea would be to create a new proposal that allows pointing a new entity (`MaintenanceSchedule `/`UpgradePolicy`)  to extend the possibilities more than a single timestamp. Later controllers could leverage on that to allow changes or not to their reconciled resources.
+
+## Open questions
+
+- Do we scheduled automatic upgrades by default? or we let customer to scheduled them (maybe suggesting in our UI a cluster upgrade need to be scheduled)?
```
marians, line 137:
```suggestion
- Do we schedule automatic upgrades by default? Or do we let customer schedule them (maybe suggesting in our UI a cluster upgrade need to be scheduled)?
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Existing CAPI controllers
+
+There was [a proposal in CAPI upstream](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md) for the KubeadmControlPlane (later included in CAPZ components) to add `upgradeAfter` parameter to influence the upgrades. Though this field only means a cluster upgrade will be triggered after that date, to force an upgrade, it is related to the goal of controlling upgrades. The idea would be to create a new proposal that allows pointing a new entity (`MaintenanceSchedule `/`UpgradePolicy`)  to extend the possibilities more than a single timestamp. Later controllers could leverage on that to allow changes or not to their reconciled resources.
+
+## Open questions
+
+- Do we scheduled automatic upgrades by default? or we let customer to scheduled them (maybe suggesting in our UI a cluster upgrade need to be scheduled)? 
+
+- Do we setup a default maintenance schedule for cluster that does not have one defined/attached?
```
marians, line 139:
```suggestion
- Do we set up a default maintenance schedule for a cluster that does not have one defined/attached?
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+- If we let customer to freeze the cluster we can apply for all type of upgrades? at some point it can be used for chart operator (or a new app upgrade operator) too to disable app upgrades?
```
marians, line 141:
```suggestion
- If we allow a customer to freeze the cluster, can apply all types of upgrades? Can this mechanism also be used at some point by chart-operator (or a new app upgrade operator) to disable app upgrades?
```

automatic-cluster-upgrades/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Additional actions
+
+- Adapt our monitoring to not trigger false positive during upgrades. Right now, the upgrades usually trigger alerts (like `ServiceLevelBurnRateTooHigh`) which are false positives (all nodes are rolled during upgrade, so it should be omitted). At the same time having specific alerts for upgrades can be valuable (like one when master is not coming up after X minutes)
```
marians, line 145:
```suggestion
- Adapt our monitoring to not trigger false positives during upgrades. Right now, the upgrades usually trigger alerts (like `ServiceLevelBurnRateTooHigh`) which are false positives (all nodes are rolled during upgrade, so it should be omitted). At the same time having specific alerts for upgrades can be valuable (like one when master is not coming up after X minutes)
```

marians (approved):

pipo02mix:
In my opinion, this RFC is mature enough to be merged. Please @fiunchinho @puja108 @mig4 raise your voice if you are against it.

I think is important to have it ready especially now that @giantswarm/team-firecracker  has started to create an upgrade scheduler operator

fiunchinho:
I think it's fine. Let's merge it Fer! That way we can point people to the RFC

pipo02mix:
ok I am going ahead, fasten your seat belts :D

PR #17, opened 2021-03-09
Add capi upgrades RFC

pipo02mix:


 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
+# RFC 016 - CLUSTER API UPGRADES
+
+## Reasoning
+
+As part of the Cluster API (CAPI) Hive Sprint, the `upgrades hive` decided to create a document to gather all actions need to be done in order to upgrade a workload cluster successfully as well as bring the functionality we have already built to trigger the upgrades.
+
+## Scope
+
+This document only cover the action to upgrade the workload cluster without solving problems of versioning of the different CAPI controllers and/or deciding which controller reconcile a cluster.
```
fiunchinho, line 9:
```suggestion
This document covers only the action to upgrade the workload cluster without solving problems related to the versioning of the different CAPI controllers and/or deciding which controller reconciles a cluster.
```

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+
+The current approach to upgrade a workload cluster in CAPI world involves a series of steps described below:
+
+- Ensure there is a `machine image` with the correct target Kubernetes version (contains kubelet and kubeadm). 
+- Upgrade the control plane machines:
+  - Generate a new infrastructure `MachineTemplate` resource (with new image ID if needed) and create it in the Management API.
+  - Set the new Kubernetes version in `KubeadmControlPlane` resource.
+  - Change the infrastructure `MachineTemplate` on the `KubeadmControlPlane` to the new one and apply the new version of `KubeadmControlPlane`.
+- Upgrade machine pools (workers):
+  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed).
```
fiunchinho, line 23:
```suggestion
  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed.
```

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+## Context
+
+The current approach to upgrade a workload cluster in CAPI world involves a series of steps described below:
+
+- Ensure there is a `machine image` with the correct target Kubernetes version (contains kubelet and kubeadm). 
+- Upgrade the control plane machines:
+  - Generate a new infrastructure `MachineTemplate` resource (with new image ID if needed) and create it in the Management API.
+  - Set the new Kubernetes version in `KubeadmControlPlane` resource.
+  - Change the infrastructure `MachineTemplate` on the `KubeadmControlPlane` to the new one and apply the new version of `KubeadmControlPlane`.
+- Upgrade machine pools (workers):
```
fiunchinho, line 22: Since we are explaining the upgrade process, I'd add here that we want to upgrade machine pools one by one, not all at once. I'd write something like 

> Upgrade machine pools (workers). We loop through all existing machine pools and for each machine pool
> - Make sure there are no `MachinePools` being already upgraded.
> - Modify the infrastructure `MachinePool` resource (with new image ID) if needed.
> - Set the new Kubernetes version in `MachinePool` resource.

What do you think?

 0016-capi-upgrades/0016-capi-upgrades.md
fiunchinho, line 9: I'd also make clear which upgrades we are focusing on.

A workload cluster can be ugpraded in different ways
- apps installed in the cluster are upgraded (out of scope here)
- cluster is reconciled by a newer CAPI controller
- machine image used by control plane or workers is updated with new k8s version or new OS version

#### Apps installed in the workload cluster
This upgrade is out of scope, and we have other non-CAPI components that already take care of that, like app-operator.

#### Reconcile Cluster with new CAPI controller release
We may want to upgrade the Cluster to a newer CAPI controller release, which may change the infrastructure created to support the cluster (network, load balancers, etc). We do this update just changing the CAPI label on the Cluster CR, and we don't need a new machine image to be created/referenced by the CR's.

#### Upgrade k8s version or OS version
Finally we may want to update the k8s version or the OS version running in the machines. In both cases, we need a new machine image using the updated components, and reference that image from the CR's.

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed).
+  - Set the new Kubernetes version in `MachinePool` resource.
+
+Also our operators watch for changes on the Cluster CR's labels to trigger an upgrade. That functionality should be kept in the new operator we build.
+
+## Open questions
+
+- Do we need to change the reference of the image ID for every upgrade?
```
fiunchinho, line 30: I tried to explain above that not all upgrades require a change in the machine image, but all upgrades related to k8s or the OS version require a new image.

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+- For `MachinePool`, do we need to make the operation in a transactional way to avoid multiple upgrade (in case we need to change infrastructure and agnostic `MachinePool`)?
+- Are the provider controllers aware of KCP upgrades and wait till they are done before start upgrading Machine Pools?
+- Do we need to allow different strategics to define the sequence `MachinePool` resoures are upgraded?
```
fiunchinho, line 33: I believe it's fine if we support one specific strategy, which is upgrading `MachinePools` one by one. I've seen customers asking to increase the time before each `MachinePool` is upgraded, like we upgrade the first `MachinePool` , and once is done, instead of upgrading the next one, we wait for some time. Do you remember other strategies/customizations requested by customers?

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed).
+  - Set the new Kubernetes version in `MachinePool` resource.
+
+Also our operators watch for changes on the Cluster CR's labels to trigger an upgrade. That functionality should be kept in the new operator we build.
+
+## Open questions
+
+- Do we need to change the reference of the image ID for every upgrade?
+- For `MachinePool`, do we need to make the operation in a transactional way to avoid multiple upgrade (in case we need to change infrastructure and agnostic `MachinePool`)?
+- Are the provider controllers aware of KCP upgrades and wait till they are done before start upgrading Machine Pools?
```
fiunchinho, line 32: No that I'm aware of. Since they are independent controllers, they will do its thing without waiting. I believe the external tool that orchestrates the upgrade would have to trigger the worker nodes upgrade only when the control plane is done upgrading.

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,33 @@
...
+- Upgrade machine pools (workers):
+  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed).
+  - Set the new Kubernetes version in `MachinePool` resource.
+
+Also our operators watch for changes on the Cluster CR's labels to trigger an upgrade. That functionality should be kept in the new operator we build.
+
+## Open questions
+
+- Do we need to change the reference of the image ID for every upgrade?
+- For `MachinePool`, do we need to make the operation in a transactional way to avoid multiple upgrade (in case we need to change infrastructure and agnostic `MachinePool`)?
```
fiunchinho, line 31: I'm not sure I understand the question.

pipo02mix:
@fiunchinho I have included what we agreed this morning, please do a quick review to ensure I don't leave anything out

fiunchinho (approved):

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,49 @@
...
+- Upgrade the control plane machines:
+  - Generate a new infrastructure `MachineTemplate` resource (with new image ID if needed) and create it in the Management API.
+  - Set the new Kubernetes version in `KubeadmControlPlane` resource.
+  - Change the infrastructure `MachineTemplate` on the `KubeadmControlPlane` to the new one and apply the new version of `KubeadmControlPlane`.
+- Upgrade machine pools. As default, we walk through all existing machine pools one by one:
+  - Make sure there are no MachinePools being already upgraded.
+  - Modify the infrastructure `MachinePool` resource (with new image ID) if needed.
+  - Set the new Kubernetes version in `MachinePool` resource.
+
+Also our operators watch for changes on the Cluster CR's labels to trigger an upgrade. That functionality should be kept in the new operator we build.
```
pipo02mix, line 35: I guess we need to change it if we upgrade to introduce the `ClusterUpgrade` resource, right @fiunchinho ?

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,49 @@
...
+
+This document covers only the action to upgrade the workload cluster without solving problems related to the versioning of the different CAPI controllers and/or deciding which controller reconciles a cluster.
+
+At the same time all the automation on top of the upgrades, like freeze periods or maintainance windows, are out of the scope here ([include in RFC 0015](/0015-automatic-cluster--upgrades)).
+
+Furthermore, a workload cluster can be upgraded in different ways:
+
+- Apps installed in the cluster are upgraded (out of scope here). We have other non-CAPI components that already take care of that, like app-operator
+
+- Cluster is reconciled by a newer CAPI controller. We may want to upgrade the Cluster to a newer CAPI controller release, which may change the infrastructure created to support the cluster (network, load balancers, etc). We do this update just changing the CAPI label on the Cluster CR, and we don't need a new machine image to be created/referenced by the CR's.
```
puja108, line 17: A new CAPI controller does not mean any change on anything. In theory, a controller change should never touch existing resources. We still want to do this in a controlled manner, as we want to reduce the risk of accidental changes as well as create full reproducibility between an existing cluster and a new one.

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Machine image used by control plane or workers is updated with new k8s version or new OS version. We may want to update the k8s version or the OS version running in the machines. In both cases, we need a new machine image using the updated components, and reference that image from the CR's.
```
puja108, line 19: what's with non-image-related changes?

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Context
+
+The current approach to upgrade a workload cluster in CAPI world involves a series of steps described below:
```
puja108, line 23: The approach below applies only to actual image changes. However, other changes like e.g. a Kubernetes version change are not as complex. In the mentioned K8s upgrade case it is just changing a spec field in the KubeadmControlPlane and then in each of the MachinePools/MachineDeployments.

 0016-capi-upgrades/0016-capi-upgrades.md
```diff
@@ -0,0 +1,49 @@
...
+## Open questions
+
+- Do we need to change the reference of the image ID for every upgrade?
+No there will be upgrades that only change a cluster CR field/version but does not require a new image.
+
+- For `MachinePool`, do we need to make the operation in a transactional way to avoid multiple upgrade (in case we need to change infrastructure and agnostic `MachinePool`)?
+No, there is no a use case where two resources has to be modified at same time. But the changes done to the `MachinePool` must be done all at once (obviously :)).
+
+- Are the provider controllers aware of KCP upgrades and wait till they are done before start upgrading Machine Pools?
+There could be provider controllers that check version of control plane before upgrading `MachinePools` but we are not relying on it and the operator we will implement will take care of the correct sequence.
```
puja108, line 46: Additiopnaly version drift between CP and workers is a supported feature upstream and can be used e.g. to patch CVEs on CP quickly without having to impact workers at all.

 0016-capi-upgrades/0016-capi-upgrades.md
puja108, line 35: IMHO a label update should NOT necessarily lead to an actual resource upgrade. For example if a new version only introduces a change in a CAPI controller version, that does not trigger any real changes, it is just a handover between two controllers and ideally should not affect anything.

 0016-capi-upgrades/0016-capi-upgrades.md
pipo02mix: but do we know if you introduce a new controller that differs how infra is configured, the controller will reconcile the existing resources to be like the new controller logic? cc/ @fiunchinho

 0016-capi-upgrades/0016-capi-upgrades.md
pipo02mix: can you elaborate? I dont get it

 0016-capi-upgrades/0016-capi-upgrades.md
pipo02mix: but kubectl and other components baked in the image needs to be updated too, right?

 0016-capi-upgrades/0016-capi-upgrades.md
pipo02mix, line 35:
```suggestion
The upgrade operator watches for changes on the Cluster CR's labels to trigger an upgrade. In case the release does not bring an OS or k8s change the upgrade operator will not trigger the process.
```

 0016-capi-upgrades/0016-capi-upgrades.md
puja108: the thing is that with k8s version changes you do not need to touch anything but the MachineTemplate, at least when I tried it I only needed to change the version field in CP and MachineDeployments, no changing of machine templates, so I'm not sure if that step is needed

 0016-capi-upgrades/0016-capi-upgrades.md
puja108: e.g. with config changes?

 0016-capi-upgrades/0016-capi-upgrades.md
puja108: IMHO we should do it like upstream and try to not change infra just by having a new controller version, but do it more specified

teemow:
Please make sure to pull the changes before making new modifications. I've renamed the files as we agreed on a slightly different structure for RFCs.

JosephSalisbury:
@pipo02mix what do you want to do with this RFC? do we have consensus to merge?

pipo02mix:
I am tend to think we can close it, @fiunchinho has brought [a new PR](https://github.com/giantswarm/rfc/pull/20/files) which is more complete and it covers the whole capi process, WDYT Jose?

fiunchinho:
Yeah I'd close this one. Thanks everyone.

PR #20, opened 2021-05-05
CAPI adoption approach regarding upgrades

fiunchinho:
Today we have a kaas sync meeting. Please, wait until that meeting (or you watch the recording) before commenting, as we are planing on explaining this during the sync.


capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Versioning of CRDs
+
+Cluster API is still in a very early stage of development and its CRD versions change once or twice a year. Kubernetes api-server can only manage one CRD version at a time as a storage version and backwards and upwards support for alternative versions must be implemented by appropriate webhooks. Depending on the change in a CRD structure these webhooks work or not.
+
+Currently the stable API version for Cluster API is v1alpha3 but v1alpha4 is already well under development. It is highly possible that when we get our first production ready version of GS managed Cluster API released, we soon must migrate to the next one involving CRD migration. This can become very problematic when the management cluster has a large number of active releases that contain components that utilize different versions of CRDs.
```
MarcelMue, line 37: I don't understand this one. This seems to be in relation to the conversion webhook?

Maybe I don't get it but wouldn't this be a problem in any configuration which should be solved by the conversion webhook?

capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+This is also more in line with native K8s controllers, where a new Deployment controller does not roll all deployments for example. It might offer new opt-in features that you can use, but you have to do that explicitly.
+
+### Immutable release of all components
+
+As briefly mentioned earlier in this document, the Giant Swarm releases are currently strictly defining all specific components that are either needed to create a workload cluster or that are running on the workload cluster. This includes operators running on management cluster, used Flatcar image version, installed Kubernetes version and several apps for both management cluster and workload cluster.
+
+When we move to fully use upstream Cluster API controllers to manage workload clusters, the whole concept of versioning changes radically. With the Cluster API CRDs and the upstream controllers, the end-user gets a much more flexible way to create, configure and manage workload clusters. At the same time the mechanisms how different components are configured, changes dramatically.
```
MarcelMue, line 50: I think this is the crux of the whole discussion:
We talk about how Cluster API manages the clusters and infrastructure but at least to me it's not clear what the "upstream" way is we describe here.
There are several different ways to run CAPI: From what I gathered this document aims at the "replace all controllers" in place and always keep them on latest style.

capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Another problem with label based versioning is that because all code in the upstream ecosystem has been built on the idea of having a single version of an operator on a given cluster, there are several optimizations that leverage this fact. For example, kubebuilder (a widely used framework) has internal re-queuing logic that will repeat reconciliation of a given CR at certain states and this bypasses completely any label filtering. Whether it is built as a label selector for watch operation or post-processing predicate filter as is the current situation with our forked controllers. This has a consequence that change of version label on any given CR does not guarantee that the change is immediately reflected on which operators are reconciling it. In fact, there would often be multiple of them which is very risky.
```
MarcelMue, line 25: I would like to know the technical details of this. Independent of this proposal.

capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Instead of strictly trying to keep control of release components, we should have sane defaults for CR templating. This would be our way to indicate “we have tested these components in this combination and found them to be rock solid, use them”.
+Focusing on templating with sane defaults rather than enforcing certain values would also mean getting rid of admission controllers that we currently use to validate and default CR content, simplifying the Kubernetes configuration.
+
+Allowing customers to customize those sane defaults would ease the testing of new combinations. For example, since upgrading Kubernetes to a new version would be only a matter of changing the required CRs, customers can test their workloads running on the new Kubernetes version on the cluster they decide at their own pace. And they would be testing that change in isolation. There would be no changes in the infrastructure resources or the installed apps.
+
+### Change our mindset when implementing operators
```
MarcelMue, line 92: I like this in general but I have some thoughts:
Currently we write opinionated infrastructure as code - meaning that we do changes to infrastructure through our operator code where the "opinion" is hardcoded in our operators.
Changing our mindset as described here would mean that we have more of these opinionated decisions in configuration through CRs.

I believe that CAPI providers also have a certain opinionated infra decisions in code. Other parts are then configurable through CRs. From what I see it leads to changes being bundled into minors - a lot of changes are being bundled into `v1alpha4` for example because the opinionated infra can not change without breaking changes in the CRDs.

At the same time too much configurability through the CR could make them hard to understand or hard to use.

To me this is about a certain balance between these approaches because either extreme has drawbacks in my mind.

capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+In order to comply with upstream operator behavior, we must change our mindset when implementing operators. Instead of writing logic that proactively changes reconciled resources immediately when the code is executed, we should adopt the approach where only change in the CRs can cause an action on the target resource.
+
+### Run multiple management clusters when needed
```
JosephSalisbury, line 96: yeah, this is the rough part here. I totally get that running multiple MCs is the more CAPI way, but i feel like it'd be a **ton** of work for us to get there

As a rough + incomplete dump of challenges / questions / thoughts:

- what do we do if customers end up sprawled over multiple releases? do we maintain a growing number of MCs? how does this affect cost?
- how do we manage DNS? do we encode version in there?
- how do users discover all of the MCs in an installation? would we attempt to make the multiple MCs transparent to the user?
- how do we migrate current MCs?
- how do we manage prometheus data if we move WCs between different MCs? do we have a separate MC for monitoring?
- what changes do we need to make to our tooling if we have more than one MC per installation?
- what would we do on KVM?
- how do we orchestrate upgrades? what component is moving CRs between MCs?

this totally feels like i'm just dumping a lot of problems and i can only apologise for that, but this would be a big ol' change (and i guess in a way, you requested some comments? :D)

JosephSalisbury (commented):
this is cool to read - can we have a legit title? this one comes up in my email weird (although i get why you wanted to push people towards the description, sure)

capi-adoption/README.md
JosephSalisbury: if i had to sum this up - are we at risk of turning a (albeit difficult) somewhat-tractable / semi-solved operator management problem into an intractable / unsolved cluster management problem?

capi-adoption/README.md
tuommaki: As far as I can read code, the predicate filtering is only done on event entry (i.e. when receiving event from informer). 

When re-queuing event, it bypasses this filtering:
- Main event queue processing loop: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L212-L215
- That then takes an event from queue and calls reconciliation on it: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L233-L255

It would require extremely careful programming of operator to have 100% confidence that no execution path causes re-queuing of event, while still taking advantage of that optimization feature in general. ...and despite of this, I'd say there would still be a bug hidden somewhere.

Because of this it is impossible to have a clear cut transfer of reconciliation on a given CR with the label change.

capi-adoption/README.md
tuommaki: The difference between our multi-version situation and upstream situation is following:

When performing in-place upgrade for single set of controllers, involving a CRD upgrade, the process is roughly following:
1. Install conversion webhooks (that provide ability to convert _old -> new_).
2. Update CRD so that old version is still used as a storage version, but new version is served (w/ conversion webhook).
3. "Lock" all CRs from modifications (with simple validating webhook which prevents any changes to CR when given _locked_ annotation is set).
4. Update all controllers. They now read the _old_ CRs as _new_ via conversion webhook.
5. Switch CRD storage version to new one. All CRs are still stored as _old_ because they don't get any changes due to validating webhook.
6. Start removing locks one-by-one from CRs, in order to have controlled upgrade of CRs into new version.

This is one-way train, with a brake lever in case something unexpected would happen. This is also the direction that upstream supports and provides support for.

In our multi-version situation we would need to have conversion webhooks both ways. In case any field gets removed in new version of CRD, we couldn't switch to use it because old operators can't persist the information. Similarly, if we  don't switch to use new version, the added new fields cannot be stored in the CRs and new operators won't work.

Two way conversions when only one can be used as storage are impossible with different sets of changes (and we don't control these changes).

capi-adoption/README.md
tuommaki: 1. Yes, we replace all controllers in place and keep them in latest stable releases. In other words: We always update to latest patches on given MC.
2. The versioning changes radically because workload cluster kubernetes version, etcd version or given in-cluster DNS application / its version are not coupled with the version of operators. This is extremely important improvement because now customers can test different kubernetes versions without getting changes to infrastructure at the same time.

capi-adoption/README.md
tuommaki: I would really switch _minor_ into _major_ in our terminologies here because that's what they really are in kubernetes ecosystem. I don't consider it as an issue that there are changes when major version jumps. Quite contrary, they are expected and therefore we need to behave accordingly. The patch updates then bring occasionally small improvements and most of the time fix bugs.

Regarding configurability, I only have good experience from CAPZ at the moment, but the opinionated infra decisions are quite small implementation details. In my experience the CAPZ controllers have been made to be very flexible and for most parts of the infrastructure you can actually "bring your own" into the setup. This allows users to create very different looking clusters on same installation, but at the same time, this also makes many of the silent modifications in infrastructure explicit.

We have quite a bit of experience from some users silently modifying the infrastructure of reconciled clusters and with the Cluster API / CAPZ approach many of these modifications can be performed via CR configuration and therefore made explicit. We could embrace this and build a tool which can show a clear diff between our defaulted infrastructure and currently configured one, in order to more easily understand what's going on when something doesn't work as expected.

capi-adoption/README.md
tuommaki: > what do we do if customers end up sprawled over multiple releases? do we maintain a growing number of MCs? how does this affect cost?

Everything here depends on how we manage those MCs. I personally don't understand why we don't utilize cloud providers manages k8s for MC when we don't need anything fancy there. This would reduce costs in general and most probably improve reliability. Another thing is how we consider monitoring to be implemented. In some discussions it has been seen as a good thing to decouple it from MC into own cluster, but this would require further thinking and discussion.

> how do we manage DNS? do we encode version in there?

This is something to think about. I think we are currently way too tightly coupled into DNS and it has been a blocker at times (e.g. inter-region MCs).

> how do users discover all of the MCs in an installation? would we attempt to make the multiple MCs transparent to the user?

When general expectation is to have only one MC, I wouldn't complicate this too much and therefore I'd just make customers to know which MC they use. i.e. no transparent cluster discovery for now.

> how do we migrate current MCs?

We install CAPx components with entry-label filtering (i.e. there needs to be distinction between GS & CAPx CRs and we can use watch filter for this). We also complete the defaulting in our templating tools and that's pretty much it. Once the last legacy GS cluster is upgraded to CAPx, then we drop all old releases and along that all old services.

> how do we manage prometheus data if we move WCs between different MCs? do we have a separate MC for monitoring?

This is subject for separate discussion. There would be benefits in separating monitoring from MCs because that would make MCs themselves a lot more agile, but downside is that reliable cluster discovery becomes slightly more difficult, but not impossible though.

> what changes do we need to make to our tooling if we have more than one MC per installation?

Depends on how much of our legacy tooling will live after we finally get into "upstream land". I wouldn't think about this too much until we get the bigger discussion solved.

> what would we do on KVM?

Depends very much on what's the future of KVM. I have no visibility on what's happening there at the moment, but if we continue with GS specific approach that `kvm-operator` has been built on, I'd say this is something for Rocket to figure out.

The VMWare approach then should align pretty well since what I've understood, that's pretty much Cluster API compliant as well. In CAPI the question whether some cluster is MC or WC is not really a thing when compared to out legacy systems where those are completely different in almost every aspect.

> how do we orchestrate upgrades? what component is moving CRs between MCs?

Most probably we need to write our own tool for this, so that we can move GS operator specific ones as well. It's not really rocket science though since I already made one PoC tool for Azure earlier this year (when I was working on migration before hive sprint).

> are we at risk of turning a (albeit difficult) somewhat-tractable / semi-solved operator management problem into an intractable / unsolved cluster management problem?

I don't think so. I somehow thought that we were already working towards having fully automated MC creation and on that bases this would just move along with that. Even in worst case scenario I can only see two MCs running at a time for any given customer. Having more than that would mean they are years behind (at least with current CAPI major bump frequency).

capi-adoption/README.md
MarcelMue: I don't quite follow this reasoning. AFAIK cluster-api tests `hub-spoke-hub` so everything I reviewed or was part of implementing (e.g. the ignition changes) needed to have conversion in both directions - from what I remember.

Sure we shouldn't have different versions of operators reconcile the same CR but with `preserveUnknownFields` enabled on the CRD I also don't see how we could lose data. Especially given the fact that `hub-spoke-hub` seems to be one of the things actually tested for.

Ref: See the conversion implemented by Mateusz here which includes both directions https://github.com/kubernetes-sigs/cluster-api/pull/4172/files#diff-65d83b423d8654f175d640d1605e28e0210e47717ea227b4ea1bd83a8800f211R48

capi-adoption/README.md
MarcelMue: 1. I get that but it's not clearly stated in the RFC doc so far. See also @JosephSalisbury confusion further down as I got the impression that he was talking about the whole "one operator version per MC and then move CRs between MCs approach".

2. I would be questioning the limitations here. Upstream was recently discussing the range of k8s versions they support (generally the last 3 IIRC). How do we convey this to the customer without having a relation between these versions (e.g. a release)? Similarly the problem exists with the new latest kubernetes versions. CAPI AFAIK is not immediately compatible with a new version but did need _some_ (not much) time to bring out a new patch to support it.
We should think about how we convey this and structure this. I agree that it needs to be build into defaulting / validation but it will still be a problem.

capi-adoption/README.md
tuommaki: > with `preserveUnknownFields` enabled on the CRD I also don't see how we could lose data.

I wasn't aware of this behavior (but based my understanding on earlier observations). With that feature the mix&match approach could theoretically work.

capi-adoption/README.md
tuommaki: Regarding limits or supported range of versions for point 2., I consider it rather as a matter of communication than a technical issue. When we don't strictly control the support but just provide what upstream provides, I see it similar to kubernetes features in general: Customers need to educate themselves in order to know what is possible with given versions.

capi-adoption/README.md
```diff
@@ -0,0 +1,114 @@
+# Road to Cluster API (over the potholes)
+
+## About
+
+This is a document to present some of the fundamental problems we have faced when trying to adapt Giant Swarm release management and operator reconciliation practices to upstream Cluster API controllers. We present the way upstream approaches it and provide a proposal to change our direction into an upstream compatible model.
+
+## Present state of Giant Swarm platform
+
+The present state of Giant Swarm platform consists of a single management cluster on each installation. All workload clusters are managed from the same management cluster on a given installation.
```
tfussell, line 9:
```suggestion
The present state of Giant Swarm platform consists of a single management cluster for each installation. All workload clusters are managed by the same management cluster in a given installation.
```

capi-adoption/README.md
tfussell: You can't use `preserveUnknownFields` with v1.CustomResourceDefinition, only v1beta1. For v1, you can add `x-kubernetes-preserve-unknown-fields: true` to the root property of the schema with `type: object` and it will apply to the object and all of its children.

I can't remember the specifics, but Kubernetes does inject an annotation containing the full JSON representation of the resource before it was converted in certain cases. I believe this is intended to avoid data loss in the case of a changed schema.

capi-adoption/README.md
fiunchinho: > As a rough + incomplete dump of challenges / questions / thoughts:

Creating a new MC for a new major release is thought for testing only, we are not advocating moving WC's from one MC to another. So while @tuommaki answers are valid, I don't think we would face that scenario regularly, if ever.

fiunchinho:
This is the recording of the last sync https://drive.google.com/file/d/1st4IT4CT5rjqZEkjKLiCgOwa58vusazf/view?usp=sharing

capi-adoption/README.md
tuommaki: > are you thinking of using multiple MCs only for Giant Swarm testing (at least initially)?

Not only Giant Swarm testing, but potentially also customer testing in case we or they would see a need for it. I could imagine some of our bigger customers to benefit from another throwaway MC where they can test & verify the major upgrade before executing it on the primary MC.

In case we need to test or reproduce results of a past version of controllers, this would also need a disposable throwaway MC.

If there are strict blockers or critical isolation needs for some environment pipelines for certain customers, I could see multiple MCs as a solution to provide the added granularity & safety guards, but this must not be a norm.

fiunchinho:
A new appendix has been added to the RFC trying to address some of the concerns.

JosephSalisbury:
@fiunchinho what's stopping us from merging here? after our last kaas sync, i feel like we have general consensus on the approach (which this RFC describes)

capi-adoption/README.md
```diff
@@ -0,0 +1,159 @@
...
+Focusing on templating with sane defaults rather than enforcing certain values would also mean getting rid of admission controllers that we currently use to validate and default CR content, simplifying the Kubernetes configuration.
+
+Allowing customers to customize those sane defaults would ease the testing of new combinations. For example, since upgrading Kubernetes to a new version would be only a matter of changing the required CRs, customers can test their workloads running on the new Kubernetes version on the cluster they decide at their own pace. And they would be testing that change in isolation. There would be no changes in the infrastructure resources or the installed apps.
+
+### Change our mindset when implementing operators
+
+In order to comply with upstream operator behavior, we must change our mindset when implementing operators. Instead of writing logic that proactively changes reconciled resources immediately when the code is executed, we should adopt the approach where only change in the CRs can cause an action on the target resource.
```
tfussell, line 94: Can you give an example of this? I thought we preferred level-based reconciliation rather than edge-based. If a controller is down for some reason and misses the change event, how could it eventually end up in the correct state?

capi-adoption/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Instead of strictly trying to keep control of release components, we should have sane defaults for CR templating. This would be our way to indicate “we have tested these components in this combination and found them to be rock solid, use them”.
+Focusing on templating with sane defaults rather than enforcing certain values would also mean getting rid of admission controllers that we currently use to validate and default CR content, simplifying the Kubernetes configuration.
+
+Allowing customers to customize those sane defaults would ease the testing of new combinations. For example, since upgrading Kubernetes to a new version would be only a matter of changing the required CRs, customers can test their workloads running on the new Kubernetes version on the cluster they decide at their own pace. And they would be testing that change in isolation. There would be no changes in the infrastructure resources or the installed apps.
```
tfussell, line 90: What would upgrading Kubernetes look like? Changing `.spec.version` in the `KubeadmControlPlane` CR? I feel like we would need a fairly robust e2e matrix test covering all of the combinations of Kubernetes version, app versions, and infrastructure versions if we're going to provide this level of freedom.

tfussell:
Just thinking out loud here--it seems to me that the end result of this would be having a single "MC release" that declares the version of each controller/service needed to create a WC. Once we test new versions of components in isolation and all together on a test MC, we would package it up and deploy it to our production MCs immediately or during a defined maintenance window. This "MC release" might also include one-off migration steps like updating CRDs, upgrading CRs, etc. 

We could re-use `release-operator` and a `Release` CR for this, but we should also be able to do it with Argo.

We would probably want two e2e tests for a "MC release", one testing upgrades from the current production "MC release", and one testing it on a brand new cluster, converting it to an MC, deploying the MC release, and creating one or more clusters. The MC testability seems like a big win here.

capi-adoption/README.md
tuommaki: This is not so much about level-based vs. edge-based reconciliation but rather about general approach on how to determine change of target resource and how to deal with it when detected.

Examples from CAPZ:
- Several infrastructure resources are expected to be ok if they exist with certain name. See: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/azure/services/loadbalancers/loadbalancers.go#L85-L137
- If a worker node is deleted, it requires specific `MachineHealthCheck` configuration in order to be replaced automatically.

> how could it eventually end up in the correct state?

It doesn't, within patch updates. But I also don't expect changes within patch updates. In k8s world so far the minor version bump has been considered a major upgrade and this is the case with Cluster API as well. Between those bumps there could be changes to infrastructure as well, but it's fine because the whole upgrade is bigger than just bugfix patches. One can also create infrastructure migrations to be performed at this point, when necessary.

tuommaki (commented):
> it seems to me that the end result of this would be having a single "MC release" that declares the version of each controller/service needed to create a WC.

Sort of, kind of, maybe. I personally would drop the whole release thinking from MC because there's no choice with this approach. There's just the present set of programs we need to run and that's it. Releases can contain formal version number and our curated set of defaults for WC but it doesn't contain any component from MC anymore. Only relation to MC might be compatibility between used CAPI version with some range maybe - i.e. when MC is running CAPI `v1alpha4`, you can use k8s `1.20 - 1.26` or something, so that the used bootstrap operator can handle those defaults.

capi-adoption/README.md
pipo02mix: Further with the latest improvements in App Platform, we could restrict the version of apps based on k8s version. Also with `AppSets` we could wrap a collection of apps certified/tested by us. I think this step (point 2) is a huge step forward the right direction

capi-adoption/README.md
pipo02mix: IMO we should warn/restrict when potentially an upgrade can break something. For example, we have CoreDNS version X and Kubernetes version Y. If we upgrade Kubernetes to version Z and this is not compatible with CoreDNS (apiGroup change or something else) we should at least warn them. With App Platform and new functionality, we can add metadata to define compatibility between components and leverage that to offer a nice UX. Most of the customers trust our release bundling and it gives them peace of mind to not have to choose versions in core components. We should preserve that even we allow expert/advance users work with low level layer (capi CRs)

stone-z:
General feedback:
I am open to the change in mentality around updating MC components. We will likely benefit from keeping pace with upstream, and at least philosophically should be able to iterate as fast as we want on MCs.  I do have the following concerns, please let me know if I'm just missing something:
- "upstream won't break stuff" is an assumption that seems pretty risky to me. We're giving up tons of control on that premise alone. Even without intentional breaking changes, bugs happen.
- moving away from a static definition of a release is problematic for me because it destroys idempotence and reproducibility. IMO, it is important for security, for testing, and for customer experience to be able to reason about what is actually running. 
  - if one year from now I want to create an old release, how do I do it? how do I know which versions of each component I should have?
  - how will upgrade paths be defined in order for them to be tested?
  - a customer needs to revert. How and to what do they revert?
  - customer upgrades for dev, staging, and production may use three different MC operator versions and are therefore not actually the same procedure. It is also no longer the upgrade path we tested for that release.

I went back and forth about writing this as it isn't immediately actionable, but I wanted to at least make those concerns known. I'm on board with going forward and figuring out the technical details as they become relevant. I would just like to see a concrete suggestion for managing those second bullet point concerns before we start running customers on clusters managed under the new ideology, otherwise I think we're going to have a bad time.

fiunchinho:
I believe no one is making the assumption that "upstream won't break stuff". We should not, upstream will definitely break things. We are not advocating for blindly deploying every `master` commit to production. We are aligning on the philosophy to try not to change things when upgrading. But we still need to upgrade to tested and validated upstream releases.

capi-adoption/README.md
fiunchinho: I don't think we are changing anything related to level-based vs edge-based reconciliation?

capi-adoption/README.md
Oshratn, line 90: Would it make sense to get some sort of feedback loop in place, so that when customers do test a certain combo and it works, we can then know this and somehow document this for the use of other customers?

Oshratn:
Towards language rather than essence...
I noticed that in the body of the text there is a lot of use of the term sane defaults, while the last section uses the (more inclusive) terms sensible defaults. Can this be made consistent across the board?

pipo02mix:
I am sharing partially the same concerns as  @stone-z 

IMO we lack experience in how upstream versioning will impact WCs, together with the fact the standard is becoming beta. My assumption is capi controllers does not modify anything without changes in the CRs, even if you update the controllers, we must verify this is true. Though immutability and reproducibility are great strengths in our system I don't think that taking the new approach should change that. In my head, I still see the release concept there, even if we change the name. Most of the customers want to have sane default so we need to offer a release/template/base/<add_your_name_here> that define a set of applications and configurations (k8s version, OS version, core components, k8s config defaults,...) that no-brainers can trust and use/upgrade without worries. BUT experienced/advanced users, as we already have, can tweak as their will without us having to bundle a new release every time.

mig4 (approved):

capi-adoption/README.md
nprokopic: Better late (comment/reply) than never :grimacing: Feedback loop as in customers giving us feedback how the testing went when they used certain combination of Kubernetes version and other component versions?

I would agree very much that we would need some sort of insight into issues that are connected to a combination of components running in a cluster (what we currently have as a "release" concept), e.g. we should be able to see (or easily discover) that running `Kubernetes 1.X.Y`, `Flatcar X.Y.Z` and `operator X v1.1.1` causes issues `A`, `B` and `C` to happen more frequently (or never).

nprokopic:
> * "upstream won't break stuff" is an assumption that seems pretty risky to me. We're giving up tons of control on that premise alone. Even without intentional breaking changes, bugs happen.

I think this is something where we can all agree, bugs happen, it will be up to us to improve our testing and contribute to the upstream around the same topic.

> * moving away from a static definition of a release is problematic for me because it destroys idempotence and reproducibility. IMO, it is important for security, for testing, and for customer experience to be able to reason about what is actually running.
>
>   * if one year from now I want to create an old release, how do I do it? how do I know which versions of each component I should have?
>   * how will upgrade paths be defined in order for them to be tested?
>   * a customer needs to revert. How and to what do they revert?
>   * customer upgrades for dev, staging, and production may use three different MC operator versions and are therefore not actually the same procedure. It is also no longer the upgrade path we tested for that release.
>
> I went back and forth about writing this as it isn't immediately actionable, but I wanted to at least make those concerns known. I'm on board with going forward and figuring out the technical details as they become relevant. I would just like to see a concrete suggestion for managing those second bullet point concerns before we start running customers on clusters managed under the new ideology, otherwise I think we're going to have a bad time.

@stone-z does this part answers your concerns to some extent?

> Instead of strictly trying to keep control of release components, we should have sane defaults for CR templating. This would be our way to indicate “we have tested these components in this combination and found them to be rock solid, use them”. Focusing on templating with sane defaults rather than enforcing certain values would also mean getting rid of admission controllers that we currently use to validate and default CR content, simplifying the Kubernetes configuration.
>
> Allowing customers to customize those sane defaults would ease the testing of new combinations. For example, since upgrading Kubernetes to a new version would be only a matter of changing the required CRs, customers can test their workloads running on the new Kubernetes version on the cluster they decide at their own pace. And they would be testing that change in isolation. There would be no changes in the infrastructure resources or the installed apps.

E.g. Kubernetes, OS and etc versions will be configurable in CRs. Since we would have those versions in the CRs, we could track how those are changing over time, e.g. there is idea to have `ClusterUpgrade` CR in https://github.com/giantswarm/rfc/pull/15 (Cluster X upgraded from A to B) that can be adapted/extended to new CAPI adoption approach, so that we always have complete information about component versions and how those changed.

stone-z:
> @stone-z does this part answers your concerns to some extent?
> 
> > Instead of strictly trying to keep control of release components, we should have sane defaults for CR templating. This would be our way to indicate “we have tested these components in this combination and found them to be rock solid, use them”. Focusing on templating with sane defaults rather than enforcing certain values would also mean getting rid of admission controllers that we currently use to validate and default CR content, simplifying the Kubernetes configuration.
> > Allowing customers to customize those sane defaults would ease the testing of new combinations. For example, since upgrading Kubernetes to a new version would be only a matter of changing the required CRs, customers can test their workloads running on the new Kubernetes version on the cluster they decide at their own pace. And they would be testing that change in isolation. There would be no changes in the infrastructure resources or the installed apps.
> 
> E.g. Kubernetes, OS and etc versions will be configurable in CRs. Since we would have those versions in the CRs, we could track how those are changing over time, e.g. there is idea to have `ClusterUpgrade` CR in #15 (Cluster X upgraded from A to B) that can be adapted/extended to new CAPI adoption approach, so that we always have complete information about component versions and how those changed.

Yeah for WC components I'm reasonably happy. In the defaulting suggestion, we effectively just move the versioning to some source of truth for defaults instead of versioning the components in `releases`. IMO, that is more convoluted than just using `releases` as the source of truth for defaulting, but for WC components I'm satisfied as long as we have written down somewhere in a reasonably immutable place which components we can expect if we create a cluster, and I can retrospectively determine what was running on a cluster after that cluster no longer exists.

I still think mixing and matching MC components is going to bite us though. Skimming the [release notes for the AWS CAPI controller](https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases) I see plenty of `Action Required` and `Breaking Change` sections in minor releases. These include things like [changes to IAM roles and security groups](https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v0.4.9) and a new base image for bastion hosts, among others.

Let's say a customer has tested an upgrade in their dev and staging clusters, and then we update our upstream controllers on their MC. When they perform the upgrade on their production clusters, the resulting cluster will be different than their other environments. Just for example, if we moved between v0.6.1 and v0.6.2, the bastion hosts for staging and production would have different operating systems. While _theoretically_ that specific change shouldn't be a problem, it definitely could be and it isn't what the customer tested or expected. Clusters upgraded at different times will have different configurations, and it would be difficult to determine retrospectively what the actual upgrade path was. If we moved from v0.4.6 to v0.4.9 there's a good chance that production upgrade would break entirely.

It has been suggested that we can drop in new upstream controllers because they're theoretically idempotent and only act when something is changed, but they're only idempotent within their own version. Clusters of the same Giant Swarm version will have different cloud configurations and in some cases different in-cluster behavior depending on which version(s) of those upstream controllers created them. And _that_ is the part I think is going to bite us if we don't also version and control them.

tfussell:
> Let's say a customer has tested an upgrade in their dev and staging clusters, and then we update our upstream controllers on their MC. When they perform the upgrade on their production clusters, the resulting cluster will be different than their other environments. Just for example, if we moved between v0.6.1 and v0.6.2, the bastion hosts for staging and production would have different operating systems. While theoretically that specific change shouldn't be a problem, it definitely could be and it isn't what the customer tested or expected. Clusters upgraded at different times will have different configurations, and it would be difficult to determine retrospectively what the actual upgrade path was. If we moved from v0.4.6 to v0.4.9 there's a good chance that production upgrade would break entirely.

I think part of the solution here could be giving customers "cycles" during which MC component versions stay consistent. If they want to test changes on dev and staging clusters before rolling out to production and expect clusters to be identical, they would need to do it while MC components aren't changing. If we only made MC changes on the first of each month, for example, they would know that they have ~4 weeks to test changes and upgrade production clusters before the next change.

puja108:
As a side note, we are moving towards automatically rolling out minors and patches even for customer WC components. We want to give this higher prio now already, as we do not want customers to get stuck on old versions.

Similarly, certain changes that were mentioned above fall into that minor/patch change category. We might however need to see how we can trigger them automatically, e.g. for a bastion host fix that only gets applied to new clusters, so we have no differences between old and new clusters with the same version.

That said, we need to test and gatekeep the whole thing to a certain extent and we will see with time what issues we face and find ways to solve them. 

I see this RFC as a good goal we want to work towards, but there might be steps and hurdles we are not 100% clear on how to solve, yet.

fiunchinho:
I think there has been enough discussion, and even we don't know all the tiny little details, we more or less agree on the direction. I'm going to merge this RFC, but feel free to open new PR's on this RFC to improve it or change it.

I want to thank you all for the time and effort you put on this.

PR #21, opened 2021-06-03
Create RFC: Bottom-up managed apps

cokiengchiara:
This RFC arose from @rossf7 's idea to offer oauth2-proxy. It’s an app, which at least 2 of our customers want (and I bet there are more), and I think we use. (See: [Slack](https://gigantic.slack.com/archives/CPC3M70UE/p1622729938024200))

### DIFF
--- a/bottom-up-managed-apps/README.md
+++ b/bottom-up-managed-apps/README.md
@@ -2,7 +2,7 @@
 
 ----
 
-This RFC arose from the idea to offer oauth2-proxy. It’s an app, which at least 2 of our customers really want (and I bet there are more), and I think we use. (See: [Slack](https://gigantic.slack.com/archives/CPC3M70UE/p1622729938024200))
+This RFC arose from the idea to offer oauth2-proxy. It’s an app, which at least 2 of our customers  want (and I bet there are more), and I think we use. (See: [Slack](https://gigantic.slack.com/archives/CPC3M70UE/p1622729938024200))
 
 **Proposal: Offer managed apps, bottom-up, instead of top-down. Get back to our ‘original intention'.**
 
--- a/bottom-up-managed-apps/README.md
+++ b/bottom-up-managed-apps/README.md
@@ -4,7 +4,7 @@
 
 This RFC arose from the idea to offer oauth2-proxy. It’s an app, which at least 2 of our customers really want (and I bet there are more), and I think we use. (See: [Slack](https://gigantic.slack.com/archives/CPC3M70UE/p1622729938024200))
 
-**Proposal: Offer managed apps, bottoms up, instead of top down. Get back to our ‘original intention'.**
+**Proposal: Offer managed apps, bottom-up, instead of top-down. Get back to our ‘original intention'.**
 
 Meaning:
 
--- a/bottoms-up-managed-apps/README.md
+++ b/bottoms-up-managed-apps/README.md
@@ -1,0 +1,53 @@
+**tl;dr Let’s offer only apps that we actually use as managed apps (of course, if customer also uses it or wants to use it), such as prometheus, oauth2-proxy, etc. and NOT: Linkerd, Harbor, etc. stuff on roadmap we don't actually use. This way, a small managed apps team can more effectively offer more managed apps for less effort, while retaining the same business and user value, which is reduce toil via managed apps. WDYT?**
+
+----
+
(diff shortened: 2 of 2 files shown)


cokiengchiara:
I won't be hurt if people tell me this is super stupid 😀

rossf7:
From my POV I think it's a good direction. Although some of the more complex apps have come from requirements from new / existing customers. So that's also a factor IMO.

Once the new Catalog CRD is ready we will create docs for how customers can create their own catalogs.
https://github.com/giantswarm/roadmap/issues/154 

I think we should include https://artifacthub.io/ in the docs as thats my goto source for finding Helm charts these days.

We need to make sure these apps are visible in our dashboards. We can use that data to discover new managed apps we could offer.

bottom-up-managed-apps/README.md
```diff
@@ -0,0 +1,53 @@
...
+3. Prioritize those as managed apps
+4. Grow our offering organically this way.
+
+This is in contrast to the apps in our roadmap. That is, observabilty → developer tooling -> use cases (Kafka, etc.)... most of which are apps we don't already use internally, or merely feel we ‘should’ use but actually don't.
+
+**Status Quo:**
+
+Assumption: The basis on which the managed apps team was created is this idea:
+
+> We use these apps already. Why not help customers with them as well? It won’t take much more effort.
```
puja108, line 22: first apps might have started that way, that's also why we first wanted to prioritize prometheus (operator) as we had experience with prometheus already, but it was quite clear from the beginning that customers need more than what we need. That said starting with what we already have is a good idea as it's a quick win.

bottom-up-managed-apps/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Risk: Giant Swarm is not representative of our customers. So what we use and would dogfood would not apply to what our customers use.
+- Cost: Not as "sexy” as ✨ shiny ✨ new ✨ apps ✨
+
+**Next Steps / Questions:**
+
+- What are some apps that (a) giant swarm uses and (b) at least 2 customers use?
+   - prometheus
+   - oauth2-proxy
+   - …?
```
puja108, line 49: tbh, i don't think there's many moreright now

bottom-up-managed-apps/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+**Top 3 Benefits**
+
+- **Increased speed - of adding managed apps, fixing issues, responding to support. Everything.** — The bottleneck of providing managed apps is having to learn a totally new app. Both technical (have to read up on documentation every step of the way) and psychological (uncertainty; fear of the unknown and not being an expert)
+- **All the benefits of dogfooding**, most important of which are (1) improved product quality, (2) better product knowledge and awareness from first hand use, and (3) truly seeing other Giant Swarm engineers as customers, (4) we can more authentically evangalize this product because we chose to use it as well, (5) Better support since more GS engineers can offer expertise
+- **Ability to manage more apps with current (small) number of people** — Because of the top two benefits above.
+
+**Risks and Costs**
+
+- Risk: Giant Swarm is not representative of our customers. So what we use and would dogfood would not apply to what our customers use.
```
puja108, line 41: Exactly, we are not our customers, our needs are often a bit different as we for example do not run 100s of services with lots of heterogeneous teams and our services are very homogeneous and most of what we do is open source so our requirements differ a lot. Even on the Kubernetes user side I would go out on a limb and say most mature customers at some point are better at "using" Kubernetes than many of our internal engineers, as our own usage is very limited.

bottom-up-managed-apps/README.md
```diff
@@ -0,0 +1,53 @@
...
+- Cost: Not as "sexy” as ✨ shiny ✨ new ✨ apps ✨
```
puja108, line 42: I wouldn't call it sexy, but there's a coherent "story" behind providing "the cloud native stack(s)" that customers need to get going. Like, people need observability, and they also need a developer pipeline and tooling, and they will need security, and data stores. I'm fine with evaluating the actual need behind certain parts of that stack and if we should provide them, but I do think that a coherent story behind a product offering is valuable beyond just sales and marketing.

puja108 (commented):
TLDRing my comments
+1 on getting quick-win through apps we already use and can easier provide
-1 on that being a strategy that holds very long as there's not many significant apps like that and the product story is kinda missing

teemow:
I disagree with this direction. We should not focus on the apps we use ourselves at all. We should add the apps that enable our customers primarily. Our own teams that need apps in the catalog will add them anyway.

Finding time and priority for us to dogfood eg linkerd or loki is a different story. But Halo should focus on customer needs imo.

cokiengchiara:
thanks for the comments! agree.

PR #22, opened 2021-07-05
RFC: CAPI for KVM

tfussell:
For https://github.com/giantswarm/giantswarm/issues/17573

Copied from https://hackmd.io/0TL8_gPVSS6T2O5KZq5q0A?both

### DIFF
--- a/rfc-capi-for-kvm-capk/README.md
+++ b/rfc-capi-for-kvm-capk/README.md
@@ -1,55 +1,71 @@
-# RFC: CAPI for KVM (CAPK)
+# RFC: Cluster API for KVM (CAPK)
 
-### Use cases
+## Glossary
+The lexicon used in this document is described in more detail
+[here](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/reference/glossary.md).
+Any discrepancies should be rectified in the main Cluster API glossary.
 
-- As a Giant Swarm (GS) Platform Engineer (PE), I want to implement Cluster API (CAPI) for KVM (CAPK) clusters so that I
-  don't need to maintain legacy services and libraries (e.g. `k8scloudconfig`, `apiextensions`, `cert-operator`,
-  `cluster-service`) for the foreseeable future.
-- As a GS PE, I want the operator(s) reconciling KVM clusters (collectively referred to as "KVM operators") to work 
-  independently of other GS components so I can add features more quickly and make usage by the wider community more
-  feasible/possible.
-- As a GS PE, I want KVM operators to work in arbitrary Kubernetes clusters so we can stop maintaining our custom
-  provisioning tooling for on-prem management clusters.
-- As a GS employee, I want KVM operators to implement the CAPI provider contract so that we can be in the list of
-  CAPI providers to improve the reputation of our company.
-- As a GS on-premises (on-prem) customer, I want KVM operator to implement CAPI so that I have access to new
-  provider-independent features developed by other GS teams.
-- As a GS oncall engineer, I want on-prem installations to be as similar as possible to the other providers so that I
-  can share best practices, on call routines, and documentation with the other teams.
-- As a GS on-prem customer, I want GS to implement a migration strategy from current kvm clusters to CAPI KVM clusters
-  so that I can access the new features without service disruption and workload migration.
-- As a GS on-prem customer, I want GS to implement a break out/migration strategy for on-prem installations so that I
-  can use other CAPI on-prem providers in the future such as CAPV (Cluster API for VSphere (VMWare)) or metal3 (Cluster
-  API for bare metal).
-- As a member of Team Rocket, I would like to have confidence in my changes to the platform through the use of reliable
-  unit, integration, and end-to-end tests.
+- **CAPK** - Cluster API infrastructure provider for KVM
+- **on-prem** - Short for on-premises, hardware running on the premises of an enterprise rather than in a public cloud.
+- **Management API** - The Kubernetes API of the management cluster (sometimes abbreviated as MAPI)
+- **KVM Operators** - The operators that reconcile KVM cluster resources. Currently `kvm-operator` and `cluster-operator`.
 
-
-### Rationale
+## Motivation
 
 With the current move to full CAPI adoption in Team Firecracker and Team Celestial, Team Rocket would be the only user
 and therefore the only maintainer of the current components and code that would be legacy for the other teams.
 
 This would mean that we would need to put a lot of effort into understanding and owning this code fully in order to be
-able to fix bugs and maintain KVM installations
-in the future.
+able to fix bugs and maintain KVM installations in the future.
 
 Instead of being the sole maintainer of a huge amount of legacy code, Team Rocket wants to invest this time in
 implementing CAPI for KVM operator and maintain relative platform parity with the other GS provider teams.
 
+This also brings obvious advantages for customers to benefit from developements and automation provided by Cluster API
+as elaborated upon in the use cases below.
+
+### User Stories
+
+#### External
+
+- As a GS on-prem customer, I want to use the Management API to manage clusters and applications via GitOps.
+- As a GS on-prem customer, I want to use the Management API to have more fine grained control over permissions and
+  roles of API users.
+- As a GS on-prem customer, I want would like for the KVM platform to use Cluster API so that I have access to new
+  provider-independent features developed by other GS teams.
+- As a GS on-prem customer, I want GS to implement a migration strategy from current KVM clusters to CAPK clusters
+  so that I can access the new features without service disruption and workload migration.
+
+#### Internal
+
+- As a GS Platform Engineer (PE), I want to implement CAPI for KVM clusters so that I don't need to maintain legacy
+  services and libraries (e.g. `k8scloudconfig`, `apiextensions`, `cert-operator`, `cluster-service`) for the
+  foreseeable future.
+- As a GS PE, I want the operators reconciling KVM clusters to work independently of other GS components so I can
+  add features more quickly and make usage by the wider community more feasible/possible.
+- As a GS PE, I want KVM operators to work in arbitrary Kubernetes clusters so we can stop maintaining our custom
+  provisioning tooling for on-prem management clusters.
+- As a GS employee, I want KVM operators to implement the CAPI provider contract so that we can be in the list of
+  CAPI providers to improve the visibility of our company.
+- As a member of Team Rocket, I would like to have confidence in my changes to the platform through the use of reliable
+  unit, integration, and end-to-end tests.
+- As a GS oncall engineer, I want on-prem installations to be as similar as possible to the other providers so that I
+  can share best practices, on call routines, and documentation with other teams.
+
 ### Goals
 
-- Customers can upgrade existing clusters to CAPK without significant workload downtime.
-- Existing CAPI-compatible tools such as kubectl-gs, happa, GS api, and clusterctl work with CAPK clusters.
+- Customers can upgrade existing KVM clusters to CAPI without significant workload downtime.
+- Existing CAPI-compatible tools such as `kubectl-gs`, `happa`, GS `api`, and `clusterctl` work with CAPK clusters.
 - Team Rocket is responsible for a significantly reduced number of repos and lines of code.
 
-### Non-goals
+### Non-goals/Future Work
 
-- Get CAPK accepted as a CNCF project.
+- Get CAPI for KVM accepted as a CNCF project.
 - Support mixed CAPK/CAPV clusters.
 - Support mixed management clusters (e.g. CAPK on AWS).
-- Support `MachineSet` or `MachinePool`.
-- Support CAPI <=v1alpha3.
+- Support high availability control planes (might "just work").
+- Support node pools (might "just work").
+- Support CAPI versions before `v1alpha4`.
 
 ### Scope
 
@@ -59,8 +75,7 @@ new operator and those we must ensure are handled by CAPI as follows:
 
 #### Features to migrate to new operator
 
-- DNS and NTP configuration for WC nodes (OS level)
-- iSCSI initiator name (control plane nodes)
+- DNS and NTP configuration for WC nodes (OS level config)
 - host volumes (worker nodes)
 - etcd storage (host path/persistent)
 - cpu and memory limits for control plane and worker node pods and QEMU VMs
@@ -89,7 +104,7 @@ The WC Kubernetes API is reachable externally (e.g., for GS engineers, customers
 `https://<cluster id>.k8s.<installation base domain>`. This DNS name should resolve to the external IP of the MC ingress
 controller.
 
-Inside the WC, control plane nodes should connect to etcd and k8s API via localhost. Workers should use a control plane
+Inside the WC, control plane nodes should connect to etcd and k8s API via `localhost`. Workers should use a control plane
 endpoint of `https://<cluster id>.k8s.<installation base domain>` or `control-plane.<cluster id>.svc` depending on
 whether external DNS or MC CoreDNS is used for WC node DNS resolution (respectively). This will be determined during
 implementation.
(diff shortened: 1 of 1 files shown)


rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+- Support mixed CAPK/CAPV clusters.
+- Support mixed management clusters (e.g. CAPK on AWS).
+- Support `MachineSet` or `MachinePool`.
+- Support CAPI <=v1alpha3.
+
+### Scope
+
+To minimize the scope, we will start with a minimal feature set and add only features that are used by the current customers. Features that are currently not used by any customer will be dropped.
+
+Current features of `kvm-operator` (with a ✅ if we decided to keep it in CAPI):
```
MarcelMue, line 43: This is a bit confusing - most of these which are related to the api-server / kubelet might still be covered by the general CAPI controllers IMO.

So would the feature just move?

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+	- [x] ssh user list
+- [ ] iscsi initiator name (masters)
+- [ ] configure storage type (host path/persistent)
+- [x] master and worker node pods: cpu and memory limits,
+- [x] size of kubelet and docker volume
+- [x] automatically delete TC node pods when NotReady for too long
+- [x] dead endpoints: don't route traffic to non-ready nodes
+- [x] vertical pod autoscaler for operator pod
+
+#### Features to drop
```
MarcelMue, line 72: Okay I believe this is what I meant in the earlier comment. So adding comments above if they are covered by CAPI in a single list would be nice.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+#### DNS
+
+The WC Kubernetes API is reachable externally (i.e. for GS engineers or customers) via an ingress in the MC of the form `https://<cluster id>.k8s.<installation base domain>`. This DNS name should resolve to the MC ingress IP.
+
+Inside the WC, control plane nodes should connect to etcd and k8s API via localhost. Workers should use a control plane endpoint of `https://<cluster id>.k8s.<installation base domain>` or `control-plane.<cluster id>.svc` depending on whether external DNS or MC CoreDNS is used for WC node DNS resolution (respectively). This will be determined during implementation.
+
+#### Machine OS Images
+
+The CAPI-native method for OS images is using an image builder to create a different image for each Kubernetes version. This implies a lot of complexity and is a big departure from our current approach of configuring nodes using ignition. During development we will use kubeadm pre- and post- commands to provision nodes and then decide if we want to go through the process of setting up an image building pipeline.
```
MarcelMue, line 95: Images built by image builder should be available soon (for flatcar as well) - so I don't think this would be a massive change for us? Maybe I am wrong.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+##### Features in operatorkit
+
+1. expose `creation_timestamp`, `deletion_timestamp` and `last_reconciled` metrics for every object we reconcile over. We use them in the [following](https://github.com/giantswarm/prometheus-rules/blob/8142cac5f47be117ca428fc422d71afd2db3f5ee/helm/prometheus-rules/templates/alerting-rules/operatorkit.rules.yml#L1) alerts.
+1. create [kubernetes events](https://github.com/giantswarm/operatorkit/blob/master/docs/using_kubernetes_events.md) on reconcilation errors
+1. sentry client: (**not used in `kvm-operator`**)
+1. [pause reconciliation](https://github.com/giantswarm/operatorkit/blob/master/docs/pause_reconciliation.md): do not reconcile if there are specific pause annotations (**not used in `kvm-operator`**)
+1. setting per controller finalizer (`operatorkit.giantswarm.io/<controller.Name>`)
+1. allow delete events to get [replayed](https://github.com/giantswarm/operatorkit/blob/master/docs/using_finalizers.md#control-flow) (by setting `finalizerskeptcontext.SetKept(ctx)`
+1. when we boot the server we also add two additional endpoints: `healtz` (service availability) and `version` (info about the go runtime)
```
MarcelMue, line 109:
```suggestion
1. when we boot the server we also add two additional endpoints: `healthz` (service availability) and `version` (info about the go runtime)
```

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+The CAPI-native method for OS images is using an image builder to create a different image for each Kubernetes version. This implies a lot of complexity and is a big departure from our current approach of configuring nodes using ignition. During development we will use kubeadm pre- and post- commands to provision nodes and then decide if we want to go through the process of setting up an image building pipeline.
+
+#### Controller Template (operatorkit vs kubebuilder)
+
+While we want to avoid too many moving parts, this is a good opportunity to continue aligning more closely with upstream operators and use kubebuilder. There are quite a few unknowns around this which we have summarized below:
+
+##### Features in operatorkit
+
+1. expose `creation_timestamp`, `deletion_timestamp` and `last_reconciled` metrics for every object we reconcile over. We use them in the [following](https://github.com/giantswarm/prometheus-rules/blob/8142cac5f47be117ca428fc422d71afd2db3f5ee/helm/prometheus-rules/templates/alerting-rules/operatorkit.rules.yml#L1) alerts.
```
MarcelMue, line 103: Why is each bullet `1.` here?

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+Inside the WC, control plane nodes should connect to etcd and k8s API via localhost. Workers should use a control plane endpoint of `https://<cluster id>.k8s.<installation base domain>` or `control-plane.<cluster id>.svc` depending on whether external DNS or MC CoreDNS is used for WC node DNS resolution (respectively). This will be determined during implementation.
+
+#### Machine OS Images
+
+The CAPI-native method for OS images is using an image builder to create a different image for each Kubernetes version. This implies a lot of complexity and is a big departure from our current approach of configuring nodes using ignition. During development we will use kubeadm pre- and post- commands to provision nodes and then decide if we want to go through the process of setting up an image building pipeline.
+
+#### Controller Template (operatorkit vs kubebuilder)
+
+While we want to avoid too many moving parts, this is a good opportunity to continue aligning more closely with upstream operators and use kubebuilder. There are quite a few unknowns around this which we have summarized below:
```
MarcelMue, line 99: More motivation would be also contribution to upstream.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    kind: KVMCluster
+    name: example
+```
+
+#### KVMCluster
+
+```yaml
+kind: KVMCluser
```
MarcelMue, line 161:
```suggestion
kind: KVMCluster
```

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+spec:
+  controlPlaneEndpoint:
+    host: example.k8s.geckon.gridscale.kvm.gigantic.io
+    port: 6443
+```
+
+#### KVMachineTemplate (control plane)
+
+```yaml
+kind: KVMachineTemplate
```
MarcelMue, line 174:
```suggestion
kind: KVMMachineTemplate
```

rfc-capi-for-kvm-capk/README.md
tfussell: That would be great if we don't have to maintain it ourselves. Any docs about that?

rfc-capi-for-kvm-capk/README.md
tfussell: It's a quirk of markdown. To create an ordered (i.e. numbered) list, you need to start each item with a number followed by a period. The number itself is ignored.

rfc-capi-for-kvm-capk/README.md
tfussell: I agree it's confusing. I wasn't sure how to present this information.

rfc-capi-for-kvm-capk/README.md
tfussell: Yes, most of them will go in the `KubeadmConfig` I think.

rfc-capi-for-kvm-capk/README.md
MarcelMue: There are some upstream docs but there has been internal discussion with kinvolk:
https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/README-flatcar.md

We would need to check how much effort into image building we might need to put in. But upstream does have open PRs to add ignition support:
https://github.com/kubernetes-sigs/cluster-api/pull/4172

rfc-capi-for-kvm-capk/README.md
tfussell: Good point. I have mentioned this now.

rfc-capi-for-kvm-capk/README.md
tfussell: Thanks. I added links to these in the RFC and mentioned that we would try to stay on top of what's going on there.

rfc-capi-for-kvm-capk/README.md
tfussell: I cleaned it up. Let me know what you think.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,425 @@
...
+  infrastructureTemplate:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    kind: KVMMachineTemplate
+    name: example-control-plane
+  kubeadmConfigSpec:
+    initConfiguration:
+      nodeRegistration:
+        name: '{{ ds.meta_data.hostname }}'
+        kubeletExtraArgs:
+          cloud-provider: kvm
```
cornelius-keller, line 223: Maybe this is nit picking but I think cloud-provider: kvm will not work as this works only for cloud providers that are included in the upstream tree. So I think it should be empty or `external`.

cornelius-keller (dismissed):

puja108 (commented):
I've checked this before in the hackmd and I'm cool with this. Only nitpick, I think this is already a spec and IIRC your team is already having plans working towards this, or do you consider "doing CAPI for KVM" still as sth that is not decided? (I see RFCs as topics where the decision still has to be made)

From my side you already got the ok on giving this priority. So as long as there's no other discussions going on or a veto from @teemow I think this can move to spec.

As for the future work package planning of this, I would suggest aligning with the alpha, beta, RC, GA phases that we defined for CAPI in general (your MVP and prod definitions below go in that direction)

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,449 @@
+# RFC: CAPI for KVM (CAPK)
+
+### Use cases
```
teemow, line 3: Additional use cases:

 * As a GS customer I want to use the Management API to manage clusters and applications via Gitops
 * As a GS customer I want to use the Management API to have more fine grained control over permissions and roles of api users

teemow, line 3: Additional use cases:

 * As a GS customer I want to use the Management API to manage clusters and applications via Gitops
 * As a GS customer I want to use the Management API to have more fine grained control over permissions and roles of api users

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,449 @@
...
+1. Metrics: `kubebuilder` already exposes a few metrics, and we can always add new ones
+   ([link](https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics)).
+1. Events: manager has a `GetEventRecorderFor` method that can be passed to the controller to register events.
+1. Finalizers: kubebuilder supports them so we can implement all the stuff we need (pause, replay).
+1. Extra endpoints: manager has `AddHealthzCheck` and `AddReadyzCheck` methods that feel similar to the `healthz`
+   endpoint.
+
+### Milestones
+
+#### MVP
```
teemow, line 138: This doesn't read like an MVP. Isnt this way too much? Why do we need to implement Pivoting? 

Imo a first MVP would be to use an existing mgmt cluster and expose CAPI-compatible CRDs of the workload clusters. Then we need to discuss what should be implemented in which order. Eg. first create and delete a cluster, then resizing of the cluster, then upgrade an existing cluster to the a CAPI cluster. Then upgrading CAPI clusters.

tfussell: I agree that pivoting shouldn't be part of the MVP. I will move it.

Separating migration of non-CAPI to CAPI from upgrades of CAPI clusters makes sense to me. I will be more specific with these.

I put these in MVP because a cluster that can't be deleted or a cluster without Calico, for example, doesn't feel useful. At least a cluster supporting this subset of features could be useful without being at feature parity with existing KVM clusters. In that case, moving upgrades out of the MVP is also reasonable.

rfc-capi-for-kvm-capk/README.md
tfussell: Good point. I'll change to one of these after reviewing docs.

rfc-capi-for-kvm-capk/README.md
tfussell: Thanks, I'll add these shortly.

tfussell:
> I've checked this before in the hackmd and I'm cool with this. Only nitpick, I think this is already a spec and IIRC your team is already having plans working towards this, or do you consider "doing CAPI for KVM" still as sth that is not decided? (I see RFCs as topics where the decision still has to be made)

We already made the decision that we would like to do this, but we were still looking for comments on the details (but not the implementation). If that makes it a spec rather than an RFC, I'll be happy to move it.

> As for the future work package planning of this, I would suggest aligning with the alpha, beta, RC, GA phases that we defined for CAPI in general (your MVP and prod definitions below go in that direction)

Good point. I forgot about these phases.

rfc-capi-for-kvm-capk/README.md
cornelius-keller: If we do the CAPI implementation following the upstream spec pivoting is a very low hanging feature and gives us new interesting opportunists, that is why I did put it in originally. It is also a good test if we implemented the upstream API correctly, as it would fail if we didn't.

teemow: Conceptually this brings a lot of work imo. You can gladly talk to @snizhana-dynnyk and @JosephSalisbury about it since we have a full team working on management cluster provisioning and automation. I am not sure if this is a good timing or not.

puja108: As with the other providers I would focus first on getting to what we have already. Sure, we should not implement anything that is not correct upstream API, but I would for now not set sth like "pivoting" as a goal per see, even if we assume it should work in theory.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Customers can upgrade existing clusters to CAPK without significant workload downtime.
+- Existing CAPI-compatible tools such as kubectl-gs, happa, GS api, and clusterctl work with CAPK clusters.
+- Team Rocket is responsible for a significantly reduced number of repos and lines of code.
+
+### Non-goals
+
+- Get CAPK accepted as a CNCF project.
+- Support mixed CAPK/CAPV clusters.
+- Support mixed management clusters (e.g. CAPK on AWS).
+- Support `MachineSet` or `MachinePool`.
```
JosephSalisbury, line 51: Why not?

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,449 @@
...
+### Scope
+
+We outlined all existing features in the GS KVM platform to define the scope of the implementation. Many features of
+platform can be provided by CAPI now. We therefore split the implementation into features that we must migrate to the
+new operator and those we must ensure are handled by CAPI as follows:
+
+#### Features to migrate to new operator
+
+- DNS and NTP configuration for WC nodes (OS level)
+- iSCSI initiator name (control plane nodes)
```
JosephSalisbury, line 63: What for? etcd?

rfc-capi-for-kvm-capk/README.md
JosephSalisbury: agree both ways. pivoting (as with upgrades) aren't part of _minimal_, but i bet they're pretty easy once we get stuff going

@cornelius-keller let's have a chat sometime about CAPK and MC bootstrap, that sounds fun

rfc-capi-for-kvm-capk/README.md
puja108: MachinePool is not a primitive that would be there on KVM, so ok to leave out.

This comment made me check out https://cluster-api.sigs.k8s.io/user/concepts.html though and isn't MachineSet just the versioned instantiation of a MachineDeployment (like in Depployments and Replicasets)? Why would we skip that then?

teemow: To complete the confusion. I'd even say we don't need MachineDeployments but MachinePools instead. The KVM provider runs on top of Kubernetes. So a MachineSet/MachineDeployment would be managing pods. A MachinePool would make our lives easier as it would use a Deployment for the VirtualMachines.


Example on AWS:
 * MachineSet => Manages individual EC2 instances
 * MachinePool => Manages an ASG

The same would be on K8s:
 * MachineSet => Manages individual pods for the VMs
 * MachinePool => Uses a Deployment that manages the pods

teemow: Oh and btw. We wanted to use a Deployment for the VMs on KVM in the first place. Just because there were some initial problems with init container, network etc we went with managing individual pods later on. Imo this would be a good time to evaluate this again especially if networking is finally different and we can simply use CNI.

puja108: In the upcoming version of KVM we are already using CNI AFAIK.

puja108: > MachinePool would make our lives easier as it would use a Deployment for the VirtualMachines.

I assumed a MachineDeployment already would deliver us the features of what we need from Deployments and we could maybe not have the need for another primitive, i.e. the Deployment, in between, but then I'm not deep enough into the CAPI primitives.

puja108: I thought about this a bit more and I think Timo's point towards MPs might actually be in line with what Thomas was drafting and that's maybe why he left out the MachineSets. If so, then that does make sense.

rfc-capi-for-kvm-capk/README.md
cornelius-keller: Yes, I am looking forward to this conversation.

rfc-capi-for-kvm-capk/README.md
cornelius-keller: How  I read the concepts `MachineSets` are not supposed to be used directly, but rather something used under the hood by the `MachineDeployment`. With regards to MachinePoos my impression is that the implementation is slightly  more complex than the MachineDeployments, also they are still experimental.  So I think it is fine to stay with MachineDeployments  for the MVP.

rfc-capi-for-kvm-capk/README.md
```diff
@@ -0,0 +1,449 @@
...
+This would mean that we would need to put a lot of effort into understanding and owning this code fully in order to be
+able to fix bugs and maintain KVM installations
+in the future.
+
+Instead of being the sole maintainer of a huge amount of legacy code, Team Rocket wants to invest this time in
+implementing CAPI for KVM operator and maintain relative platform parity with the other GS provider teams.
+
+### Goals
+
+- Customers can upgrade existing clusters to CAPK without significant workload downtime.
```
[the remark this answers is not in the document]
tfussell: Yes, I tried to define it in the title, but I could be more explicit here.

rfc-capi-for-kvm-capk/README.md
tfussell: It seems that this is actually passed in to masters and workers to fix a kubernetes bug. Here's our PR for it https://github.com/giantswarm/kvm-operator/pull/643 This is used for a systemd unit so we should be able to handle it in the `KubeadmConfig` so it doesn't need to be part of the operator. I'll drop it for now.

rfc-capi-for-kvm-capk/README.md
tfussell: It's difficult to find much information about these concepts. I agree that managing node pods with a normal `Deployment` sounds ideal in which case the `MachinePool` seems like a better fit: an ASG is similar to a Deployment for KVM.

It's not clear yet how much node-specific configuration we are going to have. If we want to, for example, mount a specific host path on a single worker node, we would need to create the `Machine` directly without a `MachineDeployment`. As long as the nodes are homogenous, a `MachineDeployment` seems like the easiest approach.

For the time being, I will drop this non-goal so we can make the final decision once the node configuration requirements are a little clearer.

tfussell:
To all involved so far: I'm copying this document to specs as that seems like a better fit. Feel free to continue discussions here, but I will be making changes and requesting reviews on the other PR: https://github.com/giantswarm/giantswarm/pull/18147

PR #23, opened 2021-07-12
Add RFC for CAPI webhook defaulting

MarcelMue:


defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,46 @@
+# Defaulting of CAPI clusters with webhooks
+
+This RFC describes a pattern for defaulting of CAPI clusters by utilizing Kubernetes webhooks extensively.
+
+## Problem statement
+
+We want clusters to be created with defaulted values.
+These values should match the clusters in our current product reasonably well.
```
puja108, line 8: link to https://intranet.giantswarm.io/docs/product/pdr/006_capi-product-guidelines-values/
 points feature parity and sensible defaults

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,46 @@
...
+Therefore we will need to default some values differently from upstream Cluster-API.
+
+We do not expect to have full control of the method of cluster creation in the future.
+Therefore we have to account for different methods of cluster creation ( e.g. kubectl, kubectl-gs, happa, ...) while keeping defaulting consistent between those methods.
```
puja108, line 12: 💯

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,46 @@
...
+- Webhooks must be able to determine which default values apply to the individual installation
+- `kubectl-gs` must be able to remove values from the CAPI templates efficiently
+
+### Source of truth for default values
+
+Defining the source of truth for default values will be hard in the future with different CAPI versions and other needs for configurability.
+We want to iterate faster at first and therefore a simple source of truth which might be inflexible is needed.
+
+Webhooks can utilize the already existing configuration options through `config-controller` to have access to global and installation specific defaults.
+It will be necessary to package webhooks (such as `Kyverno policies`) as apps which have their configuration managed through the [config](https://github.com/giantswarm/config) repository.
```
puja108, line 46: Maybe a short para below explaining how one would add a defaulting value similar to the 123 for users above would help ppl imagine how they're workflow will look like.

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,46 @@
...
+From this overview we can deduct a set requirements for our admission webhooks:
+- Webhooks must not have any side-effects (e.g. no additional objects are created or deleted)
+- Webhooks must be able to fully default a Custom Resource without context of other Custom Resources in the request (e.g. a `MachineDeployment` must be fully defaulted even if the `Cluster` CR is not known)
+- Webhooks must be able to determine which default values apply to the individual installation
+- `kubectl-gs` must be able to remove values from the CAPI templates efficiently
+
+### Source of truth for default values
+
+Defining the source of truth for default values will be hard in the future with different CAPI versions and other needs for configurability.
+We want to iterate faster at first and therefore a simple source of truth which might be inflexible is needed.
```
puja108, line 43: Let's clearly mention here that versioning of releases here is out of scope then and that it will be added in a way that still works similarly once we're over this first hurdle.

defaulting-capi-clusters/README.md
puja108: let's add gitops as a method as the saving of dry run manifests in git is a use case we definitely need to support for several customers already. this has a bit of impact on the webhook requirements you mention below

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,57 @@
...
+Versioning of configuration as releases for cluster upgrades is out of scope for this initial iteration.
+
+Webhooks can utilize the already existing configuration options through `config-controller` to have access to global and installation specific defaults.
+It will be necessary to package webhooks (such as `Kyverno policies`) as apps which have their configuration managed through the [config](https://github.com/giantswarm/config) repository.
+
+The workflow for a developer to add a new default configuration is then the following:
+1. `Developer` adds logic to new defaulting.
```
puja108, line 50: this is in kyverno-policies?

puja108: or basically in the webhook if we don't want to pin it down right now here

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,57 @@
...
+2. `Developer` adds templating to the app with local defaults in `values.yaml`.
```
puja108, line 51: then this is the kyverbno-policies/webhook app, right?

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Webhooks can utilize the already existing configuration options through `config-controller` to have access to global and installation specific defaults.
+It will be necessary to package webhooks (such as `Kyverno policies`) as apps which have their configuration managed through the [config](https://github.com/giantswarm/config) repository.
+
+The workflow for a developer to add a new default configuration is then the following:
+1. `Developer` adds logic to new defaulting (e.g. a `kyverno policy`).
+2. `Developer` adds templating to the app (e.g. [kyverno policies for aws](https://github.com/giantswarm/kyverno-policies/blob/main/helm/policies-aws/values.yaml) with local defaults in `values.yaml`.
+3. `Developer` adds global and installation specific defaults through the [config](https://github.com/giantswarm/config) repository.
+
+The workflow for a developer to then update update the default configuration is simple:
```
fiunchinho, line 54:
```suggestion
The workflow for a developer to then update the default configuration is simple:
```

fiunchinho:
I think it would be useful to explain how several mutating webhooks would work together, explaining how to tackle ordering between kyverno and other webhooks. What do you think?

MarcelMue:
> I think it would be useful to explain how several mutating webhooks would work together, explaining how to tackle ordering between kyverno and other webhooks. What do you think?

I tried to keep this discussion out for now as it's very technical. In general I believe we should make the webhooks as independent as possible from each other - so we limit the number of hard dependencies in ordering.

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,57 @@
...
+We want to iterate faster at first and therefore a simple source of truth which might be inflexible is needed.
+Versioning of configuration as releases for cluster upgrades is out of scope for this initial iteration.
+
+Webhooks can utilize the already existing configuration options through `config-controller` to have access to global and installation specific defaults.
+It will be necessary to package webhooks (such as `Kyverno policies`) as apps which have their configuration managed through the [config](https://github.com/giantswarm/config) repository.
+
+The workflow for a developer to add a new default configuration is then the following:
+1. `Developer` adds logic to new defaulting (e.g. a `kyverno policy`).
+2. `Developer` adds templating to the app (e.g. [kyverno policies for aws](https://github.com/giantswarm/kyverno-policies/blob/main/helm/policies-aws/values.yaml) with local defaults in `values.yaml`.
+3. `Developer` adds global and installation specific defaults through the [config](https://github.com/giantswarm/config) repository.
```
stone-z, line 52: When should someone store the value in `kyverno-policies` as opposed to `config`?

MarcelMue: Maybe weird phrasing: You need to make something configurable in the `values.yaml` to be able to overwrite it through the `config` repository.

stone-z: Gotcha -- I think the confusing phrase is "local defaults" which I interpreted to refer to cluster-local defaults and not code-local placeholders

defaulting-capi-clusters/README.md
```diff
@@ -0,0 +1,57 @@
...
+1. `User` triggers cluster templating with `kubectl-gs template cluster`
+2. `kubectl-gs` fetches upstream CAPI templates
+3. `kubectl-gs` **removes** values from CAPI templates which we want to default differently
+4. `kubectl-gs` returns minimal templates
+5. `User` either `kubectl apply`s minimal templates or runs `dry-run`
+6. `Admission Webhooks` default all missing values
+
+From this overview we can deduct a set requirements for our admission webhooks:
+- Webhooks must not have any side-effects (e.g. no additional objects are created or deleted)
+- Webhooks must be able to fully default a Custom Resource without context of other Custom Resources in the request (e.g. a `MachineDeployment` must be fully defaulted even if the `Cluster` CR is not known)
```
pipo02mix, line 36: I guess the validation step (after mutation) must make sure the properties are correct. Have you discussed in general how to make that? I see will be similar to defaulting (mutation)

fiunchinho (approved):
LGTM, let's do it!

njuettner (approved):

alex-dabija (approved):

defaulting-capi-clusters/README.md
MarcelMue: Yes - we currently don't focus on it but generally the workflow is the same for now.

stone-z (approved):

JosephSalisbury:
@MarcelMue we good to merge here?

PR #24, opened 2021-08-30
Add initial RFC for capi configuration

MarcelMue:


capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+      infrastructureRef:
+        name: capa-mp-0
+```
+
+This minimal custom resource can also be represented as a template which will remain valid throughout the lifecycle of `v1alpha3` in this case:
+```yaml
+apiVersion: cluster.x-k8s.io/v1alpha3
+kind: MachinePool
+metadata:
+  name: {{clusterID}}-mp-0
```
puja108, line 106: not sure what the template is for, but could/should the index in the end not also be templated?

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+- `spec.replicas`
+- `spec.template.spec.version`
+
+We can now construct a `ConfigMap` with these values which are then going to be used by any defaulting tool/webhook:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: machinepool-21-0-0
+  namespace: giantswarm
```
puja108, line 164: note here as a reminder we might want to think about MC namespace structure towards transparency and security before we just put it in `giantswarm`

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+```
+It is important to note that a assumption is that **no** values in the base file are immutable!
+
+### Cluster-api context dependent values
+
+We have some cases in `cluster-api` where a value in a custom resource is neither part of the fundamental structure nor is it easy to statically default.
+
+These values are usually highly dependent on the context in which the cluster is created in.
+For example the `CIDR` in which the cluster is created.
+We currently assume that these values will be defaulted by custom webhooks or operators with significant provider specific domain knowledge.
```
puja108, line 180: I guess this means we keep them out of scope of this RFC?

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+  v21.0.0:
+    base:
+      cluster.yaml # configmap containing the base file (defined above)
+      aws-cluster.yaml
+      machinepool.yaml
+      awsmachinepool.yaml
+      ...
+      net-exporter.yaml # configmap containing version information for defaulting!
+    overlay:
+      special-overlay.yaml # Added by customer to e.g. overlay a different flatcar version!
```
puja108, line 198:
```suggestion
      special-overlay.yaml # Added by customer/AE/SA to e.g. overlay a different flatcar version!
```

capi-configuration/README.md
MarcelMue: Yes - I just decided to keep it kind of simple here. Idea would be that e.g. happa or kubectl-gs uses this template

capi-configuration/README.md
MarcelMue: Agreed - I am not sure about namespacing yet.

capi-configuration/README.md
MarcelMue: Yes - mainly mentioned it here for completeness as it was discussed before.

puja108 (commented):
First, I like a lot where this is going. simple versioning with simple YAML as well as thinking about a more transparent interface and more collaboration with customers are both really cool.

I worry a bit, that this is almost like 3 RFCs in 1, or at least this might be 3 phases:
1. Have release versions as gitops-deployed configmaps.
2. Share the release version git with customer, to enable customizations.
3. Have customers re-use the same git for cluster creation.

I'm very fine with 1 (even if there might be dragons in form of secrets or other stuff)

With 2 I think this is a cool logical next step, but we don't need to make it already. Especially cool here would be being able to empower mature customers to cherry pick changes themselves or with an AE/SA.

With 3, while I see a lot of cool possibilities, I'm a bit more worried here, as it might be more complicated and we might reduce the freedom we give customers in their choices (of git repo). Also needs a lot of thoughts towards multi-tenancy, e.g. giving access to a certain clusters config to a certain team. I still like the general direction, but not sure it needs to be combined with this, as gitops for cluster creation will anyway be a topic most probably even before CAPI and biscuit is starting to talk to customers about it.

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+  name: machinepool-21-0-0
+  namespace: giantswarm
+data:
+  spec:
+    replicas: 3
+    template:
+      spec:
+        version: v1.21.4
+```
+It is important to note that a assumption is that **no** values in the base file are immutable!
```
stone-z, line 172:
```suggestion
It is important to note that an assumption is that **no** values in the base file are immutable!
```

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Cluster-api context dependent values
+
+We have some cases in `cluster-api` where a value in a custom resource is neither part of the fundamental structure nor is it easy to statically default.
+
+These values are usually highly dependent on the context in which the cluster is created in.
```
stone-z, line 178:
```suggestion
These values are usually highly dependent on the context in which the cluster is created.
```

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+For example the `CIDR` in which the cluster is created.
+We currently assume that these values will be defaulted by custom webhooks or operators with significant provider specific domain knowledge.
+
+### Putting together a Giant Swarm release
+
+A Giant Swarm releases is just a set of `cluster-api` base files as defined previously with added `apps`.
```
stone-z, line 184:
```suggestion
A Giant Swarm release is just a set of `cluster-api` base files as defined previously with added `apps`.
```

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+The following folder structure would be contained in `giantswarm/example-customer`:
+```yaml
+aws:
+  v21.0.0:
+    base:
```
stone-z, line 190: Why are `base` defaults not defined globally (like in `releases`)? Why per-customer?

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+In the management cluster a release is then simply a set of configmaps.
+The configmaps for e.g. release `v21.0.0` will be applied through argoCD by taking the `base` configmaps, applying any `overlay` from a customer and then deploying them.
+
+As shown in previous examples these configmaps would contain the release in their name (e.g. `machinepool-21-0-0`).
+Any in-cluster webhook or operator can then easily navigate to these configmaps to look up defaults for a cluster.
+
+A cluster creation workflow can now look as follows:
+1. Client (e.g. `happa`) submits `--dry-run` with `fundamental` custom resources to management cluster.
```
stone-z, line 213: I think the important part is that they send the minimal CR, e.g. from the example above?
```suggestion
1. Client (e.g. `happa`) submits `--dry-run` with `minimal` custom resources to management cluster.
```

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Going gitops all the way
+
+A next step in `gitops` adoption would be the addition of having the cluster custom resources in `gitops` as well.
+So the shared git repository would contain not only release information but also the clusters themselves.
+
+```yaml
+gauss: # installation
+  peu01: #clusterID
+    base: # managed fully by GS
```
stone-z, line 232: Same question here -- why do we want to duplicate a static base between customers?

capi-configuration/README.md
```diff
@@ -0,0 +1,333 @@
...
+aws: # releases as mentioned before
+  v21.0.0:
+  ...
+```
+
+Interaction with this repository could then be improved by adding functionality to our CLI `kubectl-gs`.
+
+Cluster creation would become a `pull request` which commits code generated by `kubectl-gs`.
+
+Upgrading a cluster is a bit more complex: Upgrades become a overwrite of configuration from the release to the `base` folder.
```
stone-z, line 249: This part does not make sense to me. IMO a base for a given release should never change. Wouldn't an upgrade be a change to the cluster resources?

Keeping the approach from above, wouldn't an upgrade be something like:
0. Client fetches the resource to be upgraded (from cluster or from git)
1. Client submits `--dry-run` with the new value(s) to the API server
2. Mutating webhooks apply the new defaults (like upgrade-operator)
3. Client receives fully defaulted custom resources and allows users to make manual changes.
4. User opens PR with changes to cluster resources
5. PR is merged, cluster is created.

capi-configuration/README.md
MarcelMue: In my mind there is a global definition in a GS repository which gets replicated to customer repositories. This is because templating tools such as `kustomize` can not reason across repositories AFAIK

capi-configuration/README.md
MarcelMue: There is two `base` folders:
The release has a `base` with optional `overlay`
The cluster also has a `base` with optional `overlay`

What I describe here is an update of the `base` for the cluster - essentially taking the result of `release base` + `release overlay` and applying it to `cluster base` while `cluster overlay` stays unchanged.

The webhook workflow does not work like that unfortunately. If you submit a CR with non empty fields it will become undecidable if the customer e.g. wants to have a special setting in their cluster. The webhook would need to either overwrite all values (losing any custom setting for the cluster) or touch no already set values (essentially not upgrading).

Maybe we can discuss how to clarify this better.

MarcelMue:
> First, I like a lot where this is going. simple versioning with simple YAML as well as thinking about a more transparent interface and more collaboration with customers are both really cool.
> 
> I worry a bit, that this is almost like 3 RFCs in 1, or at least this might be 3 phases:
> 
>     1. Have release versions as gitops-deployed configmaps.
> 
>     2. Share the release version git with customer, to enable customizations.
> 
>     3. Have customers re-use the same git for cluster creation.
> 
> 
> I'm very fine with 1 (even if there might be dragons in form of secrets or other stuff)
> 
> With 2 I think this is a cool logical next step, but we don't need to make it already. Especially cool here would be being able to empower mature customers to cherry pick changes themselves or with an AE/SA.
> 
> With 3, while I see a lot of cool possibilities, I'm a bit more worried here, as it might be more complicated and we might reduce the freedom we give customers in their choices (of git repo). Also needs a lot of thoughts towards multi-tenancy, e.g. giving access to a certain clusters config to a certain team. I still like the general direction, but not sure it needs to be combined with this, as gitops for cluster creation will anyway be a topic most probably even before CAPI and biscuit is starting to talk to customers about it.

I would still like to have some discussion around the 3 as I think at least a lighthouse vision would be nice.

WDYT if I split the 3 into into separate files / PRs for now?

capi-configuration/README.md
puja108: yeah, that's why I thought the MP ID should be templated, too, so the clients could create several of them at the same time

capi-configuration/README.md
puja108: I think this is a global one, just that we would distribute it (still think that's 2nd phase) to customer specific repos, to then be able to use the customer specific repo as an interface of transparency to the customer, this way they also would not need to check back with the source of truth just to see the base settings.

puja108:
> I would still like to have some discussion around the 3 as I think at least a lighthouse vision would be nice.
> WDYT if I split the 3 into into separate files / PRs for now?

Ok, let's keep them together in the RFC for a first review and the lighthouse vision for now (it's and RFC anyway, not a spec.

We can then see where people are discussing towards and if we can already pull e.g. the 1st phase off of this into a spec/ADR and start PoCing. WDYT?

capi-configuration/README.md
stone-z: I don't see how the `rel base` + `rel overlay` solves the undecidability issue though if using webhooks, since the CR will still hit the same webhooks once applied. I feel like in either case we'd have to perform the defaulting and apply the resources (bypassing the defaulting) in separate steps. 

Thinking about other UX, we might have to somehow track which fields were set by the user anyway. For example so that from a CLI tool I can just change the cluster version and all necessary additional defaulting happens behind the scenes.

capi-configuration/README.md
stone-z: Ok I see. If it is a technical limitation we can work around, I'd personally be inclined to not duplicate the source of truth, but would be worth getting some other opinions on if it's that much better to have them in the customer repo. Having only theoretically immutable resources in a bunch of shared repos just seems like begging for some well-meaning person to modify them

MarcelMue:
Adding some more reviewers while I continue to iterate here.

capi-configuration/README.md
anvddriesch: Currently `kubectl-gs` templates the machinepool name simply as the nodepool id in the same way as we do it for our current clusters.

MarcelMue: Makes sense - I will change this as I took it from upstream templates.

MarcelMue: Done.

capi-configuration/2_gitops-management.md
```diff
@@ -0,0 +1,100 @@
...
+
+We want to ...
+1. enable customers to have more agency in upgrading individual components.
+2. give a lower barrier of entry for customers to deeply interact with our product.
+3. allow customers to use known gitops tooling for cluster management.
+
+## Going gitops all the way
+
+A next step in `gitops` adoption would be the addition of having the cluster custom resources in `gitops` as well.
+So the shared git repository would contain not only release information but also the clusters themselves.
```
kopiczko, line 15: That's interesting. We assumed that customers would setup their gitops repos as they want. Ping @JosephSalisbury

capi-configuration/0_capi-releases.md
```diff
@@ -0,0 +1,205 @@
...
+
+We want to ...
+1. decouple managed applications from strict release flows.
+
+## Git-Ops approach
+
+We have already taken steps to manage configuration for our components in management clusters through open source gitops tools.
+This has set a precedent for gitops management for configuration in other areas as well.
+
+Our customers generally desire to work with an `Infrastructure As Code` approach which is extendended by currently available open source gitops tooling.
```
tfussell, line 19:
```suggestion
Our customers generally desire to work with an `Infrastructure As Code` approach which is extended by currently available open source gitops tooling.
```

capi-configuration/0_capi-releases.md
```diff
@@ -0,0 +1,205 @@
...
+We need to ...
+1. be able to define Giant Swarm releases as sets of configuration.
+2. be able to create Giant Swarm clusters in a repeatable and reliable way.
+3. be able to upgrade Giant Swarm clusters in a repeatable way with `cluster-api`.
+4. be able to default Giant Swarm clusters through all our user interfaces consistently.
+
+We want to ...
+1. decouple managed applications from strict release flows.
+
+## Git-Ops approach
```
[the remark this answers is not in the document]
MarcelMue: True - that's on me mixing things up. Would you be okay with `gitops`?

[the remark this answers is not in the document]
tfussell: I don't think it needs to be quoted, but yeah, lowercase is fine IMO.

capi-configuration/2_gitops-management.md
puja108: Biscuit definitely needs to check that with customers in general, but I tend to agree that limiting that to something controlled by us might sooner or later reach its limits when trying to accomodate for customer peculiarities.

puja108:
> Will GitOps be required for customers or is this just an alternative to keeping state only in the Kubernetes API?

I'd say making it a requirement for cluster management would limit us, but making it a requirement for versioning, version management, customizations, rather as an interface to some of our advanced functionality is fine.

capi-configuration/0_capi-releases.md
```diff
@@ -0,0 +1,205 @@
...
+These fields can not be inferred from a webhook without context knowledge.
+It is therefore necessary to supply these values _immediately_ when the custom resource is created (e.g. `kubectl-gs` or `happa` needs to fill them).
+We therefore define `fundamental templates` as minimal templates which have to be submitted to create a working cluster.
+
+We can look at a valid `MachinePool` custom resource to determine its fundamental structure:
+```yaml
+apiVersion: cluster.x-k8s.io/v1alpha3
+kind: MachinePool
+metadata:
+  name: capa-mp-0 # This is taken from upstream, Giant Swarm would use an ID like `eump1`.
```
njuettner, line 40: Just for the completeness, shall we add the namespace? I'm asking because of the organization handling.

MarcelMue: Makes sense to me - will add namespace in all examples. Do you think the namespace should be set by clients? I am not sure how we currently determine this with organisations and so on.

I would guess the client (e.g. happa) sets the namespace.

njuettner: From what I know in firecracker, we do set the namespace by default with release AWS 16.0.0 and also validate that all CR's matching with the Cluster CR namespace if that is not correct. But only with new cluster creation.

capi-configuration/2_gitops-management.md
JosephSalisbury: yeah, i think providing a structure for customer CRs is gonna be hard, we've already seen a fair bit of differentiation between customers as part of GitOps product discovery

capi-configuration/1_default-customization.md
```diff
@@ -0,0 +1,29 @@
...
+
+We want to ...
+1. give customers greater control over how their clusters are defaulted.
+2. enable customers to have more agency in upgrading individual components.
+
+## Enabling customer access
+
+We have to be able to give customers access to the `git` source in a gitops setup in order to allow effective modification of any configuration.
+
+We already have a shared repository in the Giant Swarm organization with each customer individually.
```
JosephSalisbury, line 15: So, we'd use each customer's repo as a source for gitops data. How do we synchronise across customer repos?

JosephSalisbury (commented):
I'd recommend breaking this up into a few RFCs.

e.g:
- CAPI defaulting
- using customer repos for gitops
- providing repos for cluster CRs

i'm not convinced they're super linked

JosephSalisbury:
oh lol, i should actually read all the comments, y'all have spoken a lot about whether to split it up or not already :D

capi-configuration/0_capi-releases.md
```diff
@@ -0,0 +1,209 @@
...
+        kind: AWSMachinePool
+        name: capa-mp-0
+      version: v1.16.8
+```
+- We can not infer `apiVersion` as it's an essential field.
+- We can not infer `kind` as it's an essential field.
+- We can not infer `metadata.name` as it's required to create the custom resource.
+- We can not infer `spec.clusterName` as it's unique to the created cluster.
+- We can not infer `spec.template.spec.bootstrap.configRef.name` as it's unique to the cluster.
+- We can not infer `spec.template.spec.infrastructureRef.configRef.name` as it's unique to the cluster.
```
fiunchinho, line 64:
```suggestion
- We can not infer `spec.template.spec.infrastructureRef.name` as it's unique to the cluster.
```

fiunchinho:
> > Will GitOps be required for customers or is this just an alternative to keeping state only in the Kubernetes API?
> 
> I'd say making it a requirement for cluster management would limit us, but making it a requirement for versioning, version management, customizations, rather as an interface to some of our advanced functionality is fine.

can you elaborate on this @puja108 ? I don't understand exactly what you mean. Customers would be able to create clusters with our clients, even though they don't commit the manifests to the repository?

puja108:
> can you elaborate on this @puja108 ? I don't understand exactly what you mean. Customers would be able to create clusters with our clients, even though they don't commit the manifests to the repository?

Yeah, it is should never be a requirement. But the way it is explained in this RFC it sounds like we'd be offering special features only in this way, and that's what I don't like about it.

TLDR. Fine with part 1 of the RFC, and could see customers also use part 2, but part 3 worries me more and is also heavily influenced by the gitops features we are already rolling out in biscuit right now.

My suggestion would be to go on and split off part 1 of the RFC into a spec or ADR and just get it done, this is the only thing we really need on the critical path to a CAPI product rightnow. customer customizations and such are not in the first GA scope even I'd say, for now we need to solve the versioning and try it out and find the issues with it. Fine to keep the rest of the RFC as ideas how we go forward once we have part 1.

MarcelMue:
I have transformed the first part of this RFC into an internal spec here: https://github.com/giantswarm/giantswarm/pull/18937

This is following positive feedback on my PoC in kubect-gs here: https://github.com/giantswarm/kubectl-gs/pull/463

MarcelMue:
@webwurst @ubergesundheit - we discussed some of the points I put into this RFC & some of the discussion taking place here.

(This is from my onsite notes :D )

alex-dabija (approved):
We had a deadline until yesterday to get feedback on this proposal. I'm approving as agreed. Also, it seems that the last comment was on the 4th of October 2021.

capi-configuration/README.md
stone-z: It still isn't clear to me how the webhooks will be able to differentiate between fields they should mutate or not during an upgrade. Maybe the idea is to rely on [server-side apply](https://opensource.googleblog.com/2021/10/server-side-apply-in-kubernetes.html)?

MarcelMue: We are exploring different ideas here for the upgrades. We focus on creating releases and versioned clusters for now but there will be a follow up spec to nail down `upgrades` once we have more experience.

The current slack channel is `wg-capi-releases` - we have not checked `server-side apply` as described in the article yet.

PR #25, opened 2021-11-02
Adds RFC about improving managed apps user experience

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/19349

This RFC attempts to provide a vision for improving the user experience / operations of managed apps.

### DIFF
--- a/managed-apps-vision/README.md
+++ b/managed-apps-vision/README.md
@@ -1,0 +1,30 @@
+# Managed Apps Vision
+
+This RFC describes a general vision for improving the user experience of Managed Apps.
+
+With a previous iteration of the structure of Giant Swarm, one team was responsible for all of the managed apps. As the number of apps that Giant Swarm managed increased, it became more difficult to own these apps.
+
+Giant Swarm has updated its structure so that ownership of apps is spread across multiple teams - instead of one team owning all apps, each team owns a number of apps. Each team is now responsible for an area of solutions, such as monitoring, and are then responsible for apps that help in that area, such as Prometheus. This ownership also includes concepts such as trainings or sales enablement. We've been referring to this concept as 'holistic ownership'.
+
+With this change, more teams and more people are working on apps, overall. This means we can spend more time on each app, and provide a better user experience overall - we can start working more _deeply_.
+
+As some examples of what working more deeply on apps means:
+- Working upstream to fix specific customer issues, or provide feedback to maintainers on issues with upgrade paths.
+- Building more application specific monitoring and alerting, so Giant Swarm is aware of application level issues before customers.
+- Improving support for automatic scaling, so apps don't OOM on larger clusters, or scale on custom metrics.
+- Building dashboards (or reusing upstream dashboards) and making them available to customers.
+- Working with the incident and postmortem workflows to ensure that issues with applications are learnt from, improved upon, and applied to all customers.
+
+These are just examples - it is up to each team to identify how to provide a better experience with the apps they manage. Overall, we should be aiming to improve our operations of managed apps, and reduce the amount of work our users need to do.
+
+There are going to be situations in the future where we need to work directly with customers on managed apps, as the line of responsibilities becomes better defined. We should identify where we can help reduce the operational work customers need to do, and then do so - managed apps means we help with the management of apps.
+
+Upgrades are one example. Giant Swarm can work with customers to help discover what is blocking customers from upgrading, and then help remove these blockers, such as with major version upgrades that require user interaction. Ideally, all customers run on the latest version of apps. We can improve our automation in future to start rolling out minor and patch versions automatically.
+
+Giant Swarm should ensure that new versions of applications are available as quickly as possible - ideally, within hours of new versions being available upstream.
+
+As another example, one customer cannot use automatic scaling for ingress controllers, as their network traffic spikes with certain events too quickly for automatic scaling to deal with. This means they need to manually scale beforehand. We can provide a better experience - for example, providing a system that allows for scheduling a scaling event in future. This would reduce the user's overall operational load, and provide a more managed experience.
+
+These are examples of a more general idea - identifying customer problems with managed apps, working together to find a solution, and then building that feature. Ideally, this could even be pushed or integrated upstream, or provided as a generic, open-source solution. As usual, we should evaluate whether developing a solution into a feature is worth it - for example, because it helps multiple customers, or reduces toil.
+
+To sum up - Giant Swarm should be continuously improving the operations of managed apps, both to improve the overall user experience, and reduce our user's operational load for managed apps. Managed apps are managed.


JosephSalisbury:
@giantswarm/sig-product i'd like some initial feedback, before i push this into a wider audience, including customers. fanks! <3

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
+# Managed Apps Vision
+
+This RFC describes a general vision for improving the user experience of Managed Apps.
+
+With a previous iteration of the structure of Giant Swarm, one team was responsible for all of the managed apps. As the number of apps that Giant Swarm managed increased, it became more difficult to own these apps.
+
+Giant Swarm has updated its structure so that ownership of apps is spread across multiple teams - instead of one team owning all apps, each team owns a number of apps. Each team is now responsible for an area of solutions, such as monitoring, and are then responsible for apps that help in that area, such as Prometheus. This ownership also includes concepts such as trainings or sales enablement. We've been referring to this concept as 'holistic ownership'.
```
Oshratn, line 7:
```suggestion
Giant Swarm has updated its structure so that ownership of apps is spread across multiple teams - instead of one team owning all apps, each team owns a number of apps. Each team is now responsible for a solution to a specific need. This solution is typically made up of one ore more open source apps. An example of such a solution is monitoring. The team is responsible for apps that help in that area, such as Prometheus. This ownership also includes concepts such as trainings and sales enablement. We've been referring to this concept as 'holistic ownership'.
```

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+With this change, more teams and more people are working on apps, overall. This means we can spend more time on each app, and provide a better user experience overall - we can start working more _deeply_.
```
Oshratn, line 9:
```suggestion
With this change, more teams and more people are working on apps, overall. This means we spend more time on each app, and provide a better user experience overall - we can work more _deeply_ on each solution.
```

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+As some examples of what working more deeply on apps means:
+- Working upstream to fix specific customer issues, or provide feedback to maintainers on issues with upgrade paths.
+- Building more application specific monitoring and alerting, so Giant Swarm is aware of application level issues before customers.
+- Improving support for automatic scaling, so apps don't OOM on larger clusters, or scale on custom metrics.
+- Building dashboards (or reusing upstream dashboards) and making them available to customers.
+- Working with the incident and postmortem workflows to ensure that issues with applications are learnt from and improved upon.
```
Oshratn, line 16:
```suggestion
- Working with the incident and postmortem workflows to ensure that issues with applications are learnt from, improved upon and applied to all customers.
```

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+These are just examples - it is up to each team to identify how to provide a better experience with the apps they manage. Overall, we can should be aiming to improve our operations of managed apps, and reduce the amount of work our users need to do.
```
Oshratn, line 18:
```suggestion
These are just examples - it is up to each team to identify how to provide a better experience with the apps they manage. Overall, we are aiming to improve our operations of managed apps, and reduce the amount of work our users need to do.
```

Oshratn (commented):
❤️

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+There are going to be situations in the future where we need to work directly with customers on managed apps, as the line of responsibilities becomes better defined. We should identify where we can help reduce the operational work customers need to do, and then do so - managed apps means we help with the management of apps.
+
+Upgrades are one example. Giant Swarm can work with customers to help discover what is blocking customers from upgrading, and then help remove these blockers, such as with major version upgrades that require user interaction. Ideally, all customers run on the latest version of apps. We can improve our automation in future to start rolling out minor and patch versions automatically.
+
+Giant Swarm should ensure that new versions of applications are available as quickly as possible - ideally, within hours of new versions being available upstream,
```
marians, line 24:
```suggestion
Giant Swarm should ensure that new versions of applications are available as quickly as possible - ideally, within hours of new versions being available upstream.
```

marians:
Thanks Joe for getting us started!

I think in the direction of defining quality standards for managed apps, the text leaves way too much responsibility to the teams.

> "it is up to each team to identify how to provide a better experience with the apps they manage."

For basic aspects around the managed apps experience I think we should instead list our expectations clearly. These expectations are still WIP. Just let me try to formulate some example phrases:

It's the owner team's responsibility to

- keep the app's source repository and the app's metadata in shape to meat our quality standards (which may or may not become part of the RFC)
- Provider user-friendly release notes/change logs
- continuously improve the app's documentation based on user/customer feedback and based on experience from operations
- provide troubleshooting recipes for operations and support
- ...

Just some examples. I care about the first one. I'd favour making the quality standards part of the RFC. However I expect a need to keep those up-to-date in the future. Not sure whether RFCs are meant to be updated or be immutable.

managed-apps-vision/README.md
puja108: not sure that's right, teams own a "field/area of solutions" for several interrelated needs, not just one solution for one need, and depending on use cases you might have several of those needs or just one of them for now, e.g. for obersvability the team might own a tracing solution but not everyone might have the use cases for that.

managed-apps-vision/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+As another example, one customer cannot use automatic scaling for ingress controllers, as their network traffic spikes with certain events too quickly for automatic scaling to deal with. This means they need to manually scale beforehand. We can provide a better experience - for example, providing a system that allows for scheduling a scaling event in future. This would reduce the user's overall operational load, and provide a more managed experience.
+
+These are examples of a more general idea - identifying customer problems with managed apps, working together to find a solution, and then building that feature. Ideally, this could even be pushed or integrated upstream, or provided as a generic, open-source solution.
```
puja108, line 28: I would add that the step from solution to feature is not a MUST step but one that we take when we see that a certain functionality makes sense for more than just a one-off and we want to drive it towards a real product (feature)

puja108 (approved):
Puts the concept into words pretty well.

The whole thing for a while makes me wonder though, if it should still be called "managed apps" now that the whole concept is moving much more towards solving needs/pains with solutions that might be apps but also might be (added) product features.

That doesn't have to be solved in this RFC though. LGTM on this

marians:
As said in our apps sync, I misunderstood the goal of this RFC. Feel free to ignore my lengthy comment about quality standards.

managed-apps-vision/README.md
JosephSalisbury: Does that new / last sentence look good to you?

managed-apps-vision/README.md
JosephSalisbury: Went with 'should be aiming', that makes more sense to me with what i wanna get across

managed-apps-vision/README.md
JosephSalisbury: I don't know what the correct word is for the difference between these tenses, but I think cause this is a proposal for a change in the future, putting it as 'we _can_ spend more time' makes more sense, right? similarly with 'we can start working'

managed-apps-vision/README.md
JosephSalisbury: i think it kind of depends on how you model the solutions - you can keep them at a general level of 'solve monitoring', or a more specific level of 'solve monitoring problem a', or 'solve monitoring problem b' etc.

i do think area of solutions makes more sense though, because teams don't have a _single_ solution right now

piontec (approved):
Looks really good to me. I wonder if we should add something about evolving and implementing best practices across all managed app teams. WDYT?

othylmann (approved):
I think it is in general good. I am not sure from it what we do to automate the ugprades of apps, but that is probably next. Hence approved.

JosephSalisbury:
okay, this has a bunch of actions coming out of it, let's go

comment responses:
- @puja108 i'm not doing anything else regarding naming, I assume if you want to carry that further, you have it tracked somewhere
- @piontec for sharing best practices, we have our regular Managed Apps, I've added 'sharing best practices' to our recurring agenda there, that should be good enough

actions:
- this is a fairly philosophical rfc, so there's not a ton of concrete stuff coming out of it
- for automated app upgrades see https://github.com/giantswarm/roadmap/issues/124 (fyi @piontec)
- to kickstart the process of overall improving the operations of our managed apps, see https://github.com/giantswarm/roadmap/issues/601, let's ensure that all managed apps are scraped by prometheus, and have at least a basic up/down alert. My reckoning is that should help kickstart the operational improvement cycle (i.e: shit breaks, we get alerted, we fix it). i'll push that forwards
- i'll talk with sig product about how to improve getting app updates into the catalogs, as well as helping users to upgrade. we've got a fairly lightweight process going in cabbage that i want to see how it works for a bit longer, and then share it around a bit
- i'll also talk with sig product about if we want public dashboards for all managed apps - idea here is to help users see what's going on more, as well as help improve things for our operations

JosephSalisbury:
but i think we're done with dicussion here at least, so merging. thanks gang! bam!

managed-apps-vision/README.md
Oshratn: Are our RFCs public?
The thing is that I have a preference for strong and single minded language, rather than something that creates loopholes or doubts.

PR #26, opened 2021-11-23
add RFC for CAPI WC upgrades

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/19896


capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,52 @@
+# Handling workload cluster upgrades
+
+## Problem Statement
+
+We want to ...
+1. Make all changes transparent for the user before application
+2. Continue supporting the current workflow of upgrading a cluster (Through GS staff, customers and automation)
```
MarcelMue, line 7: Can we define a bit more closely what we mean here? Like the one button upgrade, timed upgrade and so on?

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,52 @@
...
+3. Integrate the process into the proposed gitops-management
+4. Utilize the managed-configmap approach to ensure clusters adhere to the definition in their release at all times
+5. Ensure that an upgrade to a cluster is decoupled from upgrades of other components
+6. Allow easy integration into other pipelines and formats users might want to utilize for cluster management.
+
+## Triggering an Upgrade
+
+Our current method to trigger an upgrade is to change the release-version label on the `Cluster` resource. This can be done by GS staff, customers and automation through various interfaces.
+This change will trigger the actual upgrade by letting a newer set of components manage the cluster.
+However, when introducing the [new structure for releases](0_capi-releases), this process will be different. The release will no longer be tied to the components.
```
MarcelMue, line 17: I think it would be valuable to explain here why changing the release label will not work in the future. It is just completely incompatible with a gitops approach to cluster management.

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,52 @@
...
+Instead, each release version will relate to a set of configmaps, defining the defaults and basic structure for the cluster resources.
+So, triggering an upgrade in the future will mean to request the change of certain values within the CR itself. This is a breaking change. (changing the release version label alone will no longer trigger an upgrade)
+
+## Upgrade Mechanism
```
MarcelMue, line 21: I think this should explain the process of "someone pressed a button" we had discussed.

Additionally we should reference the RFC by fer for an interface:
https://github.com/giantswarm/rfc/tree/main/automatic-cluster-upgrades

anvddriesch: Hmm can you be more specific about the "someone pressed a button" we had discussed?
I was thinking to add more to the `Client` section. The automatic cluster upgrades would fit there as well.

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,55 @@
...
+7. Allow easy integration into other pipelines and formats users might want to utilize for cluster management.
+
+## Triggering an Upgrade
+
+Our current method to trigger an upgrade is to change the release-version label on the `Cluster` resource. This can be done by GS staff, customers and automation through various interfaces.
+This change will trigger the actual upgrade by letting a newer set of components manage the cluster.
+However, when introducing the [new structure for releases](0_capi-releases), this process will be different. The release will no longer be tied to the components.
+Instead, each release version will relate to a set of configmaps, defining the defaults and basic structure for the cluster resources.
+So, triggering an upgrade in the future will mean to request the change of certain values within the CR itself. This is a breaking change since changing the release version label alone will no longer trigger an upgrade.
+The reason for this is that we want to move towards a gitops-centered management of the cluster lifecycle.
```
puja108, line 21: Does this mean we would "force" everyone to use gitops or just that we know a big part of our customers will be using a gitops approach?

capi-configuration/3_upgrades.md
puja108: IMHO ideally the upgrade mechanism is always the same on a lower level, different clients would just enable it to be used in different ways.

Example:
- assumed kgs (gitops) approach would be as described below in the slides (the exact approach is not directly impactful in this context)
- automated upgrade (for non-gitops clusters) would do exactly the same, just that the whole procedure would be done by an operator
- happa/button (for non-gitops clusters) basically could re-use the same mechanics as the automated upgrade but without scheduling (or even with scheduling as someone could schedule an upgrade for tomorrow midnight)

Looking at the example, basically we can differentiate between two upgrade modes:
- gitops mode where the source of truth is in git (interfaces need to be either outputting files or doing actual PRs to git repos, automated upgrade can only be done with automation that has access to the users' git)
- non-gitops mode where the source of truth is in the MC (ideally there's an operator that can mutate resources towards an upgrade, that can then be triggered by different interfaces, i.e. CLI, Happa, auto-upgrade/scheduler:)

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,55 @@
...
+One easy way to get this information is to make a `dry-run` request to the management API to create such a cluster.
+The proposed changes can then be reviewed as a `diff` between the two clusters or just as the mutated cluster resources themselves.
+This would also give us the chance to apply them not only directly via management-API but also by [creating a commit to a git repository](2_gitops-management).
+
+## Mutable Fields
+
+With the approach described above, we need to be aware of how specific fields should be treated during an upgrade.
+If an empty field has a default value in the new version, it should be defaulted to this value.
+If a populated field has a default value in the new version, we can either overwrite it or leave it unchanged.
+In order to determine which fields are mutable and should be overwritten during an upgrade, the definition of mutable fields has to be part of the release definition itself or a separate upgrade definition.
```
puja108, line 37: This sounds like a daunting task and could lead us to be forced to make additions to the "mutable" fields, based on feature requests like "I want to enable featureflag X on the API server". I sadly have no better idea right now though.

If used in an gitops approach this might be less of a problem, as users can work around "limtations" by using sth like kustomize or just adjusting the upgrade PR with another commit (I do this with the upgrades of my git-managed blog). Not sure if we could make certain features "expert-mode"-only, saying they're only available if you move to a gitops-like approach though.

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,55 @@
...
+## Example workflow
+
+![Example](upgrade.gif)
+
+See slides [here](https://docs.google.com/presentation/d/1_ImURpdO3T8HxyBNraTsoAiPJy6ZEidgEE7s_8i86K0/edit?usp=sharing).
+
+## Open Questions
+
+- How can we simplify this procedure?
+- How does this integrate with `happa`?
```
puja108, line 54: Tried to jot down some thoughts above: https://github.com/giantswarm/rfc/pull/26/files#r756120105

TLDR: ideally happa does the same as the CLI does, but for gitops-managed clusters it might be read-only or offer a download of the new manifests or even at some point be able to create a PR directly.

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,55 @@
...
+## Client
+
+We use `kubectl-gs template` to create templates for cluster creation. It seems straight-forward to add functionality to handle upgrades as well.
+This should include the ability to review changes, apply the upgrade directly and creating commits to git repositories used for cluster management.
+
+## Example workflow
+
+![Example](upgrade.gif)
+
+See slides [here](https://docs.google.com/presentation/d/1_ImURpdO3T8HxyBNraTsoAiPJy6ZEidgEE7s_8i86K0/edit?usp=sharing).
```
puja108, line 49: on step/slide 2:
- why does kgs need to get resources for y? what does it do with those? From what I see it only needs the label for y or not?
- could there be an option to not get resources for x from the cluster but point the CLI at local manifests? I'm thinking here of a gitops use case where the CLI could for example be run in the local git folder to change/overwrite the files and create a PR

capi-configuration/3_upgrades.md
anvddriesch: We would not force anybody. However, we ourselves want to use this approach and we want to support and encourage customers to use it as well.

capi-configuration/3_upgrades.md
anvddriesch: First of all, we think it would make sense to check whether `y` exists and it's possible to upgrade `x` to `y`. The other reason is what we described above in `mutable fields`.  Whichever component does the upgrades would need to know which fields are mutable in that version so that they can be cleared before submitting the cluster object.

capi-configuration/3_upgrades.md
anvddriesch: I agree that the upgrades should always be the same on a lower level, but that seems to be at odds with our current approach to implement it to `kubectl-gs` directly.
In this case it would be indeed very similar between gitops and non-gitops because we basically generate the proposed changes, review them and then either apply them directly or to a repo.
We can probably replicate the same in happa at least somewhat. (Even if at the beginning we just have to make things read-only for gitops-managed clusters)

In case of the scheduled approach, it's a bit different. We could do the exact same workflow also automated through an operator. For gitops-managed clusters this has to lead to an automated commit to the managing repo. For other clusters it would lead to an automated apply.
But then we would lose the ability to review which IMO is not good. 
So what I'm thinking right now is that rather than the review and actual upgrade being one and the same, it would be good to define upgrades in their own resource. (I believe fer suggested something similar) and have only one component execute the upgrades based on this resource. (How the execution looks like would then differ between gitops and non-gitops as you described)
So using kubectl-gs we would still get to review or apply the upgrade. Reviewing would be as described here. If we apply it, an upgrade resource would be created. Furthermore we could define the upgrade time and include scheduling directly.
Happa would do the same.

capi-configuration/3_upgrades.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Open Questions
+
+- How can we simplify this procedure?
+- How does this integrate with `happa`?
+- ?
```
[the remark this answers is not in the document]
anvddriesch: That's a good point. :) SIG-UX is definitely on the list of people to get comments from as soon as this RFC is  little more refined. (We are still in the stage of discussing internally in the WG)

capi-configuration/3_upgrades.md
puja108: just saw this here, copying my comment from Slack:
one thing that’s on my mind reading the “scheduling everywhere” is that in a gitops approach we'd be basically scheduling a PR, but the intention of an upgrade schedule is to run the upgrade at that time, not the PR which then again needs reviewing. In this context I’m not sure scheduling outside of the actual gitops tool makes sense when using gitops. Maybe in a gitops world the PR would even be created right when the release is GA and then the user “schedules” the upgrade through git and/or gitops tool automation (e.g. bot driven auto-merge, workflows,…)

anvddriesch: Yeah I absolutely agree.
Also paraphrasing from slack, I don't think we need to worry much at all about the scheduling and gitops.
Rather than that it is more that we should make sure all upgrades, scheduled or not have the same source. So if as suggested we use `kubectl-gs` to create and review upgrades which can be used in gitops, then we should also be able to get a scheduled upgrade using the same command/basic mechanism.

PR #27, opened 2022-01-17
Email aliases for smarter ticket management

LolloneS:


JosephSalisbury:
cc @JosephSalisbury

alias-forwarder-email-service/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+In case of non-urgent need, customers can open tickets by sending an email to support@giantswarm.io. This will automatically be forwarded to the #support channel in Slack.
+
+While this works -also thanks to the fact that the tickets are broadcast- we reckon that, as Giant Swarm grows and gathers customers, having such general entry points can become hard to scale and manage.
+
+## Desired improvements
+
+We would like to be able to configure email aliases (where *alias*="some smart way to express and parse a composite email address") with different formats. These aliases should follow an easily parsable and standardized structure, which allows us to redirect the emails to the appropriate target.
+
+For instance, some emails (`urgent`) could be ridirected to Opsgenie, while others (`support`) could be redirected to Slack.
```
JosephSalisbury, line 17: I'd like a bit more on this. Sure, we could have different emails per customers. Why would we want that? What problem does it solve?

LolloneS: I'm still wrapping my head around this and gathering information to try and see whether this would actually be useful.

In my opinion we should be able to solve/improve the following:

- **urgency**: if I understand correctly, an `urgent` email gets forwarded to all the teams/areas simultaneously (https://intranet.giantswarm.io/docs/support-and-ops/processes/oncall). It would be nice to just forward the ticket to the correct team (e.g. if the problem is related to AWS, to the Cloud Area on-caller). Because of this, it would be cool to be able to filter out emails. Again, the idea is to have in somewhere in the email address, but this can change.

- **customer**: we don't really have a use case for this so far, but I can think of some:
  - analytics (therefore understanding what/how we can improve)
  - stopping receiving emails if a customer stops paying
  - maybe prioritizing?

  so we may want to have the information because it *might* be useful in the future. Again, just thinking out loud.

Finally, I will send another update to this PR today with additional questions/information 😄  I'll keep this updated as long as I have news.

JosephSalisbury: oh yeah, being able to dispatch emails directly to the right oncall area would be dope, that's legit

LolloneS:
So, I tested scenario nº2 for the Slack part and it seems to work. I successfully forwarded to Slack only emails with a specific `+loremipsum` suffix. Will try and see whether it is feasible to configure forwards to different channels instead of the "plain" slackbot

LolloneS:
The Email slack app definitely seems to do the job. I'll investigate opsgenie now.

LolloneS:
We can also create assign tickets to different teams in Opsgenie. The Email integration in Opsgenie allows for creating numerous email addresses, each of which will assign the ticket to a specific team.

Therefore, it seems like we can proceed as follows:

- Only use support@giantswarm.io and urgent@giantswarm.io
- Specify area or customer using the + symbol, e.g. support+adidas+kaas@giantswarm.io
- Using Gmail filters, either forward that email to Slack or Opsgenie

LolloneS:
@othylmann @brinker211 @JosephSalisbury any opinions? :)

brinker211:
I like the idea but my main concern is adoption of the specific addresses. It works if each team will have one to do but I wonder if customer has multiple to chose from if they will do the same one. 

Does `urgent@giantswarm.io` still main a default?

I am just thinking about if a customer were to have an issue and don't know the component affected, just that something is broken. Thoughts?

alias-forwarder-email-service/README.md
othylmann: And the customer makes them feel special and it is an additional trigger that allows us to do things. We could auto populate info for escalation through the wiki. We could do a lot of things.

othylmann:
Well, if we only use the +keyword system, the default ist still urgent and support and it just gets filtered through the + system. we need to make sure these filters are easy. we can actually start with having a filter for "+kaasaws" and just tell that to customers and ignore the customer part. That means we just need team filters for opsgenie. 

Then we also have c_adidas as a filter that automatically gets posted to the project channel of adidas. Makes the filters ind of simple because it all ends up on the same email. I think.

LolloneS:
@brinker211 I believe we can keep using "defaults". If something breaks and the customer has no idea what that is, then they can just contact urgent@giantswarm.io. If they see a CF stack failing and stuff behaving strangely on AWS, then they can likely ping urgent+kaasaws@giantswarm.io. Also, I think this is kind of a best effort that gets better with time. If they page the wrong team, the wrong team can simply page the correct one. Considering they would all be woken up anyway, this scenario's worst case basically overlaps with the current status (at least I think so).

@othylmann I like the idea of starting simple. Should we start by configuring `urgent+kaasaws@giantswarm.io`, see how that goes, and then add more?

alias-forwarder-email-service/README.md
```diff
@@ -0,0 +1,98 @@
...
+  - (scenario 1) email address creation/deletion
+  - (both scenarios) "routes": support to slack vs. urgent to opsgenie; different areas
+  I haven't been able to find a sensible solution so far, but we can dig into this a bit more.
+
+# Final thoughts
+
+Considering the pros and cons of each solution, I believe we can proceed as follows:
+
+- Only use the support@giantswarm.io and urgent@giantswarm.io email addresses
+- Specify area and customer using the + symbol, e.g. support+adidas+kaas@giantswarm.io
```
MarcelMue, line 94: My main concern is to get customers to actually follow the pattern here.

pipo02mix: if they have some tooling where they set up the patterns it will work, I know of a customer using Remedy to notify us of an incident, so the actual engineer does not about the email address, (s)he just fill the form and submit

alias-forwarder-email-service/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Proposal #2: just two email addresses
+
+This proposal relies on using the `+` sign trick to correctly redirect requests.
+
+We can define two email addresses: `urgent@giantswarm.io` and `support@giantswarm.io`.
+
+Let's consider customer `shiba`. We can discern between their requests depending on the additional information present after the first `+` sign. Some examples could be:
+
+- `deliveredto:"urgent+shiba+aws@giantswarm.io"`
```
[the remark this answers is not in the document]
LolloneS: I don't know whether it is possible to apply such filters on opsgenie and gmail -- will check tomorrow and let you know. However I am not really worried of "adversarial attacks", we can definitely perform at least some basic filtering before forwarding emails. Finally, I don't think we are currently performing any particular checks, so even growing iteratively and tackling this problem in a bit doesn't sound bad to me

alias-forwarder-email-service/README.md
LolloneS: @MarcelMue I think the idea is to give each customer only "their" email addresses. I wouldn't cross bridges before getting to them - we have time to improve.

LolloneS:
I am wondering whether beginning with a very small subset of suffixes would be a good idea to get started. I think we could at least define _some_ (even just 2 or 3) email addresses and get going.

alias-forwarder-email-service/README.md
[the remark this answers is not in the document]
LolloneS: I think we can filter based on sender. For instance: sender matches `*@adidas.com`. See https://support.google.com/a/users/answer/9300019?hl=en. I wouldn't worry much more than this at the moment,

LolloneS:
Some updates:

- as far as the `support` emails are concerned, we can even only use the customer. For instance, `support+shiba@giantswarm.io` will be forwarded to the `#support-shiba` channel in Slack. This requires a forwarding rule for each customer to be created. The forwarding rule can have a `*` wildcard in case we want to still allow specifying an area but do not want to handle it at the moment. Considering we do not gain/lose customers daily (or even weekly), having a rule per customer is quite manageable even though it needs to be done manually. 
- regarding the `urgent` emails, we believe the important information to be kept are areas and customer. Therefore, either the `urgent+customer+area` or the `urgent+area+customer` structure shall be used. The key point here is to understand _which_ areas are the ones we are interested in. So far we can think of:
  - cloud providers (AWS, Azure, VMWare, ...)
  - packs (security, monitoring, ...)
  but a better list will be provided soon. Something to be checked is in which order forwarding rules are applied (if any).

More on this next week.

LolloneS:
As explained [here](https://support.google.com/mail/thread/122916616/gmail-filter-wildcard?hl=en) wildcards are NOT supported in filters.

This means that, while [support+customer@giantswarm.io](mailto:support+customer@giantswarm.io) is NOT affected, [urgent+customer+area@giantswarm.io](mailto:urgent+customer+area@giantswarm.io) IS affected, because we cannot set up filters matching different/missing customers. For instance, it is impossible to set up a filter such as: _forward emails coming to `urgent+*+aws@giantswarm.io` to the Opsgenie email associated with the oncall AWS/KaaS team_.

In my opinion, we should either:

- limit ourselves (at least for now) to `support+customer@giantswarm.io` (forward to slack channel) and `urgent+area@giantswarm.io` (forward to area oncall person)
- find an external service which offers the possibility of creating forwarding rules with wildcards. I have contacted a service that _may_ be what we are looking for asking for clarification, and I am looking into other services that may do the job.

#### Areas

We need to define which areas we want to identify. These are important for the `urgent` emails, as they can be used to forward urgent requests to the appropriate team.

**Providers**: I think it is better to separate them. We don't know whether in the future we will have different oncall teams for the different cloud providers.

1. `AWS`
2. `Azure`
3. `VMWare`
4. `OpenStack`
5. `GCP`
6. `KVM` (being "dismissed")

**Apps**:  I still need to figure out whether we should keep a "generic" `+apps` suffix or whether we should be more specific. I'll keep them in a single suffix so far.

7. `Apps`

**Cloud native packs**: with the introduction of cloud native packs, we may want to have separate emails for the different packs. Therefore we could define the following suffixes:

8. `Connectivity`
9. `Observability`
10. `Security`
11. `Developer Experience`

alias-forwarder-email-service/README.md
```diff
@@ -0,0 +1,113 @@
...
+  - no, we can't
+# Final decisions
+
+Considering the pros and cons of each solution and the number of comments received, I believe we can proceed as follows:
+
+- Only use the support@giantswarm.io and urgent@giantswarm.io email addresses
+- With `support`, we only specify the customer name, so that the message is forwarded to the customer's Slack channel: `support+customer@giantswarm.io`
+- With `urgent`, we only specify the area, so that the message is forwarded to the correct area in Opsgenie: `urgent+area@giantswarm.io`. The areas are the following:
+
+  Cloud providers:
```
MarcelMue, line 98: I have some concerns about the selection:
It seems to me like this could cause confusion for customers. What would they do if they can't access the k8s API on AWS? set `aws`, `connectivity` or both? I feel like this could lead very easily to customers to always select their cloud provider and never a cloud native pack because they lack the knowledge to determine what the issue is.

LolloneS: @MarcelMue this is a good point. I would say that the distinction, although not really black-white, is that the cloud native packs quite strictly involve applications. An API not reachable during the night seems to me more of an infrastructure problem rather than a problem with some application regarding connectivity (I imagine Kong as part of that pack, for example).

Which categories would you suggest? 

By the way, I recognize that this is a complex problem, but I'd rather start with a first version and improve iteratively than remain stuck trying to get the "perfect" combination. I think this is still an improvement with respect to the current configuration (and I may be wrong 😄).

MarcelMue: One idea would be to have the actual names of technologies instead of the packs. The problem would the amount of combinations then. 
e.g. instead of `connectivity` there is `nginx`, `kong`, `coredns` ...
And `devex` is then `flux` and so on

I just currently feel like a developer / ops person might not have the overview in a stressful situation which pack includes their application / issue. I could be totally wrong here, I am trying to put myself into their shoes basically.

LolloneS: I absolutely agree that they should clearly know which tool belongs to which pack. We could create this many aliases, but I reckon this may become quite unmanageable.

I think we could provide customers with a list along the lines of:

- kong, nginx, coredns, ... -> connectivity
- falco, kyverno, ... -> security
- ...

Which may be some nice common ground.  This also makes it simple to "add/remove" an application, because we just need to update the list (which, for instance, can be in our docs) rather than creating/deleting many aliases.

For instance, in the email template we could add a line with the link to the page with "correspondences" our docs, so that if coredns is broken they just CTRL+F "coredns" and immediately see that that is `connectivity`. Stupid example, but I think it explains what I mean. WDYT?

MarcelMue: That would work for me as a first iteration

brinker211 (approved):
LGTM

PR #28, opened 2022-03-17
Add RFC to rethink deployment of Kyverno policies

MarcelMue:


puja108 (commented):
What about splitting into several repositories but not on an app nor a team level, i.e. splitting based on some kind of defined "domain" such as:
Security:
- PSS
- Additional Security (could also be several)

Non-Security
- Cluster Mgmt
- App Mgmt
- User / Persmission Mgmt

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Have one repository per team
+Each team has one repository where it can put all its policies.
+### Pros
+- Each team has the power to make changes however they see fit
+- Each team can handle their testing setup independently
+- Only a limited number of repositories are changed
+### Cons
+- Components might change ownership and then policies need to be moved as well
+- Teams often own multiple components and therefore adding complexity to the repo
```
stone-z, line 32:
```suggestion
### Cons
- Components might change ownership and then policies need to be moved as well
- Teams often own multiple components and therefore adding complexity to the repo
- Teams need to maintain parallel test and delivery pipelines
```

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Keep all policies in the same repository
+It's simple, everything stays the same and `kyverno-policies` is just renamed.
+### Pros
+- Not a lot of effort
+- Not much communication to make a change
+- Single source of truth for all policies
+### Cons
+- Testing setup is complex as all different environments need to be covered in one repo
+- Ownership is difficult as multiple teams own parts of the repo
+- Upgrades are cumbersome
```
stone-z, line 22: How are upgrades currently cumbersome?

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Proposals
+
+## Keep all policies in the same repository
+It's simple, everything stays the same and `kyverno-policies` is just renamed.
+### Pros
+- Not a lot of effort
+- Not much communication to make a change
+- Single source of truth for all policies
+### Cons
+- Testing setup is complex as all different environments need to be covered in one repo
```
stone-z, line 20: Can you give an example of this in the current state?

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Keep policies together with application
+Each app gets delivered directly with its policies.
+### Pros
+- Apps are tested together with their policies
+- Apps and policies can change ownership together
+- Upgrades can be coordinated easily
+### Cons
+- Policies are split across many repositories
+- Some policies can not be matched to any app
```
stone-z, line 42: Looking at our current policies, virtually none of them map cleanly to an application in my mind

stone-z (commented):
Of the splitting options, I would be +1 on the domain idea. But I don't yet understand the actual need to split the policies to begin with.

It would be nice to keep as few sources of truth as possible. We avoid sprawl, it's always clear where a policy comes from, it is easy to audit, and there's a single history for policy changes. A single repo also enables testing with all policies at once, which is important since testing them (especially mutation/generation policies) in isolation is not representative of the actual cluster state. By splitting them, we'd introduce the overhead of maintaining multiple snowflake policy test setups, none of which actually simulate production.

If there are some pain points with the current setup that I've not experienced, maybe the effort is better spent improving the quality of life there first.

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
+# Handling Kyverno Policy Deployment
+
+Our Kyverno policies have the following properties:
+1. They are owned by different teams
+2. The need to be deployed to different environments
```
AverageMarcus, line 5: Might be worth calling out different providers also.

kyverno-policy-deployment/README.md
AverageMarcus: The current setup is quite different to all our other projects. It leveraged a Python test framework and at times is a struggle to get running. For example, I personally haven't managed to successfully get the tests running on my machine so rely on CI on PRs to check if my changes work or not.

We also don't currently make use of the kyverno CLIs `test` command (mainly because at the time it wasn't able to cover most functionality)

kyverno-policy-deployment/README.md
AverageMarcus: It's not clear _how_ to upgrade and where.

The policies are part of the kyverno app. So do both need new releases? It's not really clear.

It's also not clear where policies get deployed to. (MCs? WCs? All installations?)

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,51 @@
...
+- Some policies can not be matched to any app
+
+## Repository per app and per team
+Have one repository per team as well as delivering policies with apps (essentially combine the previous 2 approaches).
+### Pros
+- All benefits of packaging apps with policies
+- Policies which can not be matched to any app, can live in the team repository
+### Cons
+- There are a lot or repositories
+- Policies are split across many repositories
```
AverageMarcus, line 51: Also has the point above `- Some policies can not be matched to any app`

AverageMarcus (commented):
Additional proposals:

* One repo per provider (e.g. in phoenix we would have 3 - AWS, GCP and Azure)
* A hybrid of some of the current proposals - a repo for "common" policies applied everywhere, repos for team/area/provider specific policies as needed, app specific policies included in the app (with appropriate Helm check to ensure kyverno is installed in the case of WCs)

kyverno-policy-deployment/README.md
stone-z: The python tests are using `ats`, which I agree is still difficult to use, but is theoretically the intended direction for app testing and is completely independent of the policies. That's actually a great example of why I'm hesitant about test quality. Multiple teams would have to duplicate and manage tests they aren't comfortable with. In the current setup, you benefit from existing tests even if you didn't write them.

I also agree it would be nice to use kyverno's own test capabilities going forward. If policies are managed in many places, it will be more difficult to align testing between them and will cause duplicated effort between teams.

kyverno-policy-deployment/README.md
stone-z: The policies are not part of the kyverno app. The app is shipped from https://github.com/giantswarm/kyverno-app and the policies are shipped from https://github.com/giantswarm/kyverno-policies/. The policies are deployed according to which collections they are pushed to, like other apps.

MarcelMue: I still think that there is also a dependency between the two repositories right now. The bigger issue is the whole collection thing though. We are pushing a bunch of charts into different collections and it's just hard to grasp where things end up (see e.g. the confusion between `shared` and `common` policies).

MarcelMue:
I tried to update the RFC with the per domain proposal. Would it make sense for us to have a session some time soon to maybe exclude some approaches we definitely _do not_ want?

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,71 @@
...
+- App Management
+- User Management
+- Provider X specific
+### Pros
+- Domains can be matched to teams
+- Manageable amount of repositories
+- Clear separation of concerns
+### Cons
+- Defining domains is difficult
+- Multiple teams likely feel responsible for the same domain (e.g. KaaS teams for Cluster Management)
```
puja108, line 51: this is another way of saying multi-team ownership of a single repo is painful, maybe we can see what pains there are actually with this and if we could get smh better at that as going forward I'm not sure if we can always ensure single team repo mappings.

stone-z:
I gave this some more brain cycles over the weekend. I accept that at minimum we'll need to split based on WC/MC scope, and in general I'm not blocking splitting the policies in whichever way is most useful to people.

Basically:
- my interest is primarily about policies in WCs. I think most others involved in this discussion are more interested in MCs, which aren't really problematic IMO
- so far, our kyverno usage has been relatively casual in that if it goes down on a management cluster, customers can't create new clusters and we might get woken up, but there's not a lot of other impact. In the near future, _every API call to every workload cluster_ will go through kyverno. Customers will also be applying their own policies on top of ours. That scares me
- testing policies is already very hard

Based on the above, my concern is that by dividing policies we will lose test quality and end up really breaking things _in workload clusters_. My hope was that by keeping the policies centralized, we could continue to re-use existing tests, and all teams would benefit from future tests without having to build their own. But if that doesn't work/make sense, we'll just have to get better at testing kyverno.

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Keep all policies in the same repository
+It's simple, everything stays the same and `kyverno-policies` is just renamed.
+### Pros
+- Not a lot of effort
+- Not much communication to make a change
+- Single source of truth for all policies
+### Cons
+- Testing setup is complex as all different environments need to be covered in one repo
+- Ownership is difficult as multiple teams own parts of the repo
```
teemow, line 23: @MarcelMue during a discussion with @stone-z we agreed that if you mix all kind of use-cases in one repo there is a risk that you will build complex interdependencies of policies. Stacking policies on top of each other and building a hierarchy will be a bad design. Policies split by use-case will be leaner and easier to test with less side-effects. Still @stone-z sees a lot of risks in side-effects if policies are split and not tested as a whole. Imo it would be good to define design principals and best practices that make sure policies are lean (not too complex), narrow in scope and independent.

MarcelMue: Let's discuss tomorrow in the session how we want to progress forward. I want to initially _exclude_ solution proposals and then talk about how we can make the remaining ones work.

Coming up with design principles makes sense to me but it's not something we can do immediately right now IMO. So talking about when & how we come up with them makes sense to me.

MarcelMue:
> I gave this some more brain cycles over the weekend. I accept that at minimum we'll need to split based on WC/MC scope, and in general I'm not blocking splitting the policies in whichever way is most useful to people.
> 
> Basically:
> 
>     * my interest is primarily about policies in WCs. I think most others involved in this discussion are more interested in MCs, which aren't really problematic IMO
> 
>     * so far, our kyverno usage has been relatively casual in that if it goes down on a management cluster, customers can't create new clusters and we might get woken up, but there's not a lot of other impact. In the near future, _every API call to every workload cluster_ will go through kyverno. Customers will also be applying their own policies on top of ours. That scares me
> 
>     * testing policies is already very hard
> 
> 
> Based on the above, my concern is that by dividing policies we will lose test quality and end up really breaking things _in workload clusters_. My hope was that by keeping the policies centralized, we could continue to re-use existing tests, and all teams would benefit from future tests without having to build their own. But if that doesn't work/make sense, we'll just have to get better at testing kyverno.

I understand the concern especially considering MC/WC. I also believe that we need to find a way forward which works in both areas. It seems like most of the pain we feel in the current approach is the MC side, so we could think about different approaches for each. 

Let's definitely discuss tomorrow and exclude the options we do not want for sure.

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,71 @@
+# Handling Kyverno Policy Deployment
+
+Our Kyverno policies have the following properties:
+1. They are owned by different teams
+2. The need to be deployed to different environments & on different providers
```
fiunchinho, line 5:
```suggestion
2. They need to be deployed to different environments & on different providers
```

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,71 @@
...
+3. They required different testing setups according to 1. and 2.
```
fiunchinho, line 6:
```suggestion
3. They require different testing setups according to 1. and 2.
```

kyverno-policy-deployment/README.md
```diff
@@ -0,0 +1,71 @@
...
+- Cluster Management
+- App Management
+- User Management
+- Provider X specific
+### Pros
+- Domains can be matched to teams
+- Manageable amount of repositories
+- Clear separation of concerns
+### Cons
+- Defining domains is difficult
```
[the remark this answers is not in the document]
fiunchinho: Apart from the pains of multi-team ownership of a single repo, I'm afraid having domains that include different providers/environments comes with the same problems that we currently have regarding a complex testing setup

PR #29, opened 2022-03-28
Add RFC for merging configmaps through app platform

MarcelMue:


merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,75 @@
+# Merging Configmaps in a gitops context
```
tfussell, line 1: I believe this applies to both Secrets and ConfigMaps so maybe it would be more accurate to just call it "config" or "app values".

MarcelMue: True - I didn't consider this but actually a really good point.

MarcelMue: Added a disclaimer :)

tfussell:
Could you maybe list alternative solutions or future work that might improve this? Otherwise it looks good.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+  config:
+    configMap:
+      name: flux01-default-apps-config
+      namespace: org-some
+  userConfig:
+    configMap:
+      name: flux01-userconfig
+      namespace: org-some
+```
+The functionality means that values in `config` will be overlayed by values from `userconfig`.
```
tfussell, line 65:
```suggestion
This functionality means that values in `config` will be overlaid by values from `userConfig`.
```

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+So far we only used this functionality to supply Giant Swarm offered configuration through `config` and encouraged end users to only use the `userconfig` field.
```
tfussell, line 66:
```suggestion
So far we have only been using this functionality to supply Giant Swarm offered configuration through `config` and encouraged end users to only use the `userConfig` field.
```

tfussell: Can we link to our public docs showing what we recommend to customers?

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+Our app platform utilizes configmaps / secrets heavily to pass configuration into apps.
+
+*The following RFC will focus on configmaps but the same problem and solution is also applicable for secrets!*
+
+These configmaps are often either shared between multiple instances of the same app or have shared values between apps.
+In gitops the intended solution for both of these cases is to use `bases` and `overwrites` which is achieved through `kustomize`.
+
+## Problem Statement
+
+The content of a configmap is a list of key-value pairs.
```
tfussell, line 12:
```suggestion
The `data` section of a configmap is a mapping of keys to values.
```

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+Therefor configuration (especially nested configuration) is usually grouped under a single key (e.g. `values.yaml`).
```
tfussell, line 13:
```suggestion
Therefore configuration (especially nested configuration) is usually grouped under a single key (e.g. `values`).
```

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
+# Merging config in a gitops context
+
+Our app platform utilizes configmaps / secrets heavily to pass configuration into apps.
+
+*The following RFC will focus on configmaps but the same problem and solution is also applicable for secrets!*
```
tfussell, line 5:
```suggestion
*The following RFC will focus on configmaps but the same problem and solution are also applicable for secrets!*
```

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+- Easy to explain to customers
+### Cons
+- Change in the intended design in app platform
+- Only sidestepping the issue, not really resolving it
+- Only one layer of overwrites possible
+
+## Future work
+
+The ideal solution would be to support the desired usecase in `kustomize` directly.
+The maintainers of `kustomize` have already stated that they would approve of several proposals for this feature.
```
tfussell, line 82: Anything we can link to here in case someone wants to check the current status in the future?

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
...
+      failureDomain: gb-lon-1
+      name: default
+      replicas: 3
+```
+All fields under `values` will be grouped into a simple `string` even though they are valid `yaml`.
+The only way to overwrite would be to overwrite the entire content of the `values` key.
+
+## Context
+
+We ran into this issue hard when we made cluster-API releases more DRY by using bases.
```
tfussell, line 39:
```suggestion
This problem became most apparent while working to make Cluster API releases more DRY by using bases.
```

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,83 @@
+# Merging config in a gitops context
+
+Our app platform utilizes configmaps / secrets heavily to pass configuration into apps.
+
+*The following RFC will focus on configmaps but the same problem and solution is also applicable for secrets!*
+
+These configmaps are often either shared between multiple instances of the same app or have shared values between apps.
+In gitops the intended solution for both of these cases is to use `bases` and `overwrites` which is achieved through `kustomize`.
```
tfussell, line 8: Can you show what this looks like here or is it too much for an RFC?

merging-configmaps-gitops/README.md
MarcelMue: I can't think of anything I could easily link for the implementation. There are several issues upstream tracking the issue such as: https://github.com/kubernetes-sigs/kustomize/issues/3787

merging-configmaps-gitops/README.md
MarcelMue: I decided to simply add a link.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,84 @@
+# Merging config in a gitops context
+
+Our app platform utilizes configmaps / secrets heavily to pass configuration into apps.
+
+*The following RFC will focus on configmaps but the same problem and solution are also applicable for secrets!*
+
+These configmaps are often either shared between multiple instances of the same app or have shared values between apps.
+In gitops the [intended solution](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/#2-create-variants-using-overlays) for both of these cases is to use `bases` and `overwrites` which is achieved through `kustomize`.
```
tfussell, line 8: It looks like `kustomize` calls this feature `overlays` instead of `overwrites`.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+The proposal is to use `config` for cases in gitops where we can not merge configmaps through `kustomize`.
+From preliminary testing this works without any issues **but** is a change to how we treated this feature so far in the app platform.
+
+### Pros
+- Easy to just use
+- Easy to explain to customers
+
+### Cons
+- Change in the intended design in app platform
```
puja108, line 76: tbh, having been part of the design team on this, I don't see it as a big change, it was intended to model 2 levels of config just like an overlay. Our use case for it was having cluster level config in the one level and user config in the 2nd. In the case of MC level usage the cluster level is "free" and we can re-use it for sth else like you are proposing here, and which I've also been hinting at before in Slack.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,84 @@
...
+- Only sidestepping the issue, not really resolving it
+- Only one layer of overwrites possible
```
puja108, line 78: there's. more layers in chart and catalog, but I guess we need to reserve those for internal usage, right?

puja108 (approved):
LGTM

merging-configmaps-gitops/README.md
MarcelMue: Are there cases (such as e.g. `prometheus`) where we already use this for WC apps? Or does this need to be checked on a case by case basis?

merging-configmaps-gitops/README.md
MarcelMue: Yes, currently we reserve them for internal usage.

merging-configmaps-gitops/README.md
MarcelMue: Fixed :+1:

piontec (commented):
I think a short section about 'consequences' would be nice. We talked today and another problem there is with this approach is that it limits any config source, including gitops, to a single overlay layer.

MarcelMue:
> I think a short section about 'consequences' would be nice. We talked today and another problem there is with this approach is that it limits any config source, including gitops, to a single overlay layer.

I believe I already covered that in the `cons` section :)

merging-configmaps-gitops/README.md
rossf7: I agree with Puja this is just using the design for a different use case.

However this will only work for MC apps. For WC apps we always default the cluster values configmap for use cases like setting the base domain. https://docs.giantswarm.io/app-platform/defaulting-validation/#defaulting

This could be changed but changing it for existing WCs apps it not straightforward.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+*The following RFC will focus on configmaps but the same problem and solution are also applicable for secrets!*
+
+These configmaps are often either shared between multiple instances of the same app or have shared values between apps.
+In gitops the [intended solution](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/#2-create-variants-using-overlays) for both of these cases is to use `bases` and `overlays` which is achieved through `kustomize`.
+
+## Problem Statement
+
+The `data` section of a configmap is a mapping of keys to values.
+Therefore configuration (especially nested configuration) is usually grouped under a single key (e.g. `values`).
```
rossf7, line 13: We went with a single values key for simplicity when designing the config layers.

Adding the key name to the userConfig section similar to how Flux does this with the helmrelease CRD could help here. 

As suggested by @ubergesundheit in https://github.com/giantswarm/roadmap/issues/865


https://github.com/giantswarm/roadmap/issues/865

merging-configmaps-gitops/README.md
MarcelMue: But would this mean if we supply a custom `config`, we just have to make sure the required fields are set? E.g. the `base domain`?

rossf7: For MC apps including the default apps bundle we're fine because `.spec.config` is not currently used.

For the child default apps in the WC the config can be passed via `.spec.userConfig` but `.spec.config` is in use.

For WC apps we could set `.spec.config` so it's not defaulted but any values that are in use like the base domain would need to be present.

merging-configmaps-gitops/README.md
```diff
@@ -0,0 +1,84 @@
...
+## Context
+
+This problem became most apparent while working to make Cluster API releases more DRY by using bases.
+All parts of our setup currently rely on `app` CRs and their configuration through configmaps.
+
+Full context of the PR can be found here: https://github.com/giantswarm/workload-clusters-fleet/pull/130
+
+## Solution Proposal
+
+Our app platform offers a layerd configuration approach already which we can utilize to enable us here.
```
pipo02mix, line 46:
```suggestion
Our app platform offers a layered configuration approach already which we can utilize to enable us here.
```

MarcelMue:
It seems like we do no have any real other alternative so far. Is there someone else I should ask for review here or are we fine with merging the RFC soon?

tfussell (approved):

PR #31, opened 2022-04-07
adds RFC for rainbow roadmap for 2022/Q1

PrashantR31:
adds RFC for rainbow roadmap for 2022/Q1


teemow:
@anvddriesch shall we close this RFC? Q1 is already over.

anvddriesch:
Yes!

PR #32, opened 2022-04-11
Added UUID domain RFC

AverageMarcus:
Signed-off-by: Marcus Noble <github@marcusnoble.co.uk>

### DIFF
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -10,8 +10,8 @@ This approach has a few drawbacks:
 
 1. The workload cluster is tied to the management cluster, making migration messy
 2. Workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense
-3. Several customers have their own domain set up, leading to inconsistency between MCs
-4. The provider (and region) is baked into the domain
+3. Several customers have their own domain set up, leading to inconsistency between MCs (e.g. `eu-central-1.aws.cps.customer.com` vs. `gauss.eu-west-1.aws.gigantic.io`)
+4. The provider (and region) is baked into the domain, effectively preventing multi-region (or multi-cloud)
 
 An alternative "flat" DNS structure has been proposed that removes the link between WC and MC. E.g.:
 
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -1,42 +1,60 @@
 # Using UUIDs for cluster endpoint domains
 
+> Assumption: this is relevant only going forward with CAPI clusters and not to be "backported" to the vintage product.
+
+## Current Approach
+
 The current approach for building cluster domains is to include the workload cluster's ID as a subdomain of the management cluster it belongs to, e.g.:
 
 ```
-https://api.j9yr2.k8s.gauss.eu-west-1.aws.gigantic.io
+Workload Cluster: https://api.j9yr2.k8s.gauss.eu-west-1.aws.gigantic.io
+Management Cluster: https://dex.g8s.gauss.eu-west-1.aws.gigantic.io
 ```
+
+The domain contains several parts:
+
+1. The name of the service being accessed (`api`)
+2. The workload cluster's ID, which can be user-provided (`j9yr2`) [Workload Cluster only]
+3. The static string `k8s` or `g8s`
+4. The codename of the management cluster (`gauss`)
+5. The region the cluster is deployed into (`eu-west-1`)
+6. The cloud provider (`aws`)
+7. The static string `gigantic.io`
 
 This approach has a few drawbacks:
 
-1. The workload cluster is tied to the management cluster, making migration messy
-2. Workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense
-3. Several customers have their own domain set up, leading to inconsistency between MCs (e.g. `eu-central-1.aws.cps.customer.com` vs. `gauss.eu-west-1.aws.gigantic.io`)
-4. The provider (and region) is baked into the domain, effectively preventing multi-region (or multi-cloud)
+1. The workload cluster is tied to the management cluster (belonging to it via domain heirarchy) which is problamatic if the cluster needs migrating to another management cluster.
+2. Similar to above, workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense.
+3. Several customers have their own domain set up, leading to inconsistency between MCs (e.g. `eu-central-1.aws.cps.customer.com` vs. `gauss.eu-west-1.aws.gigantic.io`).
+4. The provider (and region) is baked into the domain. While not a major problem today, it does make possibility of multi-region clusters slightly more complex.
+5. DNS Hosted Zones are currently set up in different ways between management clusters and workload clusters (Terraform and `dns-operator-[provider]` respectively)
 
-An alternative "flat" DNS structure has been proposed that removes the link between WC and MC. E.g.:
+## Existing proposal
+
+An alternative "flat" DNS structure has been proposed that removes the link between workload clusters and management clusters by dropping the management cluster codename from the domain and instead using the customer name (or codename).
+
+E.g.:
 
 ```
-https://api.j9yr2.k8s.[customer name].gigantic.io
+https://api.j9yr2.k8s.[customer name].eu-west-1.aws.gigantic.io
 ```
 
-While this solves some of the above drawbacks (1, 2, 4) it introduces its own complications:
+While this solves some of the above drawbacks (1 & 2) it introduces its own complications:
 
-1. The customer name (or codename) is exposed in the domain
-2. There is no enforced global uniqueness on cluster IDs so if a customer has multiple MCs there is a chance that two workload clusters have the same ID, leading to conflicting domains
+1. The customer name (or codename) is exposed in the domain (not a blocker, but to avoid unintentionally shareing information about our customers we'd need to keep these URLs out of any issues or PR comments etc.)
+2. There is no enforced global uniqueness on cluster IDs so if a customer has multiple management clusters there is a chance that two workload clusters have the same ID, leading to conflicting domains. This is especially true when customers set their own cluster IDs.
 
-## Proposal
+## Updated Proposal
 
-> Assumption: this is relevant only going forward with CAPI clusters and not to be "backported" to the vintage product.
-
-The flat DNS structure can be improved to avoid the drawbacks above my leveraging a UUID for the cluster domain and dissasociate the cluster ID/Name from URLs.
+The flat DNS structure can be improved to avoid the potential conflicts by leveraging a UUID for the cluster domain and dissasociate the cluster ID/Name from URLs.
 
 E.g.
 
 ```
-https://api.4c6aba87-061c-458d-91b6-62ad07979728.k8s.gigantic.io
+https://api.4c6aba87-061c-458d-91b6-62ad07979728.eu-west-1.aws.gigantic.io
 ```
 
-For this to work, we'd need to generate a new UUID for a cluster either when running `kubectl-gs template cluster` (for use cases where GitOps are used) or via some defaulting logic when clusters CRs are applied.
+For this to work we'd need to enforce creation of a new UUID for every cluster on creation.
 
 Cluster names and descriptions can still be used for labels and annotations to provide human-recognisable identifiers when using tools like Happa and kubectl.
 
@@ -45,16 +63,17 @@ This approach is also the same as used by many of the cloud providers for their 
 The benefits of this approach:
 
 * Clusters are no longer permanently associated with a specific management cluster
-* Clusters can be "renamed" (e.g. a cluster that was previously a dev enviroment being promoted to production) and moved (e.g. between cloud provider regions) without changes needed to the domain
-* Infrastructure / environment details are no longer needed/exposed in the domain
+* The human-friendly cluster name and description are no longer bound to the limitations of DNS. (e.g. Length and valid characters)
 * No potentially sensitive information can be leaked from the domain alone (so should be safe to use in GitHub tickets and similar)
 * Conflicts in names are avoided by the use of UUIDs, even across customers and providers
-* Workload clusters can more easily be "promoted" to management clusters by just installing the required Giant Swarm applications.
+* Workload clusters can more easily be "promoted" to management clusters by just installing the required Giant Swarm applications
 
-For customers that require the use of their own domains, some additional setup will be required to ensure the Kubernetes API's certificate is also valid for the customers domain (and then a CNAME can be set up to point to the UUID-based domain).
+The drawbacks:
+* For customers that require / desire the use of their own domains, some additional setup will be required to ensure the Kubernetes API's certificate is also valid for the customers domain (and then a CNAME can be set up to point to the UUID-based domain)
+* The domain name no longer resembles to the human-friendly name of a cluster
+* The domain alone doesn't give any indication of which customer it belongs
 
 ## Topics that need expanding
 
-* It's currently unclear how and where hosted zones will be created.
-* Our tooling (kubectl-gs, happa, etc.) will need to be updated to better surface information that may be lost in this approach (such as the region a cluster is deployed to).
-* With a GitOps approach, how would we ensure uniqueness of UUIDs?
+* The mechanism for managing the hosted zone delegation is currently unclear
+* With a GitOps approach, how would we ensure uniqueness / creation of UUIDs? - A possible solution could be to have a validating webhook prevent creation of clusters with the domain set and a mutating webhook that generates the UUID and sets it. Subsequent updates will need to ensure the domain isn't modified.
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -67,7 +67,6 @@ The benefits of this approach:
 * No potentially sensitive information can be leaked from the domain alone (so should be safe to use in GitHub tickets and similar)
 * Conflicts in names are avoided by the use of UUIDs, even across customers and providers
 * Workload clusters can more easily be "promoted" to management clusters by just installing the required Giant Swarm applications
-* The generation process for the flexible part (the UUID part) is well documented. Also creation can be done on any tier and does not have to be done by a centralized authority. This enables a workflow where the cluster creator determines the endpoint URL upfront if desired.
 
 The drawbacks:
 
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -23,7 +23,7 @@ The domain contains several parts:
 
 This approach has a few drawbacks:
 
-1. The workload cluster is tied to the management cluster (belonging to it via domain heirarchy) which is problamatic if the cluster needs migrating to another management cluster.
+1. The workload cluster is tied to the management cluster (belonging to it via domain hierarchy) which is problematic if the cluster needs migrating to another management cluster.
 2. Similar to above, workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense.
 3. Several customers have their own domain set up, leading to inconsistency between MCs (e.g. `eu-central-1.aws.cps.customer.com` vs. `gauss.eu-west-1.aws.gigantic.io`).
 4. The provider (and region) is baked into the domain. While not a major problem today, it does make possibility of multi-region clusters slightly more complex.
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -46,7 +46,7 @@ While this solves some of the above drawbacks (1 & 2) it introduces its own comp
 
 ## Updated Proposal
 
-The flat DNS structure can be improved to avoid the potential conflicts by leveraging a UUID for the cluster domain and dissasociate the cluster ID/Name from URLs.
+The flat DNS structure can be improved to avoid the potential conflicts by leveraging a UUID (v1) for the cluster domain and dissasociate the cluster ID/Name from URLs.
 
 E.g.
 
@@ -54,7 +54,7 @@ E.g.
 https://api.4c6aba87-061c-458d-91b6-62ad07979728.eu-west-1.aws.gigantic.io
 ```
 
-For this to work we'd need to enforce creation of a new UUID for every cluster on creation.
+For this to work we'd need to enforce creation of a new UUID for every cluster on creation. The use of UUID v1 ensure uniqueness.
 
 Cluster names and descriptions can still be used for labels and annotations to provide human-recognisable identifiers when using tools like Happa and kubectl.
 
--- a/uuid-cluster-dns/README.md
+++ b/uuid-cluster-dns/README.md
@@ -1,0 +1,49 @@
+# Using UUIDs for cluster endpoint domains
+
+The current approach for building cluster domains is to include the workload cluster's ID as a subdomain of the management cluster it belongs to, e.g.:
+
+```
+https://api.j9yr2.k8s.gauss.eu-west-1.aws.gigantic.io
+```
+
+This approach has a few drawbacks:
+
+1. The workload cluster is tied to the management cluster, making migration messy
+2. Workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense
+3. Several customers have their own domain set up, leading to inconsistency between MCs
+4. The provider (and region) is baked into the domain
+
+An alternative "flat" DNS structure has been proposed that removes the link between WC and MC. E.g.:
+
+```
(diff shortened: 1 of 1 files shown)


uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```
+https://api.j9yr2.k8s.gauss.eu-west-1.aws.gigantic.io
+```
+
+This approach has a few drawbacks:
+
+1. The workload cluster is tied to the management cluster, making migration messy
+2. Workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense
+3. Several customers have their own domain set up, leading to inconsistency between MCs
```
marians, line 13: Can you please illustrate this with a (fake) example? It's not clear to me what this means.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+4. The provider (and region) is baked into the domain
```
marians, line 14: Ideally you'd explain why that is a problem.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+An alternative "flat" DNS structure has been proposed that removes the link between WC and MC. E.g.:
+
+```
+https://api.j9yr2.k8s.[customer name].gigantic.io
+```
+
+While this solves some of the above drawbacks (1, 2, 4) it introduces its own complications:
+
+1. The customer name (or codename) is exposed in the domain
+2. There is no enforced global uniqueness on cluster IDs so if a customer has multiple MCs there is a chance that two workload clusters have the same ID, leading to conflicting domains
```
MarcelMue, line 25: I think this is a problem but to me it reveals a recurring issue:
We wanted `cluster IDs` to be a unique and "random" identifier. But users wanted to be able to control what it would look like and add meaning to them. Also to make it easier for automation to predict the ID.

Now we are introducing another random identifier - how can we prevent the same issues from reappearing?

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Proposal
+
+The flat DNS structure can be improved to avoid the drawbacks above my leveraging a UUID for the cluster ID rather than a user-providable short string.
+
+E.g.
+
+```
+https://api.4c6aba87-061c-458d-91b6-62ad07979728.k8s.gigantic.io
```
MarcelMue, line 34: What would the impact on automation be? E.g. how could a user easily find this URL in e.g. some automation pipeline?

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+For this to work, we'd need to remove the ability for user-provided cluster IDs and encourage the use of descriptions instead. If cluster names are still desired we could introduce the concept similar to how the description is (label or annotation).
+
+This approach is also the same as used by many of the cloud providers for their managed Kubernetes offerings.
+
+The benefits of this approach:
```
MarcelMue, line 41: FTR I agree with all these benefits, just wanted to also voice some concerns.

uuid-cluster-dns/README.md
AverageMarcus: Added.

uuid-cluster-dns/README.md
AverageMarcus: Done :)

uuid-cluster-dns/README.md
AverageMarcus: Isn't that what the `kubectl-gs template` command is for?

> But users wanted to be able to control what it would look like and add meaning to them.

This is the purpose of the description annotation isn't it?

uuid-cluster-dns/README.md
AverageMarcus: How do they currently find the kubeconfig?

MarcelMue: But doesn't this have impact beyond the kubeconfig - e.g. for ingress names?

Finding the kubeconfig is usually done through a CLI where it's not a big issue. Maybe I am overthinking it here

uuid-cluster-dns/README.md
MarcelMue: Still customers choose IDs such as `peu01` as a specific means to give them reason.

AverageMarcus: There's no reason that names / IDs can't still be a thing. Just under the hood every cluster is given a UUID to be used for things that really need to be unique like DNS.

marians:
### Some questions

- As the example uses the domain `gigantic.io`, is the assumption that this will be the only domain for clusters? Or are we also planning to accommodate customers who want to have ownership over the domain used?
- Are we assuming that there will be a DNS zone `.k8s.gigantic.io` hosted in a central location where we have to register a subdomain for every new cluster, independent of the provider they use?
- Assuming `gigantic.io` is _not_ the only domain we consider using, how important is it for us to be able to move clusters between domains? And if it is important, how important is it then to keep the identifier part (here: the UUID) in the host name unchanged while moving between domains? (My guess is that moving clusters between domains will not be a feature we consider supporting because of the complexity.)
- Is the idea that the creator of the cluster can simply define the UUID and construct the DNS name from that, without any central check for existence? (If yes, this should be made explicit in the document.)

### UX considerations

#### Length

You're proposing `HOSTNAME.IDENTIFIER.k8s.gigantic.io` where IDENTIFIER is a UUID.

A UUID comes with the promise to practically never collide with any other UUID. To achieve that, it has to be pretty long. And this length makes it, just in general, hard to deal with. (Longer strings are harder to communicate about. People are more likely to miss a difference when comparing them. They need more screen real estate to be displayed. Just to name a few.)

However, we don't need that global uniqueness guarantee. If all clusters have to register a subdomain under `.k8s.gigantic.io` or some other domain anyway, we only have to find an identifier that's not used in that scope yet.

As an alternative for a UUID I would propose a "random string" similar in format to what we do currently for defaulted cluster names. E. g. 5 or even 7 characters using both lowercase letters and numbers, where the first one must be a letter. At 7 chars this gives billions of permutations. Since there is the ultimate uniqueness check in the DNS management, we don't have to worry about the unlikely case of a collision.

#### Relationship to other identifiers

My understanding is that the proposed identifier (UUID part) is only for DNS purposes. So the cluster name (currently: the installation codename or the workload cluster name) will also still exist, just because a Cluster CR needs a name. Correct? We are basically making the DNS part independent from the cluster name.

Having DNS independent from the cluster name for me has the benefit that the public DNS name does not expose the cluster name that, now that customers set this consciously, more and more often carries some internal semantics.

### Misc

Can UUIDs start with a number? Because DNS subdomains must not start with a number.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+2. There is no enforced global uniqueness on cluster IDs so if a customer has multiple MCs there is a chance that two workload clusters have the same ID, leading to conflicting domains
+
+## Proposal
+
+The flat DNS structure can be improved to avoid the drawbacks above my leveraging a UUID for the cluster ID rather than a user-providable short string.
+
+E.g.
+
+```
+https://api.4c6aba87-061c-458d-91b6-62ad07979728.k8s.gigantic.io
```
stone-z, line 34: I'm not super opinionated on this so this isn't intended as opposition. There are nice technical/functional upsides of the UUID approach.

Just adding here - it's not very nice to look at or work with as a human. Granted, we try to minimize manual touches and wrap these details in CLI sugar, but in cases where someone has to type this manually, or where they encounter a UUID and have to figure out which cluster it belongs to, this isn't great since they're unordered and opaque by design. I also sometimes benefit from seeing some of those details (e.g. region) in the base domain. Maybe others do as well.

So I'm not opposing UUIDs. Most of the downsides could be mitigated with `kgs` features, so maybe it's worth simultaneously opening an issue for tracking usability improvements if this is adopted.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+For this to work, we'd need to remove the ability for user-provided cluster IDs and encourage the use of descriptions instead. If cluster names are still desired we could introduce the concept similar to how the description is (label or annotation).
+
+This approach is also the same as used by many of the cloud providers for their managed Kubernetes offerings.
+
+The benefits of this approach:
+
+* Clusters are no longer tied to a specific management cluster
+* Clusters can be "renamed" and moved without changes needed to the domain
```
marians, line 44: Let's make clear that renaming clusters is currently not a goal. UX wise, we currently have a cluster identifier (also the name of the Cluster resource) that is understood as immutable. In addition we offer the description which can be modified at will and labels which are also editable. Being able to rename clusters would be hard to justify IMO.

uuid-cluster-dns/README.md
marians: I think we'd have to place this in the Cluster CR somewhere, e. g. as an annotation. Clients like happa have to expose this to the end user e. g. in order to show how to get access to the Kubernetes endpoint.

Also this is related to a question I had: Should the creator of a cluster be able to specify the DNS name?

uuid-cluster-dns/README.md
AverageMarcus: I kinda agree with your points but also agree these should be dealt with through tooling and we shouldn't rely on a domain to give insight into specifics (as we already have customer-provided domains that don't follow that convention)

uuid-cluster-dns/README.md
AverageMarcus: We have seen instances where the ability to rename a cluster would be beneficial to a customer though. E.g. "promoting" a dev cluster to a production one. But I'm not suggesting that to be used as a driving force in any way.

AverageMarcus:
@marians I'll try and answer your questions as best I can here:

> As the example uses the domain gigantic.io, is the assumption that this will be the only domain for clusters? Or are we also planning to accommodate customers who want to have ownership over the domain used?

This is covered in the [last line](https://github.com/giantswarm/rfc/pull/32/files#diff-d6657aa0950afab037b900049a8ac14ddfd859f73e564ae92021a03255c1a52bR49) and would be a little more complex to set up but should be possible.

> Are we assuming that there will be a DNS zone .k8s.gigantic.io hosted in a central location where we have to register a subdomain for every new cluster, independent of the provider they use?

This is a good question. I'm not all that clear on how the zones are currently laid out but I assume that _someone_ there is a root that delegates as needed. Maybe we need to just use the root, I'm not sure. Hopefully someone with more knowledge of this area can provide a better answer.

> Assuming gigantic.io is not the only domain we consider using, how important is it for us to be able to move clusters between domains? And if it is important, how important is it then to keep the identifier part (here: the UUID) in the host name unchanged while moving between domains? (My guess is that moving clusters between domains will not be a feature we consider supporting because of the complexity.)

Keeping the identifier and domain when moving would be very important as we specifically have the case of wanting the ability to migrate a workload cluster to become a new management cluster if needed. 

> Is the idea that the creator of the cluster can simply define the UUID and construct the DNS name from that, without any central check for existence? (If yes, this should be made explicit in the document.)

No, I'm thinking of a process similar to how the cloud providers do it - you ask for a new cluster, it goes and creates one along with generating a new UUID based domain.

> However, we don't need that global uniqueness guarantee. If all clusters have to register a subdomain under .k8s.gigantic.io or some other domain anyway, we only have to find an identifier that's not used in that scope yet.

That doesn't work in the situation where a WC wants to migrate to a different MC (or different region or however we group the subdomain)

> As an alternative for a UUID I would propose a "random string" similar in format to what we do currently for defaulted cluster names. E. g. 5 or even 7 characters using both lowercase letters and numbers, where the first one must be a letter. At 7 chars this gives billions of permutations. Since there is the ultimate uniqueness check in the DNS management, we don't have to worry about the unlikely case of a collision.

The problem here is atomic operations, the time between a domain being assigned and a domain being created (in the context of us creating a cluster with it) could be fairly long and it's totally possible for a conflict to arise within that time. 

> Having DNS independent from the cluster name for me has the benefit that the public DNS name does not expose the cluster name that, now that customers set this consciously, more and more often carries some internal semantics.

Exactly! I like this level of seperation.

> Can UUIDs start with a number? Because DNS subdomains must not start with a number.

Yes they (subdomains) can. Kubernetes resources can't.

uuid-cluster-dns/README.md
AverageMarcus: I'm not sure. What issues specifically are you thinking?

I imagine that if you have a kubeconfig then all other endpoints you need can be queried from the kube API. Not the nicest if doing manually but can be hidden away with some tooling perhaps.

uuid-cluster-dns/README.md
AverageMarcus: In fact, based on what @marians said it might actually be beneficial to have them as different things.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,49 @@
...
+The current approach for building cluster domains is to include the workload cluster's ID as a subdomain of the management cluster it belongs to, e.g.:
+
+```
+https://api.j9yr2.k8s.gauss.eu-west-1.aws.gigantic.io
+```
+
+This approach has a few drawbacks:
+
+1. The workload cluster is tied to the management cluster, making migration messy
+2. Workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense
```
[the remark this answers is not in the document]
AverageMarcus: Yes, it's something that we'd like with CAPI. (It's prob best to take this whole document as being for CAPI as it's unlikely we'll update vintage to match)

uuid-cluster-dns/README.md
teemow: Just for context. The customer domains are old kvm installations that will be replaced with new installations and then turned off. There is only one (the oldest) AWS installation that uses a customer provided domain. We've stopped doing this because it is very painful.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+The flat DNS structure can be improved to avoid the drawbacks above my leveraging a UUID for the cluster domain and dissasociate the cluster ID/Name from URLs.
+
+E.g.
+
+```
+https://api.4c6aba87-061c-458d-91b6-62ad07979728.k8s.gigantic.io
+```
+
+For this to work, we'd need to generate a new UUID for a cluster either when running `kubectl-gs template cluster` (for use cases where GitOps are used) or via some defaulting logic when clusters CRs are applied.
```
teemow, line 39: `kubectl-gs template cluster` will generate an AppCR and the configmap with this values.yaml, correct? 
https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.yaml

Cluster CRs won't get applied against the API directly anymore.

Ok, I've found it here for openstack: https://github.com/giantswarm/kubectl-gs/blob/891b222d7f003731005f09d95e86e236cabfa1f5/cmd/template/cluster/provider/capo.go#L28

@marians does it make sense to generate the unique identifier in the values yaml? I guess it needs to be this way because the AppCR AppName needs to be unique too and at least at the moment is generated by the cluster name within `kubectl gs` already.

https://github.com/giantswarm/kubectl-gs/blob/891b222d7f003731005f09d95e86e236cabfa1f5/cmd/template/cluster/provider/capo.go#L117

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Cluster names and descriptions can still be used for labels and annotations to provide human-recognisable identifiers when using tools like Happa and kubectl.
+
+This approach is also the same as used by many of the cloud providers for their managed Kubernetes offerings (e.g. EKS).
+
+The benefits of this approach:
+
+* Clusters are no longer associated with a specific management cluster
```
teemow, line 47: They are :smile:. But not in a hierarchical DNS structure.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,60 @@
...
+* Infrastructure / environment details are no longer needed/exposed in the domain
+* No potentially sensitive information can be leaked from the domain alone (so should be safe to use in GitHub tickets and similar)
+* Conflicts in names are avoided by the use of UUIDs, even across customers and providers
+* Workload clusters can more easily be "promoted" to management clusters by just installing the required Giant Swarm applications.
+
+For customers that require the use of their own domains, some additional setup will be required to ensure the Kubernetes API's certificate is also valid for the customers domain (and then a CNAME can be set up to point to the UUID-based domain).
+
+## Topics that need expanding
+
+* It's currently unclear how and where hosted zones will be created.
```
teemow, line 58: This is important to clarify. Especially from a security point of view we can not simply make this flat in terms of regions/installations. It would be good to discuss with sig-security imo and at best with our customers as well.

Btw the topic just came up here: https://github.com/giantswarm/roadmap/issues/1037#issuecomment-1096889203

uuid-cluster-dns/README.md
AverageMarcus: Good point. I've updated to make it clearer.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Proposal
+
+> Assumption: this is relevant only going forward with CAPI clusters and not to be "backported" to the vintage product.
+
+The flat DNS structure can be improved to avoid the drawbacks above my leveraging a UUID for the cluster domain and dissasociate the cluster ID/Name from URLs.
+
+E.g.
+
+```
+https://api.4c6aba87-061c-458d-91b6-62ad07979728.k8s.gigantic.io
```
alex-dabija, line 36: My understanding of the flat DNS structure was that we would still have a hosted zone per customer. At least that was what @kopiczko proposed if I'm not mistaken. This would still allow us to convert a workload cluster into a management cluster for the same customer.

We in Phoenix talked about slightly changing this by having a hosted zone per customer and cloud provider because otherwise we would have to do cross cloud DNS configuration, which is rather complicated (e.g. credentials management).

AverageMarcus: The problem with that though is you're (potentially) exposing the customer name and it's then an issue if the customer changes for any reason (I believe we've seen this with one of our customers haven't we?)

uuid-cluster-dns/README.md
puja108: I think for UX purposes we can still have names/IDs, and UUIDs are then only for DNS. AFAICS customers do not really look at the DNS of their APIs, but rather care about the context name in the kubeconfig, but we can verify that assumption.

marians:
Currently one problem I have with this draft is that this RFC is mentioning some goals and possibilities which I'm not even sure are a valid goals to pursue, or would at least require more discussion to verify.

For example: getting rid of the cluster name (which is currently immutible), or being able to rename a cluster. This is a big thing to me which might have a lot of consequences. I don't want this sort of thing to slip in as a sub-goal with an RFC on the DNS schema.

What's the main goal of this initiative? Is it to treat management clusters and workload clusters the same in terms of DNS, so that conversion of an MC into a WC is possible, and vice versa?

uuid-cluster-dns/README.md
pipo02mix: but we dont need to put the real name there, just a unique name that serves to identify and discriminate between customers

AverageMarcus:
Updated with the following changes:
* Removed the mentions of renaming clusters (cc @marians)
* Highlighted the current differences between WC and MC domains
* Specifically mentioned each part of the current DNS structure
* Listed the drawbacks of the proposed approach
* Added a possible solution for the GitOps approach.

uuid-cluster-dns/README.md
kopiczko: Do we have any requirements beyond `opsctl` and `kubectl gs` working?

AverageMarcus: Happa also. But not sure beyond that.

uuid-cluster-dns/README.md
kopiczko: We were talking about "customer ID" which isn't really related to the name. We should also be able to assign more than one domain ID per customer.

uuid-cluster-dns/README.md
marians: Yeah, bottom line: DNS and cluster name (as an identifier) are separate things.

uuid-cluster-dns/README.md
marians: @AverageMarcus In happa we explain how to get a workload cluster's ingress base domain. See https://happa.g8s.gollum.westeurope.azure.gigantic.io/organizations/giantswarm/clusters/c68pn/ingress for example:

image

As long as we can derive the base domain from the workload clusters Kubernetes API endpoint, and as long as we can find that one in the Cluster CR, I think we are fine. And, in contrast to what I wrote above regarding an annotation, I'm assuming this will always be found in the cluster CR's `.spec.controlPlaneEndpoint.host` field.

AverageMarcus: Oh that's very nice! :D

AverageMarcus:
Does anyone have any outstanding comments on this? If not I'd like to get it merged.

teemow:
Did you refine the security model? Let's have consensus about the new structure before continuing please.

AverageMarcus:
Ah you're right. No, I didn't get any input from anyone around that.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,79 @@
...
+2. The workload cluster's ID, which can be user-provided (`j9yr2`) [Workload Cluster only]
+3. The static string `k8s` or `g8s`
+4. The codename of the management cluster (`gauss`)
+5. The region the cluster is deployed into (`eu-west-1`)
+6. The cloud provider (`aws`)
+7. The static string `gigantic.io`
+
+This approach has a few drawbacks:
+
+1. The workload cluster is tied to the management cluster (belonging to it via domain heirarchy) which is problamatic if the cluster needs migrating to another management cluster.
```
marians, line 26:
```suggestion
1. The workload cluster is tied to the management cluster (belonging to it via domain hierarchy) which is problematic if the cluster needs migrating to another management cluster.
```

AverageMarcus: Thank you! 😁

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,79 @@
...
+2. Similar to above, workload clusters cannot be "promoted" to management clusters as the domain would no longer make sense.
+3. Several customers have their own domain set up, leading to inconsistency between MCs (e.g. `eu-central-1.aws.cps.customer.com` vs. `gauss.eu-west-1.aws.gigantic.io`).
```
marians, line 28: From @teemow 's comment I understood that we should rather not strive for supporting custom domains in the future.

uuid-cluster-dns/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+This approach is also the same as used by many of the cloud providers for their managed Kubernetes offerings (e.g. EKS).
+
+The benefits of this approach:
+
+* Clusters are no longer permanently associated with a specific management cluster
+* The human-friendly cluster name and description are no longer bound to the limitations of DNS. (e.g. Length and valid characters)
+* No potentially sensitive information can be leaked from the domain alone (so should be safe to use in GitHub tickets and similar)
+* Conflicts in names are avoided by the use of UUIDs, even across customers and providers
+* Workload clusters can more easily be "promoted" to management clusters by just installing the required Giant Swarm applications
```
marians, line 69: I would add another benefit:

- The generation process for the flexible part (the UUID part) is well documented. Also creation can be done one any tier and does not have to be done by a centralized authority. This enables a workflow where the cluster creator determines the endpoint URL upfront if desired.

marians:
Do we care which version (1 ... 4) of UUID is used? If yes, that should be added to the document.

uuid-cluster-dns/README.md
AverageMarcus: I'm not sure we'll be able to allow that. Otherwise a customer could pick any UUID they want (which could be already used by another customer). The creation of the UUID would need to be within our control.

uuid-cluster-dns/README.md
AverageMarcus: Updated to state that custom domains are a non-goal. 👍

AverageMarcus:
@marians  Updated the doc to state the use of UUID v1 to ensure uniqueness

marians:
Does v4 not ensure uniqueness?

AverageMarcus:
Doesn't guarantee it, no. This has a pretty nice comparison: https://www.sohamkamani.com/uuid-versions-explained/

marians:
Interesting.

Based on that I would think v4, being more random, but not encoding any data regarding the originating machine or creation time, discloses less and is definitely more than unique enough in a practical sense.

PR #33, opened 2022-04-13
Automatic App Upgrades RFC

kubasobon:
Towards https://github.com/giantswarm/roadmap/issues/391

I recommend reading the RFC [like this](https://github.com/giantswarm/rfc/blob/auto-app-updates/automatic-app-upgrades/README.md), so you can see the graph and use the links.

### DIFF
--- a/automatic-app-upgrades/README.md
+++ b/automatic-app-upgrades/README.md
@@ -20,7 +20,7 @@ We are currently using CircleCI to publish Helm charts to designated GitHub repo
 
 Flux operators, specifically Image Reflector Controller and Image Automation Controller, can monitor image registries for new tags. Then, based on automation rules, they can update Git repositories by making commits to a specified branch. This allows us to monitor OCI registries and update Git repositories with a new version automatically.
 
-We could set up rules for Flux controllers to update App CRs in a repository whenever a new chart version (matching the automation rule) is pushed to the Azure Container Registry. Then Flux GitOps controllers would see the update in the repository and apply it to the cluster, effectively updating the App CR with the new version. All of this would be triggered automatically after a new release is crafted for any of the apps.
+We could set up rules for Flux controllers to update App CRs in a repository whenever a new chart version (matching the automation rule) is pushed to the Azure Container Registry. Then Flux GitOps controllers would see the update in the repository and apply it to the cluster, effectively updating the App CR with the new version. All of this would be triggered automatically after a new release is crafted for any of the apps. The updated App version would be committed to the branch watched by GitOps controllers so that the update happens automatically.
 
 We acknowledge some of the Apps could have specific requirements regarding versioning, be it freezing the version, following just one major release's minors and patches, etc. It is solvable by having a custom automation setting, as permitted by [ImagePolicy's Policy section](https://fluxcd.io/docs/components/image/imagepolicies/#policy). For example:
 
--- a/automatic-app-upgrades/README.md
+++ b/automatic-app-upgrades/README.md
@@ -1,0 +1,35 @@
+# Automatic App upgrades
+
+## Solution proposal
+
+We propose to use OCI registries, Flux controllers and GitHub repositories to fully automate App upgrades.
+
+## Motivation
+
+We would like to automate App upgrades. Since CAPI clusters will be packaged as Apps as well, this would, by extension, allow us to automate their upgrades as well. This would serve both our customers, as well as unblock some of the epics and stories, like pull-based app collections.
+
+## Solution in detail
+
+Use OCI registries to hold Helm charts. Use GitOps, in particular Flux's Image Automation Controllers, to monitor the registry and upgrade apps when a newer version of the chart is pushed.
+
+Since version 3.8.0, Helm [enables storing charts in OCI registries](https://helm.sh/blog/storing-charts-in-oci/). OCI registries, or Open Contaniners Initiative registries, allow storing arbitrary artifacts other than containers in them using a common storage standard.
+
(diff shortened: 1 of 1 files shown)


automatic-app-upgrades/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Flux operators, specifically Image Reflector Controller and Image Automation Controller, can monitor image registries for new tags. Then, based on automation rules, they can update Git repositories by making commits to a specified branch. This allows us to monitor OCI registries and update Git repositories with a new version automatically.
+
+We could set up rules for Flux controllers to update App CRs in a repository whenever a new chart version (matching the automation rule) is pushed to the Azure Container Registry. Then Flux GitOps controllers would see the update in the repository and apply it to the cluster, effectively updating the App CR with the new version. All of this would be triggered automatically after a new release is crafted for any of the apps.
+
+We acknowledge some of the Apps could have specific requirements regarding versioning, be it freezing the version, following just one major release's minors and patches, etc. It is solvable by having a custom automation setting, as permitted by [ImagePolicy's Policy section](https://fluxcd.io/docs/components/image/imagepolicies/#policy).
```
puja108, line 25: Cool bonus is that OCI registries also support tags and the imagepolicy can filter by tag, so we could use that to even have certain (pre-)releases that only get rolled to test clusters or sth.

automatic-app-upgrades/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+![Graph showcasing the described solution](https://user-images.githubusercontent.com/4587658/163184967-d8fa5e6b-18ac-42e5-bf6c-7fd8e7df3ab6.png)
+
+## Future work
+
+As mentioned previously, `app-operator` and `chart-operator` are ready to interact with OCI registries. Various registry providers have been tested as part of [OCI registry support for App Platform](https://github.com/giantswarm/roadmap/issues/391) issue. The viability of the proposed solution has been tested and confirmed as [a part of the same issue](https://github.com/giantswarm/roadmap/issues/391#issuecomment-1096522248).
+
+The remaining bits:
+- Setting up a repository holding App CRs
```
puja108, line 34: isn't that charts?

automatic-app-upgrades/README.md
puja108: That might solve our "promotion" use case, as a promotion could then be just adding another tag to an existing OCI artifact

kubasobon: Exactly. There's a lot of flexibility in what gets rolled when using the ImagePolicies.

kubasobon: And OCI registries are just regular image registries underneath, so all tagging options still apply.

automatic-app-upgrades/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+We propose to use OCI registries, Flux controllers, and GitHub repositories to fully automate App upgrades.
+
+## Motivation
+
+We would like to automate App upgrades. Since CAPI clusters will be packaged as Apps as well, this would, by extension, allow us to automate their upgrades as well. This would serve our customers, as well as unblock some of the epics and stories, like pull-based app collections.
+
+## Solution in detail
+
+Use OCI registries to hold Helm charts. Use GitOps, in particular Flux's Image Automation Controllers, to monitor the registry and upgrade apps when a newer version of the chart is pushed.
```
MarcelMue, line 13: Can you also link to the Automation controller repo here?

automatic-app-upgrades/README.md
MarcelMue: Can you also add some examples here e.g. a policy which only tracks patch versions or which only tracks patches + minors?

automatic-app-upgrades/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Since version 3.8.0, Helm [enables storing charts in OCI registries](https://helm.sh/blog/storing-charts-in-oci/). OCI registries, or Open Containers Initiative registries, allow storing arbitrary artifacts other than containers in them using a common storage standard.
+
+One of the registries that are fully conformant with the OCI specification is Azure Container Registry. The registry we currently use as a mirror or backup to our other registries.
```
MarcelMue, line 17: Let's explicitly state that this would also mean we use ACR in a first implementation iteration. It's just kind of implied here :)

automatic-app-upgrades/README.md
kubasobon: Done :)

whites11:
For my understanding, would this be 100% automated? In other words, when `ImageUpdateAutomation` updates the app CR in the gitops repo, will this be a PR to be reviewed by humans or will this be 100% automatic?

kubasobon:
It's either/or. Every `ImageUpdateAutomation` has a field (`.spec.git.push.branch`) which specifies a branch to commit to. 
If we set it to commit to `main` (or any other branch plugged directly into GitOps), it will happen 100% automatically. Otherwise, a merge/PR of the updated branch into `main` will be necessary to roll out the change. 

My suggestion would be to have it fully automatic and make an exception where it makes sense. I'd like to avoid having hundreds of automated PRs.

puja108 (approved):

MarcelMue (approved):
Thanks for the added context. I like it a lot. But I am also very biased :D

kubasobon:
Time to merge :D

pipo02mix:
I arrived late to the party, but I have a comment. Since some upgrades could fail (missing properties, unexpected breaking changes, helm upgrade failures,...) do we have a plan for rolling back when it does not succeed?

PR #35, opened 2022-05-02
Propose a label scheme to indicate the priority/importance of clusters

marians:
Towards https://github.com/giantswarm/giantswarm/issues/13342

Closes https://github.com/giantswarm/giantswarm/issues/21886

### DIFF
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -52,7 +52,7 @@ If the label is used with a value different from the well-defined ones, the labe
 
 ## Additional considerations and explanations
 
-- Understand this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.
+- Understanding of this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.
 
 - Individual customers may or may not use the system. It is not required that all customers adopt it for it to become usable.
 
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -30,7 +30,7 @@ Label name: `giantswarm.io/service-priority`
 
 The label is to be used on the main resource defining the cluster, usually of resource type `clusters.cluster.x-k8s.io`.
 
-To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be applied on other resources of the cluster. Systems MUST ignore the label used on other resources in the same cluster. This means that it is not possible, for example, to differentiate between the importance of sevewral node pools within the same cluster.
+To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be applied on other resources of the cluster. Systems MUST ignore the label used on other resources in the same cluster. This means that it is not possible, for example, to differentiate between the importance of several node pools within the same cluster.
 
 ### Well-defined label values
 
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -17,7 +17,7 @@ Contents:
 
 - One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging (`s`). This customer also has clusters which don't fall into either of these categories and their names start with arbitrary letters.
 - Many customers use specific terms and phrases in the cluster description, e. g. "Production", "test", "testing", "ignore alerts", "prd", "prod", "nonprod".
-- Some customers use labels on the `clusters.cluster.x-k8s.io` resource (the cluster's main resource) indicating the importance. E. g.
+- Some customers use labels on the `clusters.cluster.x-k8s.io` resource (the cluster's main resource) indicating the importance, e.g:
     - `odp/environment: development`
     - `odp/environment: staging`
     - `odp/environment: production`.
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -58,7 +58,7 @@ If the label is used with a value different from the well-defined ones, the labe
 
 - Customers that already have a classification scheme implemented should be encouraged to also implement the system proposed here. Several systems can co-exist side by side, however in that case Giant Swarm staff will be instructed to consider the system described here as the authoritative one.
 
-- The label, by default, has no effect on monitoring, alerting, support, usage cost etc. of the cluster.
+- The label, by default, has no effect on monitoring, alerting, support, usage cost, routing priority, workload quality of service (QoS) etc. of the cluster. However it MAY serve as an input for such automation in the future.
 
 - The `kubectl gs template cluster` command SHOULD be extended to include the proposed label, with the default value being `production` and the other well-defined labels to be set via flags. This will help give the system more visibility and increase adoption.
 
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -15,7 +15,7 @@ Contents:
 
 ## Current practice
 
-- One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging `s`. This customer also has clusters which don't fall into either of these categories and their names start wtith arbotrary letters.
+- One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging (`s`). This customer also has clusters which don't fall into either of these categories and their names start with arbitrary letters.
 - Many customers use specific terms and phrases in the cluster description, e. g. "Production", "test", "testing", "ignore alerts", "prd", "prod", "nonprod".
 - Some customers use labels on the `clusters.cluster.x-k8s.io` resource (the cluster's main resource) indicating the importance. E. g.
     - `odp/environment: development`
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -48,7 +48,7 @@ The absence of the label SHOULD be interpreted as if the label was present with 
 
 ### Using undefined values
 
-If the label is used with a value different from the well-defined ones, the label SHOULD be interpreted as if the value was `development`.
+Undefined values will not be allowed to avoid confusion and for potential future use cases to integrate with tooling, like incident management. Customers can still add any other labels they need.
 
 ## Additional considerations and explanations
 
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -36,15 +36,15 @@ To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be ap
 
 There are three well-defined label values to be used.
 
-- `giantswarm.io/service-priority: production`: A cluster labelled this way serves production traffic (i. e. end-user facing). Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label.
+- `giantswarm.io/service-priority: critical`: This is the highest priority class. Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label. We expect customers to use this class for clusters serving user-facing applications, production traffic, etc.
 
-- `giantswarm.io/service-priority: staging`: Clusters carrying this label are used to simulate running on a production cluster, but are not serving production traffic. They are considered less important than ones carrying the above `production` label.
+- `giantswarm.io/service-priority: important`: Clusters carrying this label are considered less important than `critical`, but still more important than clusters classified as `unimportant`. Typical use cases would be staging clusters, clusters handling batch workloads, or development clusters that cannot be replaced easily by creating a new cluster.
 
-- `giantswarm.io/service-priority: development`: A cluster with this label is used by developers to develop and test applications. It is considered less important than a cluster labelled as `production` or `staging`.
+- `giantswarm.io/service-priority: unimportant`: The lowest priority class. Customers are expected to use this class for clusters that isn't relied on, or which can be replaced easily and quickly.
 
 ### Absence of the label
 
-The absence of the label SHOULD be interpreted as if the label was present with the value `production`.
+The absence of the label SHOULD be interpreted as if the label was present with the value `critical`.
 
 ### Using undefined values
 
@@ -60,7 +60,7 @@ If the label is used with a value different from the well-defined ones, the labe
 
 - The label, by default, has no effect on monitoring, alerting, support, usage cost, routing priority, workload quality of service (QoS) etc. of the cluster. However it MAY serve as an input for such automation in the future.
 
-- The `kubectl gs template cluster` command SHOULD be extended to include the proposed label, with the default value being `production` and the other well-defined labels to be set via flags. This will help give the system more visibility and increase adoption.
+- The `kubectl gs template cluster` command SHOULD be extended to include the proposed label, with the default value being `critical` and the other well-defined labels to be set via flags. This will help give the system more visibility and increase adoption.
 
 - While it is possible to set labels on clusters via the Management API or in cluster manifests, it is also recommendable to extend the web UI to set the proposed label and choose a well-defined value on cluster creation.
 
@@ -72,13 +72,13 @@ To help interpreting the proposal and judge its significance, here we suggest so
 
 - In the web UI there are several places where the proposed label can be displayed as useful information for users:
 
-    - In the cluster list, as an indicator especially for clusters labelled as `production` and `staging`.
+    - In the cluster list, as an indicator especially for clusters classified as `critical` and `important`.
 
     - On the cluster details page.
 
-    - When deleting a cluster, especially in case the cluster is labelled as `production` and `staging`, to highlight the purpose of the cluster.
+    - When deleting a cluster, especially in case the cluster is classified as `critical` and `important`, to highlight the purpose of the cluster.
 
-- In kubectl-gs, when listing clusters using the `get clusters` command, an additional column `SERVICE PRIORITY` can be added to display the value based on the label (or the default value `production`, in case of absence).
+- In kubectl-gs, when listing clusters using the `get clusters` command, an additional column `SERVICE PRIORITY` can be added to display the value based on the label (or the default value `critical`, in case of absence).
 
 - We could add the label to our monitoring, to include the priority information as a criterium in visualization or alerting.
 
--- a/classify-cluster-priority/README.md
+++ b/classify-cluster-priority/README.md
@@ -36,15 +36,15 @@ To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be ap
 
 There are three well-defined label values to be used.
 
-- `giantswarm.io/service-priority: critical`: This is the highest priority class. Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label. We expect customers to use this class for clusters serving user-facing applications, production traffic, etc.
+- `giantswarm.io/service-priority: highest`: This is the highest priority class. Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label. We expect customers to use this class for clusters serving user-facing applications, production traffic, etc.
 
-- `giantswarm.io/service-priority: important`: Clusters carrying this label are considered less important than `critical`, but still more important than clusters classified as `unimportant`. Typical use cases would be staging clusters, clusters handling batch workloads, or development clusters that cannot be replaced easily by creating a new cluster.
+- `giantswarm.io/service-priority: medium`: Clusters carrying this label are considered less important than `highest`, but still more important than clusters classified as `lowest`. Typical use cases would be staging clusters, clusters handling batch workloads, or development clusters that cannot be replaced easily by creating a new cluster.
 
-- `giantswarm.io/service-priority: unimportant`: The lowest priority class. Customers are expected to use this class for clusters that isn't relied on, or which can be replaced easily and quickly.
+- `giantswarm.io/service-priority: lowest`: The lowest priority class. Customers are expected to use this class for clusters that isn't relied on, or which can be replaced easily and quickly.
 
 ### Absence of the label
 
-The absence of the label SHOULD be interpreted as if the label was present with the value `critical`.
+The absence of the label SHOULD be interpreted as if the label was present with the value `highest`.
 
 ### Using undefined values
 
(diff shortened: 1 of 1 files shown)


classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+1. User stories
+2. Current practice
+3. Proposal
+4. Considerations and explanations
+5. Possible applications
+
+## User stories
+
+- As a cluster admin, I want to easily communicate the purpose and importance of a cluster to those interactig with it, so that they can act accordingly.
+- As Giant Swarm operations staff I want to understand the importance of clusters so I can make appropriate decisions, e. g. when prioritizing problem mitigation in an incident affecting multiple clusters.
```
anvddriesch, line 14: We might want to mention escalation as an important user story as well. We could derive from the label whether getting the customer or other engineers involved outside working hours is needed.

anvddriesch (commented):
From my side it would be okay to request review from outside the team. The proposal outlined here makes sense to me.

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+The absence of the label SHOULD be interpreted as if the label was present with the value `production`.
+
+### Using undefined values
+
+If the label is used with a value different from the well-defined ones, the label SHOULD be interpreted as if the value was `development`.
+
+## Additional considerations and explanations
+
+- Understand this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.
+
```
kuosandys, line 56:
```suggestion
- Understanding of this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.

```
Small suggestion here to improve clarity

kuosandys:
Looks good to me as well 👍

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Proposal
+
+We introduce a label and a schema for the label values to indicate the relative importance of a cluster.
+
+Label name: `giantswarm.io/service-priority`
+
+The label is to be used on the main resource defining the cluster, usually of resource type `clusters.cluster.x-k8s.io`.
+
+To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be applied on other resources of the cluster. Systems MUST ignore the label used on other resources in the same cluster. This means that it is not possible, for example, to differentiate between the importance of sevewral node pools within the same cluster.
```
AverageMarcus, line 33:
```suggestion
To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be applied on other resources of the cluster. Systems MUST ignore the label used on other resources in the same cluster. This means that it is not possible, for example, to differentiate between the importance of several node pools within the same cluster.
```

What's the reason for this?

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Well-defined label values
+
+There are three well-defined label values to be used.
+
+- `giantswarm.io/service-priority: production`: A cluster labelled this way serves production traffic (i. e. end-user facing). Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label.
+
+- `giantswarm.io/service-priority: staging`: Clusters carrying this label are used to simulate running on a production cluster, but are not serving production traffic. They are considered less important than ones carrying the above `production` label.
+
+- `giantswarm.io/service-priority: development`: A cluster with this label is used by developers to develop and test applications. It is considered less important than a cluster labelled as `production` or `staging`.
```
AverageMarcus, line 43: I'm not sure these values match the label. They're more environment than priority. Add to this the fast that "staging" has a few dozen different names in different orgs.

I was thinking something like `Critical` (e.g. production sales), `Important` (e.g. CI or job clusters, back office services with failover capabilities), `Low` (e.g. clusters that can afford downtime, especially out of working hours)

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- The label, by default, has no effect on monitoring, alerting, support, usage cost, routing priority, workload quality of service (QoS) etc. of the cluster. However it MAY serve as an input for such automation in the future.
+
+- The `kubectl gs template cluster` command SHOULD be extended to include the proposed label, with the default value being `production` and the other well-defined labels to be set via flags. This will help give the system more visibility and increase adoption.
+
+- While it is possible to set labels on clusters via the Management API or in a cluster manifests, it is also recommendable to extend the web UI to set the proposed label and choose a well-defined value on cluster creation.
+
+- While it is already possible to set, modify, and delete cluster labels via the web UI, it is also recommended to make it simpler for end users to specify the proposed label with a well-defined value for a given cluster.
+
```
AverageMarcus, line 68: I'd also like to see `kubectl gs get clustets` output expanded to highlight this label if present.

AverageMarcus: Ah! You have it below! :D Nice!

classify-cluster-priority/README.md
gianfranco-l: I agree on Marcus proposal here

gianfranco-l:
@marians it would be good for the whole @giantswarm/team-honeybadger to have a look at this. At the moment, half of the team is afk (will be back in 2 days) + a severe priority we are taking care of. Can this PR be kept open until next monday?

classify-cluster-priority/README.md
brinker211: Agreed. This also would help with the use-case of clusters being prepared. Example: Customer is building up a cluster for production but traffic is not there yet meaning its a production cluster with low priority but could switch label for Critical when actually serving traffic.

AverageMarcus: ☝️  That's a very good point. I hadn't thought about that use case. Might be worth explicitly calling it out in the user stories.

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+1. User stories
+2. Current practice
+3. Proposal
+4. Considerations and explanations
+5. Possible applications
+
+## User stories
+
+- As a cluster admin, I want to easily communicate the purpose and importance of a cluster to those interactig with it, so that they can act accordingly.
```
Oshratn, line 13:
```suggestion
- As a cluster admin, I want to easily communicate the purpose and importance of a cluster to those interacting with it, so that they can act accordingly.
```

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+- As Giant Swarm operations staff I want to understand the importance of clusters so I can make appropriate decisions, e. g. when prioritizing problem mitigation in an incident affecting multiple clusters.
+
+## Current practice
+
+- One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging `s`. This customer also has clusters which don't fall into either of these categories and their names start wtith arbotrary letters.
```
Oshratn, line 18:
```suggestion
- One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging (`s`). This customer also has clusters which don't fall into either of these categories and their names start with arbitrary letters.
```

classify-cluster-priority/README.md
Oshratn: I agree with the above. 
Another use case supporting the above - a customer may decide their dev cluster has higher importance, since they don't want their developers to be waiting around doing nothing.

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+- `giantswarm.io/service-priority: production`: A cluster labelled this way serves production traffic (i. e. end-user facing). Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label.
+
+- `giantswarm.io/service-priority: staging`: Clusters carrying this label are used to simulate running on a production cluster, but are not serving production traffic. They are considered less important than ones carrying the above `production` label.
+
+- `giantswarm.io/service-priority: development`: A cluster with this label is used by developers to develop and test applications. It is considered less important than a cluster labelled as `production` or `staging`.
+
+### Absence of the label
+
+The absence of the label SHOULD be interpreted as if the label was present with the value `production`.
+
```
Oshratn, line 48: Considering the proposal to use different values than the original proposal. Should a non-labeled cluster default to 'critical' or 'low'?

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+### Using undefined values
+
+If the label is used with a value different from the well-defined ones, the label SHOULD be interpreted as if the value was `development`.
+
+## Additional considerations and explanations
+
+- Understanding of this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.
+
+- Individual customers may or may not use the system. It is not required that all customers adopt it for it to become usable.
```
Oshratn, line 57: If customers don't use this, do the clusters get the default labels?
If so, I would think labels should default to low, so as not to overwhelm the people doing operations and to encourage customers to use the label scheme.

classify-cluster-priority/README.md
AverageMarcus: I'd say `critical`. If we have no info on a clusters importance we need to assume it's mission critical until told otherwise. (IMO)

classify-cluster-priority/README.md
AverageMarcus: We currently treat every cluster equally important (unless we actually have some inside knowledge of what it's used for) so defaulting to critical wouldn't change anything.

Oshratn: Makes sense, thanks for the clarification.

classify-cluster-priority/README.md
anvddriesch: I agree with this. For now we treat every cluster as production unless a clear indicator was added on the customer side. The idea is not to change this approach but to offer this as a feature and therefore make it uniform across customers and usable for us.

AverageMarcus: Yeah, and if you take the comment above about a customer potentially wanting their dev to be critical it makes sense to assume critical unless otherwise stated.

classify-cluster-priority/README.md
marians: Nice input, thank you!

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- `giantswarm.io/service-priority: development`: A cluster with this label is used by developers to develop and test applications. It is considered less important than a cluster labelled as `production` or `staging`.
+
+### Absence of the label
+
+The absence of the label SHOULD be interpreted as if the label was present with the value `production`.
+
+### Using undefined values
+
+If the label is used with a value different from the well-defined ones, the label SHOULD be interpreted as if the value was `development`.
```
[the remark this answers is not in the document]
marians: Yes, true. Do you think that this must be part of the RFC? I would like to keep the RFC lean in terms of the required efforts and I would think it's not terrible if non-defined values exist, as long as we know how to interpret those.

classify-cluster-priority/README.md
marians: Right. Will resolve the comment.

classify-cluster-priority/README.md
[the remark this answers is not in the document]
AverageMarcus: No, just wanted to point out that there's options :)

marians:
> @marians it would be good for the whole @giantswarm/team-honeybadger to have a look at this. At the moment, half of the team is afk (will be back in 2 days) + a severe priority we are taking care of. Can this PR be kept open until next monday?

Sure!

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Individual customers may or may not use the system. It is not required that all customers adopt it for it to become usable.
+
+- Customers that already have a classification scheme implemented should be encouraged to also implement the system proposed here. Several systems can co-exist side by side, however in that case Giant Swarm staff will be instructed to consider the system described here as the authoritative one.
+
+- The label, by default, has no effect on monitoring, alerting, support, usage cost, routing priority, workload quality of service (QoS) etc. of the cluster. However it MAY serve as an input for such automation in the future.
+
+- The `kubectl gs template cluster` command SHOULD be extended to include the proposed label, with the default value being `production` and the other well-defined labels to be set via flags. This will help give the system more visibility and increase adoption.
+
+- While it is possible to set labels on clusters via the Management API or in a cluster manifests, it is also recommendable to extend the web UI to set the proposed label and choose a well-defined value on cluster creation.
```
JosephSalisbury, line 65:
```suggestion
- While it is possible to set labels on clusters via the Management API or in cluster manifests, it is also recommendable to extend the web UI to set the proposed label and choose a well-defined value on cluster creation.
```

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+## User stories
+
+- As a cluster admin, I want to easily communicate the purpose and importance of a cluster to those interacting with it, so that they can act accordingly.
+- As Giant Swarm operations staff I want to understand the importance of clusters so I can make appropriate decisions, e. g. when prioritizing problem mitigation in an incident affecting multiple clusters.
+
+## Current practice
+
+- One customer uses the first character of the cluster name to distinguish between production (`p`), development (`d`), and staging (`s`). This customer also has clusters which don't fall into either of these categories and their names start with arbitrary letters.
+- Many customers use specific terms and phrases in the cluster description, e. g. "Production", "test", "testing", "ignore alerts", "prd", "prod", "nonprod".
+- Some customers use labels on the `clusters.cluster.x-k8s.io` resource (the cluster's main resource) indicating the importance. E. g.
```
JosephSalisbury, line 20:
```suggestion
- Some customers use labels on the `clusters.cluster.x-k8s.io` resource (the cluster's main resource) indicating the importance, e.g:
```

classify-cluster-priority/README.md
marians: Coming in with the next commit. I'm resolving this thread.

classify-cluster-priority/README.md
piontec: Very much on point. From ours (honeybadger) point of view we even have some use cases, where we would like to route certain alerts on dev clusters directly to customers, without involving the team

classify-cluster-priority/README.md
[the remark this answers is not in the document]
piontec: Maybe let's be strict here? Customers can add any other labels they want, indicating internal priority and so on. But for us, it will be probably the easiest to have a limited set of options.

piontec (approved):
Very good scope and definitions, IMO. I love RFC-type obligations defintions

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Well-defined label values
+
+There are three well-defined label values to be used.
+
+- `giantswarm.io/service-priority: critical`: This is the highest priority class. Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label. We expect customers to use this class for clusters serving user-facing applications, production traffic, etc.
+
+- `giantswarm.io/service-priority: important`: Clusters carrying this label are considered less important than `critical`, but still more important than clusters classified as `unimportant`. Typical use cases would be staging clusters, clusters handling batch workloads, or development clusters that cannot be replaced easily by creating a new cluster.
+
+- `giantswarm.io/service-priority: unimportant`: The lowest priority class. Customers are expected to use this class for clusters that isn't relied on, or which can be replaced easily and quickly.
```
uvegla, line 43: Instead of `unimportant` I would name this level `standard`.

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+We introduce a label and a schema for the label values to indicate the relative importance of a cluster.
+
+Label name: `giantswarm.io/service-priority`
+
+The label is to be used on the main resource defining the cluster, usually of resource type `clusters.cluster.x-k8s.io`.
+
+To simplify maintenance and avoid logical conflicts, this label SHOULD NOT be applied on other resources of the cluster. Systems MUST ignore the label used on other resources in the same cluster. This means that it is not possible, for example, to differentiate between the importance of several node pools within the same cluster.
+
+### Well-defined label values
```
uvegla, line 35: This whole concept slightly me of how log levels are usually implemented. Instead of fix values we could define a scale of let's say 0-1000 for simplicity (usually this is much wider I think). Then we can name certain point for example: { 250: Standard, 500: Important, 750: Critical }. This still gives room to make a difference between critical and critical and even makes the system easily extendable in the future if the spaces are big enough.

Labels could still be by both numeric value and human readable value, CLI tools could rely on the standard named levels and assign the proper labels and values.

The label values and the they are in sync can be enforced via admission and validation webhooks I think.

classify-cluster-priority/README.md
uvegla: I think it has a better feeling to it and contains no negative message to users for example.

marians: Well, except for `standard` sounds a lot like this would be the default, which it isn't. `critical` would be default.

Can anyone think of a better term?

classify-cluster-priority/README.md
marians: If you are suggesting a numeric value system, I am not a fan of it for the reason that numbers, especially when it comes to ranking priority, are not self-explanatory. As a user, many will wonder: is 1 a high priority or a low one? (Many people in day to day communication say things like "this is priority one for me". But it's also a low number compared to 2 and 3. So people will always have to look up some reference.

I think natural strings like we have here are easier to get, while of course there is not guarantee that they work for every user.

AverageMarcus: Yeah I’m not keen on under scale either. Customers won’t know what value to set and therefore just leave it as default.

classify-cluster-priority/README.md
AverageMarcus: Yeah, our standard is to treat everything as critical. Maybe something like “inessential”?

puja108: why not go with the typical priority scale of, high, medium, low (critical could be more than just high)? Unimportant or inessential have connotations that a customer could interpret as service level (e.g. no alerting), but alerting and such is handled separately, so priority is solely on a comparison between two or more clusters.

classify-cluster-priority/README.md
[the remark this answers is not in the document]
puja108:
```suggestion
Undefined values will not be allowed to avoid confusion and for potential future use cases to integrate with tooling, like incident management. Customers can still add any other labels they need.
```

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+    - In the cluster list, as an indicator especially for clusters classified as `critical` and `important`.
+
+    - On the cluster details page.
+
+    - When deleting a cluster, especially in case the cluster is classified as `critical` and `important`, to highlight the purpose of the cluster.
+
+- In kubectl-gs, when listing clusters using the `get clusters` command, an additional column `SERVICE PRIORITY` can be added to display the value based on the label (or the default value `critical`, in case of absence).
+
+- We could add the label to our monitoring, to include the priority information as a criterium in visualization or alerting.
```
puja108, line 83: having it in alerting/paging would be great

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- `giantswarm.io/service-priority: critical`: This is the highest priority class. Clusters with this label value are considered the most important ones, relative to other clusters not carrying this label. We expect customers to use this class for clusters serving user-facing applications, production traffic, etc.
+
+- `giantswarm.io/service-priority: important`: Clusters carrying this label are considered less important than `critical`, but still more important than clusters classified as `unimportant`. Typical use cases would be staging clusters, clusters handling batch workloads, or development clusters that cannot be replaced easily by creating a new cluster.
+
+- `giantswarm.io/service-priority: unimportant`: The lowest priority class. Customers are expected to use this class for clusters that isn't relied on, or which can be replaced easily and quickly.
+
+### Absence of the label
+
+The absence of the label SHOULD be interpreted as if the label was present with the value `critical`.
```
pipo02mix, line 47: I would add a kyverno rule that adds a default value. IMO for new clusters, I will default it to `unimportant` as it is usually the case till customer add important services. For existing clusters we can set `critical`

classify-cluster-priority/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Using undefined values
+
+If the label is used with a value different from the well-defined ones, the label SHOULD be interpreted as if the value was `development`.
+
+## Additional considerations and explanations
+
+- Understanding of this system by Giant Swarm staff is crucial. We have to make sure that everybody in operations know about the system.
```
pipo02mix, line 55: we could propagate this to monitoring and act based on importance

classify-cluster-priority/README.md
marians: Setting a different default value for existing vs. new clusters is something I find too inconsistent and would rather avoid.

In the past we have always dealt with customer clusters in a way that we treated them as production, unless we knew they were not.

I would instead favour making it easy for cluster creators to set the priority, e. g. by having the option it in the cluster creation form in the web UI. The goal here would be to make more users consciously set the priority instead of leaving it at the default.

classify-cluster-priority/README.md
marians: Then here is my alternative suggestion:

- `highest`
- `medium`
- `lowest`

This should make it clear for each of the values, simply from looking at it, what it means relative to others. You don't even have to memorize the entire scale.

marians:
Thanks for the reviews, everyone! I think your feedback helped to improve the proposal significantly.

If you could be so kind to come back here for either approval or more wishes for improvement, that would be great. Thanks!

kuosandys (approved):

gusevda (approved):

PR #37, opened 2022-05-30
ClusterAPI custom resource monitoring

bavarianbidi:
Signed-off-by: Mario Constanti <mario@constanti.de>

Status:
- 2023-05-25: PR will be closed (without merging it) soon. The current state is documented in PR https://github.com/giantswarm/giantswarm/pull/27092 and all future discussions will be moved into an own issue


fiunchinho:
Thanks for the RFC.

[The CASM upstream proposal "future work" section](https://github.com/chrischdi/cluster-api/blob/2a9d16a0f1f7a415cc8cb8942d51c47d982397a6/docs/proposals/20220411-cluster-api-state-metrics.md#future-work), mentions adding provider metrics in the future, so I guess solving this means discussing with upstream about the best solution.

Talking about the proposed solutions here (and without having talked to upstream people yet), I feel like having different implementations for different providers living in different applications/repos is more aligned with how the CAPI project is currently made/designed, making different providers to be implemented as different applications. As it's mentioned in the RFC, trying to put all potential providers metrics implementations into a single app would bring many many code dependencies into the project (or a complex use of `unstructured` objects using conditionals to use different fields for different providers?). Apart from code dependencies, it would introduce provider dependency, meaning that development for one provider could affect other providers.

Using different implementations for different providers makes it harder to deploy, but that's a problem that we are already solving for CAPI controllers anyway.

cluster-api-state-metrics-adoption/README.md
```diff
@@ -0,0 +1,237 @@
...
+- Management Cluster (MC) is the Kubernetes Cluster where CAPI components are deployed and which is responsible for creating new Kubernetes Clusters (WCs)
+- Workload Cluster (WC)
+
+## Problem statement
+
+As Cluster Operator we want to get know the state of any Cluster API related object (no matter if Cluster API Core or infrastructure provider specific).
+
+As different states have impact on the expected behavior of the reconciliation e.g.
+
+- A cluster upgrade won't finish successful if an already existing OpenStackMachine is in an error state (even if the machine is in a transient error).
```
erkanerol, line 32: Note that we have MachineHealthCheck and it fixes transient errors.

cluster-api-state-metrics-adoption/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+The CASM code will be contributed to the Cluster API project (Process can be tracked in [issue #6458](https://github.com/kubernetes-sigs/cluster-api/issues/6458)).
+
+### Infrastructure provider specific metrics
+
+Currently CASM only provide metrics of Cluster API Core components (e.g. `clusters`, `kubeadmcontrolplanes`,`machinedeployments`,`machines` or `machinesets`) and it's currently not clear how infrastructure provider specific metrics can be added/generated once the entire code base is living under `kubernetes-sigs/cluster-api`.
+
+But as the code isn't that complex it's very easy to extend the existing implementation to support infrastructure provider specific metrics.
+
+Beside generated changes in `go.mod` and `go.sum` only the following changes are needed to add the two new metrics `capi_openstackcluster_created` and `capi_openstackcluster_paused` into CASM.
```
erkanerol, line 49: would it be so complex&dirty to add different providers by config? like passing GVK as config + using unstructured in the code.

cluster-api-state-metrics-adoption/README.md
bavarianbidi: > Note that we have MachineHealthCheck and it fixes transient errors.

unfortunately no, as the Node itself is up and running and healthy. Only the state/status of the machine might be in a failure state due a transient error with e.g. the Openstack API - API temp. not available.

cluster-api-state-metrics-adoption/README.md
bavarianbidi: > would it be so complex&dirty to add different providers by config? like passing GVK as config + using unstructured in the code.

my assumption was to align with the already existing implementation (import the type and reuse it) as much as possible.

If we're using `unstructured` objects, we have to deal with different fields for all used different providers, not sure if it's worth to doing it.

@fiunchinho already mentioned this above:
> As it's mentioned in the RFC, trying to put all potential providers metrics implementations into a single app would bring many many code dependencies into the project (or a complex use of `unstructured` objects using conditionals to use different fields for different providers?)

IMO:
As upstream implementation doesn't have a solution for provider specific metrics yet i guess it's worth to start with a simple solution for now (as the benefits for us might be huge on a short term). And once the discussion starts upstream we could take part there as we already have to refactor this component once upstream has found a good solution. WDYT @erkanerol ?

cluster-api-state-metrics-adoption/README.md
erkanerol: When `OpenStackMachine` has an error, isn't it propagated to the corresponding `Machine` CR?

bavarianbidi: no, doesn't propagated to the `Machine` (xref: https://github.com/giantswarm/giantswarm/issues/22077)

cluster-api-state-metrics-adoption/README.md
erkanerol: I think it is the simplest solution for us (OpenStack provider) but I am not sure it is the most optimal solution for the whole GiantSwarm. We can mention this in the next kaas sync. 

Nevertheless, I don't see any big risk in starting with this solution and then improving in time. So +1 from my side.

bavarianbidi: added a note to the upcoming KaaS-Sync

cluster-api-resource-monitoring/README.md
```diff
@@ -0,0 +1,327 @@
+# Cluster API custom resource monitoring
+
+<!-- toc -->
+- [Glossary](#glossary)
+- [Problem statement](#problem-statement)
+- [Cluster API state metrics](#cluster-api-state-metrics)
+  - [Infrastructure provider specific metrics](#infrastructure-provider-specific-metrics)
+    - [Variant 1: all providers in one generic CASM instance](#variant-1-all-providers-in-one-generic-casm-instance)
+    - [Variant 2: one CASM instance per infrastructure provider](#variant-2-one-casm-instance-per-infrastructure-provider)
+    - [Variant 2: one CASM instance per infrastructure provider plus one CASM instance for CAPI core](#variant-2-one-casm-instance-per-infrastructure-provider-plus-one-casm-instance-for-capi-core)
```
pipo02mix, line 10:
```suggestion
    - [Variant 3: one CASM instance per infrastructure provider plus one CASM instance for CAPI core](#variant-2-one-casm-instance-per-infrastructure-provider-plus-one-casm-instance-for-capi-core)
```
Would it be the third one?

cluster-api-resource-monitoring/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+- CAPI generic `CustomResourceStateMetrics` configuration could be easily re-used on different providers
+- KSM-App already exists
+
+#### Cons
+
+- Scope of KSM between management- and workload cluster differs
+  - resource requirements
+  - permissions
+- Changes in KSM-App affect all clusters, even if changes are only done for management clusters (e.g. version bump)
```
pipo02mix, line 281: can we just change those values at deployment time depending on the target cluster? I think we do for other components too

cluster-api-resource-monitoring/README.md
```diff
@@ -0,0 +1,327 @@
...
+##### Pros
+
+- Scope of this KSM instance is clearly focused on everything CAPI
+- Any further management cluster CR specific monitoring clearly goes there
+
+##### Cons
+
+- Cluster API versions must be reflect in the `CustomResourceStateMetrics` configuration. We will potentially diverge on the used Cluster API versions per infrastructure provider.
+
+#### Variant 2.2: dedicated `kube-state-metrics` instance on management cluster per CAPI providers
```
kopiczko, line 300: For this I understand we could put a bunch of `{{ if eq .Values.provider "openstack" }}` in the helm chart and that would do right?

cluster-api-resource-monitoring/README.md
erkanerol: My understanding is YES and this seems the best solution to me. We already use helm templates to differentiate behavior of apps (including monitoring stuff) according to the environment.

bavarianbidi: yes, it's possible. It's more about accepting the impact when we go this way

cluster-api-resource-monitoring/README.md
bavarianbidi: With Variant 1 and Variant 2.1 we will have a lot of `{{ if eq .Values.provider "openstack" }}` in the repos (and therefore we have to decide how to go with corresponding APP releases - as we could break other providers monitoring)

cluster-api-state-metrics-adoption/README.md
pipo02mix: what was the conclusion in that sync?

pipo02mix:
I am still confused if this will be implemented in CASM or KSM will fadeit out

cluster-api-state-metrics-adoption/README.md
bavarianbidi: > what was the conclusion in that sync?

that all others will take a look at this PR. Will propagate it again in the next couple of days.

bavarianbidi:
> I am still confused if this will be implemented in CASM or KSM will fadeit out

upstream discussion about KSM is still open (https://github.com/kubernetes/kube-state-metrics/issues/1755#issuecomment-1160113320 - and there is no feedback from KSM folks right now) but i still asume that we will go with KSM.

My idea was to get feedback about both possible implementations and once upstream has a decision we can immediately start with our implementation.

bavarianbidi:
for now i would start with a dedicated app - as i won't run the kube-system based KSM from a forked custom branch (https://github.com/kubernetes/kube-state-metrics/compare/master...chrischdi:poc-additional-metric-types-2?expand=1)

cluster-api-resource-monitoring/README.md
bavarianbidi: That's exactly one idea, KSM folks mentioned someday in slack (TODO marioc: link to that statement) that each third party component can ship there own `CustomResourceStateMetrics` configuration and `KSM` is taking care of it.

Currently `CustomResourceStateMetrics` is only defined as key in the configmap, which `KSM` will read.

bavarianbidi:
@JosephSalisbury as you take the CAPI-Monitoring thing into an upcoming "product-sync" - i'm currently think about closing this RFC (with a comment) and creating a new RFC to 
* adopt [variant 2.1](https://github.com/giantswarm/rfc/blob/3a6cfb248f5bca134d53a6d7f960092733d8d906/cluster-api-resource-monitoring/README.md#variant-21-dedicated-kube-state-metrics-instance-on-management-cluster-for-all-capi-providers) as a general purpose for future `CR`-Monitoring in ManagementClusters (and eventually in WCs as well - if needed)
* and if we (as GS) belive in a solution, where 
  * `KSM` in a long-term provides a `CustomResourceStateMetrics` CRD and watch/list/get all `CustomResourceStateMetrics` CRs 
  * and each controller implementation (e.g. `ClusterAPI`, `Flux`, ...) creates/serves an own MetricConfiguration of type `CustomResourceStateMetrics` 
  
  might be the better approach - we could think about implementing/providing this feature to KSM.
  
  During the initial implementation of `cluster-api-monitoring-app` and demonstrating the special KSM-instance i often got asked, if `CustomResourceStateMetrics` is a `CRD` in a cluster (currently it's only the "header" of the configuration and only used as anchor for generating the configuration)
  
  Until now, the Idea itself got only published once in a [slack thread](https://kubernetes.slack.com/archives/CJJ529RUY/p1654600684383799?thread_ts=1654593034.854759&cid=CJJ529RUY) by KSM maintainer Manuel Krug:

  > I'm also currently thinking about a kube-state-metrics-customresourcemetrics CRD, so each CRD could ship their own set of metric definitions with its distribution and kube-state-metric would pick it up

JosephSalisbury:
yeah, i reckon recreating / updating https://github.com/giantswarm/rfc/pull/37 to focus on variant 2.1 (i.e: present all the options, make 2.1 the one we're going to go with) would be neat, and focus it more on cr monitoring in geenral

i'm not sure yet about customeresourcestatemetrics crs. it feels a bit too far away for us to make a bet on. i could see us having some app that we deploy, and we have our config there, then if customeresourcestatemetrics crs become a thing, we break that apart - does that sound reasonable?

pipo02mix:
yeah it would be a good idea. WDYT @bavarianbidi ?

cluster-api-state-metrics-adoption/README.md
gawertm: whats left for this rfc now?

bavarianbidi: my time to convert it into a new RFC ... still on my list

bavarianbidi:
> will try to invest some time this week/next week to make the current status clear.

oh man, what was I thinking :see_no_evil: let's give them another try in the next couple of weeks :crossed_fingers:

bavarianbidi:
PR will closed soon.
current state is documented here: https://github.com/giantswarm/giantswarm/pull/27092

All the future discussions about the possibilities with the CustomResourceMonitoring feature of KSM will be moved into a dedicated issue as there is a lot of progress on upstream atm.

bavarianbidi:
Current CAPI monitoring is described in https://intranet.giantswarm.io/docs/monitoring/capi_monitoring/.
This RFC will be closed as we have a solution for now. Any additional improvements will be handled in https://github.com/giantswarm/roadmap/issues/2513

pipo02mix:
But it is not the first variant? I was thinking why not merging instead of closing and don't lose the work you did

bavarianbidi:
> But it is not the first variant? I was thinking why not merging instead of closing and don't lose the work you did

@pipo02mix 
I didn't want to merge the PR as we didn't had a real consent about the implementation and due high priorities and the operational need we started with variant 2.1 - so there was no "real discussion"

I've recently documented this setup in our intranet and also referenced to this PR for possible alternative implementations (https://intranet.giantswarm.io/docs/monitoring/capi_monitoring/#reason-for-current-architecture).
And also upstream KSM is extending the current feature which should have impact for the entire CR monitoring thing within GS (https://github.com/giantswarm/roadmap/issues/2513)

PR #39, opened 2022-06-15
added RFC about multi-layer configs

piontec:


multi-layer-app-config/README.md
```diff
@@ -0,0 +1,193 @@
...
+This seems unrealistic, as we really a lot on app platform and its features. Still, bypassing App CR in GitOps
+scenario would solve the problem.
+
+#### Pros
+
+- we don't implement anything in code
+
+#### Cons
+
+- this might mean phasing out app platform operators or leaving in two worlds (with and without app platform)
```
kubasobon, line 151: nit:
```suggestion
- this might mean phasing out app platform operators or living in two worlds (with and without app platform)
```

kubasobon (approved):
Looks good. I'm leaning towards _Enhancing App CR_ approach for now.

multi-layer-app-config/README.md
```diff
@@ -0,0 +1,193 @@
...
+  name: nginx-ingress-controller-app
+  namespace: kube-system
+  version: 2.7.0
+```
+
+#### Pros
+
+- solves the problem everywhere, in App CR and in GitOps
+- backward compatible
+- easy to implement
```
ljakimczuk, line 116: But is it easy to release? Doesn't it tie back to the problem of patching release CRs? I mean, if I am not mistaken, implementation should be farily easy for unique App Operator and CAPx (org-namespaced) App Operators, because there are either latest or close to it, but there are also older versions of App Operator in use that we would need to patch in order to support these, right? Of course if customers upgrade the clusters then that's not a problem anymore.

multi-layer-app-config/README.md
```diff
@@ -0,0 +1,193 @@
...
+This seems unrealistic, as we really a lot on app platform and its features. Still, bypassing App CR in GitOps
+scenario would solve the problem.
+
+#### Pros
+
+- we don't implement anything in code
+
+#### Cons
+
+- this might mean phasing out app platform operators or living in two worlds (with and without app platform)
```
puja108, line 151: not sure if phasing out is a con or maybe even a pro, if there's really not much more to current app platform use cases than what we can see already there in Flux, why should we keep it?

Sure, we have the metadata stuff in there, but there could be ways to have that on the side of or integrated in Flux.

I think similar to phasing out vintage for CAPI a decision to phase out our operators could be a strategic pro in the end.

puja108: cc @teemow

multi-layer-app-config/README.md
```diff
@@ -0,0 +1,193 @@
...
+- App CRs and app-platform are deeply integrated into our product, we can't get rid of it easily
+
+### Dropping the 'values' key in AppCR's CMs/Secrets and moving them to top level
```
uvegla, line 154: I am not very fond of this one. Leaning more towards App CR or getting rid of app platform if that makes sense.

multi-layer-app-config/README.md
MarcelMue: I think the benefit here is, that we only add a field. Adding a field in a CRD is a non-breaking change and could be rolled out easily. We acquire slightly more technical debt but the general roll-out should be no problem.

multi-layer-app-config/README.md
MarcelMue: We discussed this option multiple times in our team. I think pivoting makes sense but is not the right decision at this very moment. Just my 2c though

multi-layer-app-config/README.md
piontec: I agree. This new version of app-op can just create a new release that customers have to upgrade to if they want to get benefits. We don't have to patch existing releases.

multi-layer-app-config/README.md
piontec: I still think app platform has at least two unique values that still make it valid: real API representation of Helm charts and their deployments and extended meta data that we can use for better user experience. Also, dropping it would be a major effort right now, so I also think we need app platform right now.

piontec:
@giantswarm/team-honeybadger Do we have an agreement then and we want to add config lists like in Flux to the App CR? Please confirm and I'll move this RFC to an actual ADR.

PR #41, opened 2022-07-01
Add priorities to extra configs field in the multi layer app config RFC

uvegla:
The new `priorities` field makes it very flexible to apply new layers of configuration while still keeping backward compatibility. It supports our `gitops` use case and `late-binding` of configurations for `CAPI` clusters.

### DIFF
--- a/multi-layer-app-config/README.md
+++ b/multi-layer-app-config/README.md
@@ -67,8 +67,7 @@ The merging algorithm is as follows:
 5. Configuration from `userConfig` entry (C)
 6. All entries from `extraConfigs` with priority of P: C < P
 
-In case of multiple items in `extraConfigs` having the same priority, the order on the list is binding, with the item lower on the list being merged later (overriding those higher on the list)
-in the list will take the priority.
+In case of multiple items in `extraConfigs` having the same priority, the order on the list is binding, with the item lower on the list being merged later (overriding those higher on the list).
 
 The idea is modeled after Flux's [HelmRelease configuration](https://fluxcd.io/docs/components/helm/api/#helm.toolkit.fluxcd.io/v2beta1.ValuesReference).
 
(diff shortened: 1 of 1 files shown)


multi-layer-app-config/README.md
```diff
@@ -40,6 +40,36 @@ We can extend App CR with on optional list of `ConfigMap`/`Secret` objects. Conf
...
+possible to apply some of the `extraConfigs` between the `config` and `userConfig` entries or even after `userConfig`.
+The bedrock is still considered to be what is in the catalog. It is not possible to apply `extraConfigs` before that.
+
+#### Merging algorithm
+
+Assuming the following priorities for the platform layers:
+
+- Catalog: A (e.g.: 0)
+- Cluster (`config`): B (e.g.: 50)
+- User (`userConfig`): (e.g.: 100)
```
piontec, line 55:
```suggestion
- User (`userConfig`): C (e.g.: 100)
```

multi-layer-app-config/README.md
```diff
@@ -40,6 +40,36 @@ We can extend App CR with on optional list of `ConfigMap`/`Secret` objects. Conf
...
+The merging algorithm is as follows:
+
+1. Configuration from the catalog (A)
+2. All entries from `extraConfigs` with priority of P: A <= P <= B
+3. Configuration from `config` entry (B)
+4. All entries from `extraConfigs` with priority of P: B < P <= C
+5. Configuration from `userConfig` entry (C)
+6. All entries from `extraConfigs` with priority of P: C < P
+
+In case of multiple items in `extraConfigs` appear with the same priority then the item that appears first
```
piontec, line 70:
```suggestion
In case of multiple items in `extraConfigs` having the same priority, the order on the list is binding, with the item lower on the list being merged later (overriding those higher on the list)
```

piontec, line 70: I think it makes sense to go with merging from top to bottom, so the last one added (appended to the list) is potentially overriding the others.

piontec (approved):
LGTM, nice examples!

PR #42, opened 2022-07-01
State that we intend to keep config and userConfig fields

uvegla:


uvegla:
Will have some more small corrections to it soon.

kubasobon (approved):

PR #43, opened 2022-07-06
RFC for SIG meeting improvements

MarcelMue:
This RFC follows the meeting improvement initiative. Merging this RFC will signify agreement to the proposed changes and lead to actual implementation of the changes proposed here.

### DIFF
--- a/sig-meeting-improvement/README.md
+++ b/sig-meeting-improvement/README.md
@@ -1,0 +1,50 @@
+# SIG Meeting Improvement Initiative
+
+## What are the problems we are experiencing?
+- We have too many meetings which employees are expected to attend.
+- Meetings are often not relevant for all participants.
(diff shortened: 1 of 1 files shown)


Mircohelle (approved):
lgtm

sig-meeting-improvement/README.md
```diff
@@ -0,0 +1,50 @@
+# SIG Meeting Improvement Initiative
+
+## What are the problems we are experiencing?
+- We have too many meetings which employees are expected to attend.
```
pipo02mix, line 4: BTW I have always found challenging the fact of enforcing to have ambassadors from all the teams and having less meetings. SIGs is something that grow, but team size used to be fixed

MarcelMue: I agree in general. We need to see how we can find a good balance :)

PR #44, opened 2022-08-08
Add RFC to create a public part of the intranet

MarcelMue:
This is a quick write-up of my thoughts around opening up our intranet to the public. Feel free to comment as I refine the ideas presented here.

### DIFF
--- a/make-intranet-public/README.md
+++ b/make-intranet-public/README.md
@@ -1,0 +1,49 @@
+# Making parts of the intranet public
+
+## Status quo
+
+We currently have an internal `hugo` page which we use for internal documentation.
+This page is called `intranet` and contains all kinds of documentation around our operational work as well as security and development.
+The sourcecode for our `intranet` is in a private github repository which contains other security and business relevant data.
+
+## Goal
+
+The non-security relevant parts of the intranet should be public.
+Making them public enables us to collaborate with other companies more easily.
+It additionally allows us to enable outside contributors to work on projects more independently.
+
+## Considerations
+
+Content in the `intranet` is not fit for our general customer base.
+It does not fit into our general product documentation but some of the content may be interesting to them.
+I will not specify which parts of the current intranet may be better suited for our product documentation,
+that is a decision which can be made independently of this proposal.
+
+## Proposed solution
+
+We create a new public repository called `handbook`.
+This repository uses the same general structure and hugo setup as our intranet.
+
+We then import the content from the `handbook` repository into the `intranet` source - this should be easily possible as described [here](https://discourse.gohugo.io/t/building-content-from-multiple-repositories/34636).
+Importing the `handbook` repository will mean that the search functionality in the `intranet` will automatically scrape both private and public docs.
+Therefore the `intranet` can remain as the single source of truth for Giant Swarm employees to aggregate information.
+
+The `handbook` repository itself should also be setup with `hugo` to host its content under `handbook.giantswarm.io` making it easily accessible for outside collaborators.
+
+The complete setup would then look as follows:
+- `handbook` only contains information we are comfortable with having public.
+- `intranet` only contains truly private information but imports all content from `handbook` into a unified search.
+- `docs` contains product documentation which is directly linked on our website as before.
+
+## Technical implementation
+
+There are some open technical questions with this proposal.
+1. Will it be possible to easily edit files from the hugo UI (e.g. edit button) even though the actual source of truth is split into 2 repositories now?
+2. Will it be possible to structure the `handbook` repository in a way that it is a standalone hugo page but also easy to import into the `intranet`?
+
+## Precedent in other companies
+
+Gitlab follows a similar pattern as the one proposed here:
+1. https://about.gitlab.com/handbook/ is the public documentation of gitlab itself (hiring, development, meeting guidelines, ...)
+2. The private part is https://internal-handbook.gitlab.io/ but is insaccessible to the public.
+3. https://docs.gitlab.com/ is the product documentation which is purely user focussed.
--- a/make-intranet-public/README.md
+++ b/make-intranet-public/README.md
@@ -4,7 +4,7 @@
 
 We currently have an internal `hugo` page which we use for internal documentation.
 This page is called `intranet` and contains all kinds of documentation around our operational work as well as security and development.
-The sourcecode for our `intranet` is in a private github repository which contains other security and business relevant data.
+The source code for our `intranet` is in a private GitHub repository which contains other security and business relevant data.
 
 ## Goal
 
--- a/make-intranet-public/README.md
+++ b/make-intranet-public/README.md
@@ -12,6 +12,9 @@ The non-security relevant parts of the intranet should be public.
 Making them public enables us to collaborate with other companies more easily.
 It additionally allows us to enable outside contributors to work on projects more independently.
 
+A very direct (and most urgent) goal is to enable outside collaborators quickly who need to work on docs for us.
+This is mainly related to companies who will help us implement features in our managed apps but will not get full access to our private docs.
+
 ## Considerations
 
 Content in the `intranet` is not fit for our general customer base.
--- a/make-intranet-public/README.md
+++ b/make-intranet-public/README.md
@@ -35,6 +35,16 @@ The complete setup would then look as follows:
 - `intranet` only contains truly private information but imports all content from `handbook` into a unified search.
 - `docs` contains product documentation which is directly linked on our website as before.
 
+### Transitioning content into the public handbook
+
+Content will be transitioned into the public handbook on a "if-needed" basis for now.
+We have some usecases which require our internal docs to be public (e.g. relevant docs to external collaborators, people related docs which help with hiring, ... ).
+
+Once the new concept is proven with the currently needed docs being public, then a `public by default` policy should be put in place.
+This should then by followed by all non-critical docs being moved to the public repo in quick succession.
+
+Docs should always be made public by the owners of the docs.
+
 ## Technical implementation
 
 There are some open technical questions with this proposal.
--- a/make-intranet-public/README.md
+++ b/make-intranet-public/README.md
@@ -27,9 +27,10 @@ that is a decision which can be made independently of this proposal.
 We create a new public repository called `handbook`.
 This repository uses the same general structure and hugo setup as our intranet.
 
-We then import the content from the `handbook` repository into the `intranet` source - this should be easily possible as described [here](https://discourse.gohugo.io/t/building-content-from-multiple-repositories/34636).
-Importing the `handbook` repository will mean that the search functionality in the `intranet` will automatically scrape both private and public docs.
+A similar search setup to our docs should be setup which covers the `handbook` as well as the `intranet`.
+That search will then be accessible through the `intranet`, linking to both sources of truth.
 Therefore the `intranet` can remain as the single source of truth for Giant Swarm employees to aggregate information.
+Most importantly, it becomes trivial to link URLs from either source as search results will forward you to the specific article.
 
 The `handbook` repository itself should also be setup with `hugo` to host its content under `handbook.giantswarm.io` making it easily accessible for outside collaborators.
 


Oshratn:
So this does not include making things easier for non-techies?
I also believe that we need to adopt the mindset of defaulting to public (I'm parroting @teemow , I know).
This means that after the initial migration, I recommend that in the docs PR template, we should add 2 questions to consider:
1. Is this document public?
2. If this document is not public, why not?

MarcelMue:
> So this does not include making things easier for non-techies? 

Yes, no improvements for non-techies would be made in this. It's something we would have to tackle separately.


> I also believe that we need to adopt the mindset of defaulting to public (I'm parroting @teemow , I know). This means that after the initial migration, I recommend that in the docs PR template, we should add 2 questions to consider:
> 
>     1. Is this document public?
> 
>     2. If this document is not public, why not?

We can add this as a PR template in `giantswarm/giantswarm` when we do this change.

AverageMarcus:
> Yes, no improvements for non-techies would be made in this. It's something we would have to tackle separately.

I think it would be much easier to tackle now, rather than trying to retroactively improve things. If not done now I doubt it ever will be done. E.g. Hugo isn't the most non-tech friendly (not even sure I'd call it tech-friendly) and moving to something else at a later point in time would be too much effort I think.

MarcelMue:
> > Yes, no improvements for non-techies would be made in this. It's something we would have to tackle separately.
> 
> I think it would be much easier to tackle now, rather than trying to retroactively improve things. If not done now I doubt it ever will be done. E.g. Hugo isn't the most non-tech friendly (not even sure I'd call it tech-friendly) and moving to something else at a later point in time would be too much effort I think.

I understand the concern but I unfortunately have a blocker now to have docs public - or at least a part of them - for collaboration with outside collaborators. As far as I know, we do not have any proposal on how to improve usability for non-techies.

Basically docs need to be public to an extend soon and more usable but I have no idea how to achieve both requirements.

AverageMarcus:
Yeah, totally get that. Not saying we should block this, just that we should be realistic and not assume that easier outside contributions will be something to easily tackle later. If that is a priority it needs to be considered now, if it's not (as you say, we've no proposal now so likely not) then we should continue but with correct expectations in place 😃

MarcelMue:
I have left my thoughts on usability in the (fittingly private) issue here:

https://github.com/giantswarm/giantswarm/issues/16422#issuecomment-1209136264

I would be happy if we could focus here purely on the public / private content topic and discuss usability of the intranet in https://github.com/giantswarm/giantswarm/issues/16422

make-intranet-public/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+We then import the content from the `handbook` repository into the `intranet` source - this should be easily possible as described [here](https://discourse.gohugo.io/t/building-content-from-multiple-repositories/34636).
+Importing the `handbook` repository will mean that the search functionality in the `intranet` will automatically scrape both private and public docs.
+Therefore the `intranet` can remain as the single source of truth for Giant Swarm employees to aggregate information.
+
+The `handbook` repository itself should also be setup with `hugo` to host its content under `handbook.giantswarm.io` making it easily accessible for outside collaborators.
+
+The complete setup would then look as follows:
+- `handbook` only contains information we are comfortable with having public.
+- `intranet` only contains truly private information but imports all content from `handbook` into a unified search.
```
puja108, line 35: I'd make it even stronger:
1. default to handbook, everything that is not product docs goes in there
2. only things that are truly private (to be defined) go to intranet

make-intranet-public/README.md
```diff
@@ -0,0 +1,49 @@
...
+- `docs` contains product documentation which is directly linked on our website as before.
+
```
puja108, line 37: I was kinda missing a step stating we move content from intranet to handbook

make-intranet-public/README.md
MarcelMue: I don't follow - we need to import into the intranet because otherwise we don't have one search over both repos. I want to prevent a situation where we have a split brain documentation.

make-intranet-public/README.md
MarcelMue: I added some explanations of what I envision here. Not sure what others are thinking about it.

make-intranet-public/README.md
```diff
@@ -0,0 +1,59 @@
...
+### Transitioning content into the public handbook
+
+Content will be transitioned into the public handbook on a "if-needed" basis for now.
+We have some usecases which require our internal docs to be public (e.g. relevant docs to external collaborators, people related docs which help with hiring, ... ).
+
+Once the new concept is proven with the currently needed docs being public, then a `public by default` policy should be put in place.
```
gianfranco-l, line 43: I suggest to do the following:
- define clearly what success look like (so clearly explicit what you mean by "new concept is proven")
- clearly set the tentative roadmap (eliciting steps/milestones) and set expectations for the "public by default" and "content migration" part (i.e. once handbook is "successful" then these roles X are expected to implement these Y features within Z time in this order T)

make-intranet-public/README.md
puja108: this is independent from the import, I agree that the import is a good idea at least for the current technical solution.

I want to make our guidelines more clear with a clear goal on defaulting to public and only keeping sensitive info private

puja108: I just saw you also mention the "public by default" policy further down, fine if you wanna stage it as you need to get unblocked

make-intranet-public/README.md
MarcelMue: I agree but I don't think the RFC is the place to do this already. I would generally like agreement on the repo split proposal first :)

gianfranco-l: then write it in the rfc the steps that are left for further later elicitation rather than writing half-baked ideas no?

teemow:
I like this proposal. I've never thought about separating a handbook and our public docs.

I agree that UX should be addressed separately. 

Outside collaboration (on the docs themselves) is not top prio imo. Transparency is the main driver. Being able to point people to how we do things etc. 

Although the public docs also use hugo and we would need an up to date design for the handbook etc. Which will bring it closer to the docs. So it could as well be just a part of our docs. And we could still try to merge the search in the intranet. This way we would find all docs through the intranet search. We'd also remove the barrier to move content between the handbook and the product docs, it's easier to link in between product docs and the handbook (because it is one repo) etc.

So the question for me is: What are the benefits of three hugo intallations for docs? Why not reduce it to one public and one private?

MarcelMue:
> So the question for me is: What are the benefits of three hugo intallations for docs? Why not reduce it to one public and one private?

My main reason currently is that the shift from private to public seems a lot harder if we have a single public `docs` page. I initially tried to come up with such a proposal but there are some core issues which I couldn't come up with solutions:
1. How can we prevent the public docs' search becoming useless for users/customers? We currently have curated articles and guides which come up in the public docs search. By adding things which we make public, it might become confusing for users which doc is applicable to them in the search.
2. How do we signal that some content is not meant for customer debugging? Some ops-recipes are IMO part of the content which we would make public. There has to be a disclaimer for each one to explain that only GS employees is able to use it. That doesn't seem like a nice UX thing from my POV, especially combined with 1.
3. We currently have quite high standards for documentation in our `docs`. Following these standards would either mean that we have to re-write a lot of our docs before making them public or our `docs` would become inconsistent in terms of quality.

Overall it just seems easier to me to achieve what we want through a separate repo, compared to tackling these hard problems.

teemow:
@marians what do you think about the last two comments from your UX point of view?

make-intranet-public/README.md
```diff
@@ -0,0 +1,59 @@
+# Making parts of the intranet public
+
+## Status quo
+
+We currently have an internal `hugo` page which we use for internal documentation.
+This page is called `intranet` and contains all kinds of documentation around our operational work as well as security and development.
+The sourcecode for our `intranet` is in a private github repository which contains other security and business relevant data.
```
marians, line 7:
```suggestion
The source code for our `intranet` is in a private GitHub repository which contains other security and business relevant data.
```

marians:
## Handbook content in two places

According to this proposal, handbook content is published in two places:

- for Giant Swarm staff, the handbook `intranet.giantswarm.io/handbook/`
- for everyone else, the handbook is `handbook.giantswarm.io`.

The reason mentioned in the draft is that with this, the intranet search would cover both internal and handbook content. This is good.

On the downside, each article would have two different URLs. If we as GS staff members wanted to point some outside collaborator to a piece of content we find via the intranet, we'd have to modify the URL in order to make it fit. Or go to the handbook site and find the content there, then copy the URL. Or build a widget that shows the public URL for each handbook page in the Intranet. All of this look like a big no-no to me. Sharing a URL is such a basic, universal part of interacting with a website. It would be extremely hard to teach everyone to be mindful about this.

Another concern I have with the handbook site being integrated into the intranet is internal links. Setting links from a handbook page to a non-public intranet page would have to be avoided, otherwise public users run against walls. Links from an intranet page to a handbook page, and links within the handbook section, must decide what path to use: either `/handbook/...` or `/...`. In order to work in both contexts, the hyperlink would have to be rewritten on one of the two places, which requires an additional technical solution and high consistency in our markdown content.

Another, less important aspect: If most content should be public, the "handbook" subsection would probably be larger than the entire rest of the intranet's site tree. This doesn't make for a great navigation experience if most intranet journeys would go through Front page > Documentation > Handbook > sub page.

So I am in favour of publishing the handbook content in one place only. There would be one URL for each page only. No mental load when sharing a URL. No fiddling with hyperlinks. The side effect would be that GS staff would have to think about where to find some content ("Is that handbook or rest of intranet?"). A search covering everything would help, but would not completely solve this. Sometimes we simply want to navigate.

### On search

I'm thinking that we can provide a search experience to both public users and GS staff users that fits their needs. The simple idea would be that

- if a GS staff member searches (maybe within the intranet), it covers both intranet and handbook
- if a public user searches, it covers handbook

The client-side search we have in our intranet isn't great anyway. Our docs search (which also covers blog posts, btw.) based on elasticsearch is much more capable and extensible. If we can solve the authorization part (let only GS staff find content from intranet), I think we can re-use/extend this for intranet and handbook.

With this, a search UI in the intranet could actually yield intranet, handbook, docs, blog content, if the users likes. Just like the docs search currently yields docs and blog content.

Marcel asked:

> How can we prevent the public docs' search becoming useless for users/customers?

By not presenting handbook content in docs search results. At a later point, if we find that docs search should also present handbook content if the users wants, we can extend the UI to make this optional and useful.

### Separate site for handbook

Whether the handbook content should be a separate site, or part of docs, to me depends on a bunch of things.

- **Target audience**. Is it different from the audience of docs? As I see it: yes. "Collaborators" haven't been defined more, but I reckon these are not our customers. Or there is little overlap. Which speaks for separate sites. There might be a need to move content from handbook to docs every now and then, which is fine.

- **Quality standards**: As Marcel pointed out, he expects the handbook content to have lower quality. While I would love to see high standards for all content, we must set priorities. And I wouldn't favor lowering the standards of all content in order to reach consistency. This, too, speaks for separate sites.

- Let's call it **Affinity**. Assuming the handbook content is written for a different audience than docs, let's imagine that each article could be neatly attached as an appendix etc. to a section in docs, then it might make more sense to integrate the handbook content in docs. This should then be accompanied require the mentioned disclaimers like "Content written for ..., take with grain of salt.". However, if we expect handbook content to be quite separate, it would actually become a challenge to accommodate for it in the docs navigation structure. This would speak for a separate site.

### Another alternative to rule out

It would also in theory be possible to publish handbook content as an intranet sub section e. g. `intranet.giantswarm.io/handbook/` and make that section publicly accessible. But I don't like that for the reason that it would expose a huge main navigation full of options that are not accessible to the public user.

MarcelMue:
> So I am in favour of publishing the handbook content in one place only.

I am ok to have it in one place - if -

> With this, a search UI in the intranet could actually yield intranet, handbook, docs, blog content, if the users likes. Just like the docs search currently yields docs and blog content.

We can implement this search with our custom search without too much effort. I have no idea how our elasticsearch search works - like at all. Therefore I have very little inside into the feasibility here. Would it be simpler to have 2 elastic based searches - the one we currently have and then a new one which covers everything and is only available to us for now?

MarcelMue:
> Outside collaboration (on the docs themselves) is not top prio imo. Transparency is the main driver. Being able to point people to how we do things etc.

I want to come back to this and clarify: Outside collaboration is the **MAIN** driving factor for me here. We want to have outside collaborators work on some of our managed app. To do this it's absolutely crucial that they can access some of our currently private docs AND collaborate / contribute to them.

I will also clarify this in the RFC text, but to me there is a real need here which needs to be fulfilled before any collaboration can start.

teemow:
I think the target audience will not be disjunctive. Many customers are transforming their organization and our culture, organization and communication structures and processes are interesting to them. 

But I leave it up to you guys what is better. Thanks for all the thoughtful considerations.

teemow:
@MarcelMue you misunderstood me (and probably there was some misunderstanding before). I was talking about outside collaborators to our handbook. I was not talking about enabling outside collaborator or I summarized that via "transparency".

marians:
Is there a misunderstanding about contributions to content (editing) vs. consuming/reading the content, possibly? If yes, maybe that should also be clarified.

Regarding search and one vs. 2 separate ES, I think that can be clarified outside the RFC. There will be pros and cons for both.

teemow:
@marians yes @MarcelMue was refering to my comment: https://github.com/giantswarm/rfc/pull/44#issuecomment-1210520329

And I was talking about content collaboration. But Marcel is talking about collaboration in the sense of having handbook docs for externals (which I called transparency).

Oshratn:
> The side effect would be that GS staff would have to think about where to find some content ("Is that handbook or rest of intranet?"). A search covering everything would help, but would not completely solve this. Sometimes we simply want to navigate.

In the long run, this would be solved, because GS staff will get used to looking at docs (including handbook) first. Even now, when searching for something I find myself sifting through Docs/Intranet/Github repos. IMO, the problem already exists and if what we are doing here makes the problem smaller, not bigger then I would count it as a win.

MarcelMue:
Can we get some agreement that the general idea in the RFC is a good starting point? I want to get started with e.g. defining milestones and tasks & work towards making the first docs public.

MarcelMue:
> Minor question: I don't know what this means:
> 
> > Most importantly, it becomes trivial to link URLs from either source as search results will forward you to the specific article.

Basically just saying that the URL will be the actual piece of content - not the imported content like I initially suggested.

teemow (approved):
Happy. We will improve along the way.

make-intranet-public/README.md
```diff
@@ -0,0 +1,63 @@
...
+- `intranet` only contains truly private information but imports all content from `handbook` into a unified search.
+- `docs` contains product documentation which is directly linked on our website as before.
+
+### Transitioning content into the public handbook
+
+Content will be transitioned into the public handbook on a "if-needed" basis for now.
+We have some usecases which require our internal docs to be public (e.g. relevant docs to external collaborators, people related docs which help with hiring, ... ).
+
+Once the new concept is proven with the currently needed docs being public, then a `public by default` policy should be put in place.
+This should then by followed by all non-critical docs being moved to the public repo in quick succession.
```
pipo02mix, line 48: Should we decide a initial structure that will help ppl to move content?

Example

-> Operations
    -> Operation model
    -> Run books
    -> ...

-> Organization structure
    -> Chapters
    -> SIGs
    -> ...
    
-> ...
    
I think it will be useful if we put some effort into drawing an outline instead of leaving ppl to classify content freely

make-intranet-public/README.md
```diff
@@ -0,0 +1,63 @@
...
+We currently have an internal `hugo` page which we use for internal documentation.
+This page is called `intranet` and contains all kinds of documentation around our operational work as well as security and development.
+The source code for our `intranet` is in a private GitHub repository which contains other security and business relevant data.
+
+## Goal
+
+The non-security relevant parts of the intranet should be public.
+Making them public enables us to collaborate with other companies more easily.
+It additionally allows us to enable outside contributors to work on projects more independently.
+
```
Oshratn, line 14: I'd add that it will also work towards branding and especially employer branding.

make-intranet-public/README.md
Oshratn: I like this idea.

make-intranet-public/README.md
```diff
@@ -0,0 +1,63 @@
...
+### Transitioning content into the public handbook
+
+Content will be transitioned into the public handbook on a "if-needed" basis for now.
+We have some usecases which require our internal docs to be public (e.g. relevant docs to external collaborators, people related docs which help with hiring, ... ).
+
+Once the new concept is proven with the currently needed docs being public, then a `public by default` policy should be put in place.
+This should then by followed by all non-critical docs being moved to the public repo in quick succession.
+
+Docs should always be made public by the owners of the docs.
```
Oshratn, line 50: Which means we need to ensure Intranet docs ownership, similar to what we have on Docs at the moment.

puja108: IIRC we did give ownership of intranet to team team

make-intranet-public/README.md
```diff
@@ -0,0 +1,63 @@
...
+Therefore the `intranet` can remain as the single source of truth for Giant Swarm employees to aggregate information.
+Most importantly, it becomes trivial to link URLs from either source as search results will forward you to the specific article.
+
+The `handbook` repository itself should also be setup with `hugo` to host its content under `handbook.giantswarm.io` making it easily accessible for outside collaborators.
+
+The complete setup would then look as follows:
+- `handbook` only contains information we are comfortable with having public.
+- `intranet` only contains truly private information but imports all content from `handbook` into a unified search.
+- `docs` contains product documentation which is directly linked on our website as before.
+
```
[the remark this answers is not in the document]
MarcelMue: Yes, this is something I want in the future to advertise it. Initially this is not a requirement.

pipo02mix (approved):
LGTM by now

PR #46, opened 2022-09-12
Add RFC regarding single point of failure in management cluster access

anvddriesch:
towards https://github.com/giantswarm/roadmap/issues/1384

### DIFF
--- a/management-cluster-access/README.md
+++ b/management-cluster-access/README.md
@@ -8,7 +8,7 @@ For Giant Swarm staff this is done via `dex` using github as the _only_ identity
 Our vintage product also offers the possibility to create a client certificate to access the management cluster.
 This means that in the event of github or `dex` being unavailable, we do not get locked out of the management cluster.
 
-For pure CAPI installations on the other hand, no other authentication method is available at this point.
+For pure CAPI installations on the other hand, the only fallback authentication method is a single static kubeconfig that is stored in lastpass for emergency access.
 This poses a real risk to get locked out of the management cluster in case of problems with OIDC.
 We also rely on the github api being available as an external dependency we can not control.
 
@@ -41,6 +41,7 @@ At the moment this is not the case for the CAPI product.
 - A long living client-certificate granting admin access is a security risk since permissions can not be revoked.
 - We do not want this to be the primary authentication method for human users. (That should remain SSO)
 - If we want to support this, we need to decide whether we want to use vault or something else.
+- Whichever method we choose, we need to ensure that TTL duration is limited to minimize the security risk. 
 
 ### Storing kubeconfig in lastpass
 
@@ -48,8 +49,11 @@ In the CAPI product, a valid kubeconfig is stored in lastpass as emergency fallb
 
 - This method of authentication is completely separate from SSO and does not have shared dependencies.
 - It is not integrated in our tooling and we need to pull the kubeconfig using lastpass cli.
-- We should revisit how this works. What type of authorization is used here? Is access revokable? Is this rotated?
+- As it is right now, access is neither easily revokable nor is there a rotation. 
+- In case of a security threat this implies rotating the api server CA and all certificates.
 - Since it is already integrated, it makes sense to keep it for emergencies while there is no alternative.
+- If we want to support this for a longer time we need to improve security.
+- We could support this as fallback method in `opsctl login` by managing access to `lastpass` or calling the `lastpass` cli to ease operations.
 
 ### Introduce a second identity provider for SSO
 
--- a/management-cluster-access/README.md
+++ b/management-cluster-access/README.md
@@ -1,0 +1,95 @@
+# Ensure no single point of failure in management cluster access
+
+## Access to MCs for giant swarm staff in vintage vs the CAPI product
+
+Our default access method to management clusters is now single-sign-on.
+For Giant Swarm staff this is done via `dex` using github as the _only_ identity provider. 
+
+Our vintage product also offers the possibility to create a client certificate to access the management cluster.
+This means that in the event of github or `dex` being unavailable, we do not get locked out of the management cluster.
+
+For pure CAPI installations on the other hand, no other authentication method is available at this point.
+This poses a real risk to get locked out of the management cluster in case of problems with OIDC.
+We also rely on the github api being available as an external dependency we can not control.
+
+Furthermore, loss of access to the management clusters implies loss of access to all workload clusters that do not have OIDC access for giant swarm staff set up.
+This is due to the fact that in order to create a client certificate for these workload clusters, we need to access the management cluster first.
+At this moment this likely applies to all workload clusters on pure CAPI installations.
+
+However, even if OIDC access was set up, it would still leave github as a single point of failure.
+
+Let's ensure that management cluster access is highly available for operations through GS staff in the future.
+
+## References
+
+- [PKI epic](https://github.com/giantswarm/giantswarm/issues/15981)
+- [Issue regarding client cert creation for CAPO ](https://github.com/giantswarm/giantswarm/issues/21740)
+- [story for backup idp in dex](https://github.com/giantswarm/roadmap/issues/603)
+- [proposal to use azure AD for GS staff](https://github.com/giantswarm/giantswarm/issues/21627)
+- 
+
+## Possible solutions
+
+### Client certificates
+
+In the vintage product, client certificates can be created from outside the cluster using `vault` for signing.
+At the moment this is not the case for the CAPI product.
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- As of now, client certificate creation is harder to monitor and control than SSO.
+- Client certificates should be short lived and therefore would need to be recreated by users frequently.
+- A long living client-certificate granting admin access is a security risk since permissions can not be revoked.
+- We do not want this to be the primary authentication method for human users. (That should remain SSO)
+- If we want to support this, we need to decide whether we want to use vault or something else.
+
+### Storing kubeconfig in lastpass
+
+In the CAPI product, a valid kubeconfig is stored in lastpass as emergency fallback. This is part of the bootstrap process.
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- It is not integrated in our tooling and we need to pull the kubeconfig using lastpass cli.
+- We should revisit how this works. What type of authorization is used here? Is access revokable? Is this rotated?
+- Since it is already integrated, it makes sense to keep it for emergencies while there is no alternative.
+
+### Introduce a second identity provider for SSO
+
+Dex supports a wide array of identity providers. Customers are already using many of them. At the present moment, giant swarm staff can only login using github.
+However, we could add more providers.
+
+- We favor SSO when it comes to identity management for human engineers.
+- Could be introduced for workload cluster access as well.
+- Using e.g. azure AD would make sense since customers are using it.
+- We might be able to simplify account automation when using another identity provider than github.
+- We would still have a dependency on dex and this method will not completely remove the single point of failure.
+- Adding another provider might mean individual setup for each existing configuration. (e.g. in github individual apps are needed) We should avoid this if possible since it already does not scale.
+- While dex allows an array of connectors, we currently use a fixed naming for our single `giantswarm` connector. We need to revisit what relies on this convention to ensure the second connector is equal.
+- Regardless of other means auf authentication being implemented, adding another dex idp would be beneficial.
+
+### Service account tokens
+
+Other than client certficiates, we could also look at service accounts as a means of managing access.
+
+- Service accounts can be rotated so they carry less risk in that regard than client certificates.
+- This should not be a standard access method for human engineers.
+- We would need to discover how to implement this. A dedicated controller for managing service account access is thinkable.
+
+### Other access methods
+
+- To be discussed
+
+### Related stories
+
+This discussion is focused on management cluster access by GS staff and removing the single point of failure here should remain the goal.
+However, our chosen solution for other access related problems has impact on the issue of management cluster access by giant swarm staff. We should aim to resuse the same mechanisms for workload clusters and management clusters as much as possible. Likewise we should strive to dogfood mechanisms we offer to our customers.
+On the other hand, we should be conscious of differences between access for automation and controllers versus access for humans. 
+Related problems include:
+
+- Access to management clusters for controllers
+- Access to management clusters for customers
+- Access to workload clusters for controllers
+- Access to workload clusters by gs staff
+- Access to workload clusters by customers
+
+### Next steps
+
+- to be discussed
--- a/management-cluster-access/README.md
+++ b/management-cluster-access/README.md
@@ -2,21 +2,21 @@
 
 ## Access to MCs for giant swarm staff in vintage vs the CAPI product
 
-Our default access method to management clusters is now single-sign-on.
-For Giant Swarm staff this is done via `dex` using github as the _only_ identity provider. 
+Our default access method to management clusters is now single-sign-on (SSO).
+For Giant Swarm staff this is currently done via Dex using GitHub as the _only_ identity provider. 
 
 Our vintage product also offers the possibility to create a client certificate to access the management cluster.
-This means that in the event of github or `dex` being unavailable, we do not get locked out of the management cluster.
+This means that in the event of GitHub or Dex being unavailable, we do not get locked out of the management cluster.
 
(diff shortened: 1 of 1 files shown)


management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+However, even if OIDC access was set up, it would still leave github as a single point of failure.
+
+Let's ensure that management cluster access is highly available for operations through GS staff in the future.
+
+## References
+
+- [PKI epic](https://github.com/giantswarm/giantswarm/issues/15981)
+- [Issue regarding client cert creation for CAPO ](https://github.com/giantswarm/giantswarm/issues/21740)
+- [story for backup idp in dex](https://github.com/giantswarm/roadmap/issues/603)
+- [proposal to use azure AD for GS staff](https://github.com/giantswarm/giantswarm/issues/21627)
```
pipo02mix, line 28:
```suggestion
- [Story for backup IDP in dex](https://github.com/giantswarm/roadmap/issues/603)
- [Proposal to use Azure AD for GS staff](https://github.com/giantswarm/giantswarm/issues/21627)
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Client certificates
+
+In the vintage product, client certificates can be created from outside the cluster using `vault` for signing.
+At the moment this is not the case for the CAPI product.
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- As of now, client certificate creation is harder to monitor and control than SSO.
+- Client certificates should be short lived and therefore would need to be recreated by users frequently.
+- A long living client-certificate granting admin access is a security risk since permissions can not be revoked.
```
pipo02mix, line 41: We can limit TTL duration in our API

anvddriesch: Agreed. IMO that has to be a requirement if we want to support this.

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+- If we want to support this, we need to decide whether we want to use vault or something else.
+
+### Storing kubeconfig in lastpass
+
+In the CAPI product, a valid kubeconfig is stored in lastpass as emergency fallback. This is part of the bootstrap process.
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- It is not integrated in our tooling and we need to pull the kubeconfig using lastpass cli.
+- We should revisit how this works. What type of authorization is used here? Is access revokable? Is this rotated?
+- Since it is already integrated, it makes sense to keep it for emergencies while there is no alternative.
```
marians, line 52: I think we can add here that integration in our tooling (`opsctl login`) should be considered as an option. opsctl could somehow handle the user's LastPass authentication (e. g. via OAuth and the browser) and then retrieve the kubeconfig, which must be stored under a specific name.

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+However, we could add more providers.
+
+- We favor SSO when it comes to identity management for human engineers.
+- Could be introduced for workload cluster access as well.
+- Using e.g. azure AD would make sense since customers are using it.
+- We might be able to simplify account automation when using another identity provider than github.
+- We would still have a dependency on dex and this method will not completely remove the single point of failure.
+- Adding another provider might mean individual setup for each existing configuration. (e.g. in github individual apps are needed) We should avoid this if possible since it already does not scale.
+- While dex allows an array of connectors, we currently use a fixed naming for our single `giantswarm` connector. We need to revisit what relies on this convention to ensure the second connector is equal.
+- Regardless of other means auf authentication being implemented, adding another dex idp would be beneficial.
```
marians, line 66: My 2 cts here is that using GitHub as identity provider isn't ideal, as it needs a separate OAuth app for every (management and workload) cluster we use Dex with.

So I would definitely favour a more flexible provider like Azure AD, where we can have a single application configured with multiple callback URLs. That should be our default IdP. Our second (backup) IdP should be as flexible as the primary one.

Regarding connector IDs: I think Dex requires unique connector identifiers, so one could be `giantswarm1` and the second could be `giantswarm2`. Of course this has an effect on our RBAC bindings, as the connector IDs have to be present in the group names. So instead of having one group `giantswarm:my-admin-group-name` become admins, we need two groups to be bound: `giantswarm1:my-admin-group-name-from-idp1`, `giantswarm2:my-admin-group-name-from-idp2`.

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Service account tokens
+
+Other than client certficiates, we could also look at service accounts as a means of managing access.
```
marians, line 70:
```suggestion
Other than client certficates, we could also look at service accounts as a means of managing access.
```

management-cluster-access/README.md
anvddriesch: Good point! I will add this.

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Access to MCs for giant swarm staff in vintage vs the CAPI product
+
+Our default access method to management clusters is now single-sign-on.
+For Giant Swarm staff this is done via `dex` using github as the _only_ identity provider. 
+
+Our vintage product also offers the possibility to create a client certificate to access the management cluster.
+This means that in the event of github or `dex` being unavailable, we do not get locked out of the management cluster.
+
+For pure CAPI installations on the other hand, no other authentication method is available at this point.
```
puja108, line 11: There's a CAPi kubeconfig that gets created on MC creation and that we store in Lastpass AFAIK

management-cluster-access/README.md
```diff
@@ -0,0 +1,95 @@
...
+- We do not want this to be the primary authentication method for human users. (That should remain SSO)
+- If we want to support this, we need to decide whether we want to use vault or something else.
+
+### Storing kubeconfig in lastpass
+
+In the CAPI product, a valid kubeconfig is stored in lastpass as emergency fallback. This is part of the bootstrap process.
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- It is not integrated in our tooling and we need to pull the kubeconfig using lastpass cli.
+- We should revisit how this works. What type of authorization is used here? Is access revokable? Is this rotated?
```
puja108, line 51: No, and only way to rotate it would be rotating API server CA and all certs, it's not a great thing security wise

management-cluster-access/README.md
anvddriesch: I agree about github. It's a hassle especially when we look at rolling out oidc access for gs staff to workload clusters, it just does not scale.
At the same time, we are using it now and it works. We would likely have to do something like adding another provider (e.g. azure) first, then eventually adding another one and then removing github/no longer supporting it in new installations. 

As for the connector naming, we have to get it right from the beginning. Using `giantswarm1` etc could work or perhaps something more specific like `giantswarm-ad`. We should make sure that connectors and related rbac configuration can be easily removed, replaced and added. 

When it comes to selecting the connector, I personally like the idea of the selection menu. It is in line with how SSO login is generally presented and we wouldn't have to know/manage flag values if something changes (Looking at SSO on WCs here, too)
Shouldn't the extra click be fine for the most part? We do not need to log in again too often since our context can be refreshed.

anvddriesch:
Waiting for @marians review.

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
+# Ensure no single point of failure in management cluster access
+
+## Access to MCs for giant swarm staff in vintage vs the CAPI product
+
+Our default access method to management clusters is now single-sign-on.
```
marians, line 5:
```suggestion
Our default access method to management clusters is now single-sign-on (SSO).
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+For Giant Swarm staff this is done via `dex` using github as the _only_ identity provider.
```
marians, line 6:
```suggestion
For Giant Swarm staff this is currently done via Dex using GitHub as the _only_ identity provider. 
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Access to MCs for giant swarm staff in vintage vs the CAPI product
+
+Our default access method to management clusters is now single-sign-on.
+For Giant Swarm staff this is done via `dex` using github as the _only_ identity provider. 
+
+Our vintage product also offers the possibility to create a client certificate to access the management cluster.
+This means that in the event of github or `dex` being unavailable, we do not get locked out of the management cluster.
+
+For pure CAPI installations on the other hand, the only fallback authentication method is a single static kubeconfig that is stored in lastpass for emergency access.
```
marians, line 11:
```suggestion
For pure CAPI installations on the other hand, the only fallback authentication method is a single static kubeconfig that is stored in LastPass for emergency access.
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+This poses a real risk to get locked out of the management cluster in case of problems with OIDC.
```
marians, line 12: Since there is a fallback kubeconfig in the password manager, this is not really true, is it? However, access is somewhat inconvenient.

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+We also rely on the github api being available as an external dependency we can not control.
```
marians, line 13:
```suggestion
We also rely on the GitHub API being available as an external dependency we can not control.
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Furthermore, loss of access to the management clusters implies loss of access to all workload clusters that do not have OIDC access for giant swarm staff set up.
```
marians, line 15:
```suggestion
Furthermore, loss of access to the management clusters implies loss of access to all workload clusters that do not have OIDC access for Giant Swarm staff set up.
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+This is due to the fact that in order to create a client certificate for these workload clusters, we need to access the management cluster first.
+At this moment this likely applies to all workload clusters on pure CAPI installations.
+
+However, even if OIDC access was set up, it would still leave github as a single point of failure.
```
marians, line 19:
```suggestion
However, even if OIDC access was set up, it would still leave GitHub as a single point of failure.
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+- This method of authentication is completely separate from SSO and does not have shared dependencies.
+- As of now, client certificate creation is harder to monitor and control than SSO.
+- Client certificates should be short lived and therefore would need to be recreated by users frequently.
+- A long living client-certificate granting admin access is a security risk since permissions can not be revoked.
+- We do not want this to be the primary authentication method for human users. (That should remain SSO)
+- If we want to support this, we need to decide whether we want to use vault or something else.
+- Whichever method we choose, we need to ensure that TTL duration is limited to minimize the security risk. 
+
+### Storing kubeconfig in lastpass
```
marians, line 46:
```suggestion
### Storing kubeconfig in LastPass
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+In the CAPI product, a valid kubeconfig is stored in lastpass as emergency fallback. This is part of the bootstrap process.
```
marians, line 48:
```suggestion
In the CAPI product, a kubeconfig is stored in LastPass as an emergency fallback, during the bootstrap process.
```

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Other than client certificates, we could also look at service accounts as a means of managing access.
+
+- Service accounts can be rotated so they carry less risk in that regard than client certificates.
+- This should not be a standard access method for human engineers.
+- We would need to discover how to implement this. A dedicated controller for managing service account access is thinkable.
+
+### Other access methods
+
+- To be discussed
```
bavarianbidi, line 82: Having a secret-engine (e.g. hashicorp `vault` "somewhere" else), we could create the CA from this secret-engine before the cluster got created and generate all required client-certificates from that CA.

Whenever we need to access the API-Server, we could create "on-demand" a new client-certificate (with a short TTL) from that CA (by using the remote secret-engine). 

Pros:
* no dependency to a running management-cluster for accessing a workload cluster
* Accessing the Workload Cluster via a short lived client-certificate is also possible
* secret-engine could be used to store/share more cluster-specific secretes with other components (out of scope for this discussion)

Cons:
* secret-engine must exist

bavarianbidi, line 82: Creating a controller (running on each MC) which creates a client-certificate per existing cluster (from the in-management cluster stored CA) and write this secret back to `lastpass` (e.g. via [`lastpass-go`(https://github.com/ansd/lastpass-go)]

Pros:
* TTL of the client-certificate could be configurable
* no dependency to a running management-cluster for accessing a workload cluster
* Accessing the management-cluster via that client-certificate is also possible

> **Note**
> * does lastpass has support for "technical users" (via tokens)
> * we have to check of we are able to create some kind of folder per installation

bavarianbidi, line 82: Creating a controller (running on each MC) which creates a client-certificate per existing cluster (from the in-management cluster stored CA) and write this secret back to a gitrepo via `sops`.

Pros:
* By revoking access to the "specific" git-repo offboarding might be easy as the encrypted kubeconfigs might be only valid for ~1 week (this might reduce issues with github unavailability)
* we already start using `sops` more and more

Cons:
* might lead to merge conflicts

management-cluster-access/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Access to management clusters for controllers
+- Access to management clusters for customers
+- Access to workload clusters for controllers
+- Access to workload clusters by gs staff
+- Access to workload clusters by customers
+
+## Next steps
+
+### Adding azure AD as a second identity provider for SSO
```
bavarianbidi, line 99: out of curiosity: why not using `google` as second OIDC provider? Everyone at GS has a google-account per default. Azure AD is something additional?

marians: The Google connector in Dex is marked as "alpha", while the Azure one is mature, and we don't have any experience with it.

We already maintain user accounts in Azure AD.

management-cluster-access/README.md
anvddriesch: If I am not mistaken, this would be similar to what we are doing at the moment for vintage management clusters. However, you suggest this to be used for workload clusters well, right?
I agree that it would be worth looking into when it comes to secret and certificate management for other components and perhaps replace the default CAPI mechanisms.
The big question would be which team would own this since it's quite a large story.

anvddriesch: So basically, the controller would ensure that this certificate is rotated often, right?
It would be a big improvement in comparison to just creating it on bootstrap.

anvddriesch: I guess this method would be a bit similar to the one above with lastpass but we could perhaps integrate a little bit better with the gitops approach.

bavarianbidi: > So basically, the controller would ensure that this certificate is rotated often, right?

I want change anything for the cluster bootstrap/lifecycle. I just thought about creating a "short-lived" client-certificate which could used to authenticate against a kubernetes API. And to remove the runtime-dependency to a management-cluster, we could sync (periodically) this client-certificate back to `keepass`. (And as a management-cluster is from a CAPI point of view similar to a workload cluster, this will also give us "emergency" access to the management cluster)

bavarianbidi: yes, same as above but with another backend. Just wanted to not mix the discussion of possible backends for that approach.

bavarianbidi: > you suggest this to be used for workload clusters well, right?

Yes, currently the certificates/CA got generated in the MC. Why not generating the certificates/CA from an external system instead.

> The big question would be which team would own this since it's quite a large story.

* Providing a "central" secret-engine API - Team shield (or e.g. using Vault as a Service)
* Creating controllers to generate required secrets/CAs - KaaS Teams for creating CAs and Secrets for cluster lifecycle, Honeybadger for creating secrets required by other Apps

pipo02mix: The only problem is then we have a central place (somewhere) with credentials for all customers. I think this was main reason to have a Vault for customer. Not sure if legally we can have all in one place

puja108: as long as we have a way to separate/isolate secrets within that secret store, e.g. with access control and separate accounts, we're ok

PR #48, opened 2022-10-21
Crossplane deployment RFC

kubasobon:
Towards <https://github.com/giantswarm/giantswarm/issues/23934>
WIP

### DIFF
--- a/crossplane/README.md
+++ b/crossplane/README.md
@@ -104,12 +104,43 @@ varies, depending on the Provider.
 
 #### The list of Providers is pre-defined and immutable
 
+We deploy a provider that best matches the Cloud Provider's infrastructure:
+- AWS <https://github.com/crossplane-contrib/provider-aws>
+- Azure <https://github.com/crossplane-contrib/provider-azure>
+- CAPA <https://github.com/crossplane-contrib/provider-aws>?
+- GCP <https://github.com/crossplane/provider-gc>
+- KVM not supported; obviously it's customers' own infrastructure
+- OpenStack TBD (<https://github.com/crossplane-contrib/provider-terraform>???)
+
+No other providers are available, even upon request. The offering is "managed".
+
 #### Customers can pick and choose from a well defined list of vetted Providers
+
+We offer a curated selection of [all Crossplane
+providers](https://github.com/orgs/crossplane-contrib/repositories). Clients
+are free pick the ones they like for us to install. The offering is "managed".
 
 #### Customers can request any Provider they like
 
+Customers can select any of Provider they like to be installed. This means
+Giant Swarm tests and vets the Provider to be able to support the customer and
+truly "manage" the deployment.
+
 #### Customers can install any Provider they like
+
+Customers are free to create and configure `Provider` resources however and
+whenever they like. Only stage 1 is managed. Customers are responsible for
+running Providers much like with ArgoCD and custom operators on Management
+Clusters.
+
+#### 4. OpenStack provider
+
+To our best knowledge, there are no dedicated OpenStack Crossplane providers.
 
 ## Decision
 
+TBD
+
 ### Next steps
+
+TBD
--- a/crossplane/README.md
+++ b/crossplane/README.md
@@ -2,11 +2,11 @@
 
 Managed Crossplane offering is under development as per customers' requests.
 This document discusses all the required elements of Crossplane's deployment
-and design decisions we make along the way.
+and the design decisions we make along the way.
 
 ## Crossplane building blocks
 
-For the ease of describing each proposal, we will divide a fully functional
+For ease of describing each proposal, we will divide a fully functional
 Crossplane deployment into three stages.
 
 ### Stage 1 - Crossplane
@@ -32,7 +32,7 @@ infrastructure yet.
 ### Stage 2 - Crossplane Providers
 
 Infrastructure is managed by Providers. Creating a
-`providers.pkg.crossplane.io` resource will cause Crossplane operator's to
+`providers.pkg.crossplane.io` resource will cause Crossplane operators to
 deploy provider operators and all necessary resources.
 
 - Provider-specific Operator
@@ -43,13 +43,13 @@ deploy provider operators and all necessary resources.
 ### Stage 3 - Provider Configuration
 
 Just deploying Providers is not enough for them to be functional. They still
-need to be configured with passwords, secrets, access keys, API keys and the
-like. Each Provider comes with their own configuration CRD, usually
+need to be configured with passwords, secrets, access keys, API keys, and the
+like. Each Provider comes with its configuration CRD, usually
 `providerconfigs.<provider-name>.crossplane.io`.
 
-For example, AWS provider will require a `ProviderConfig` similar to the one
-below to be able to perform reconciliation. `crossplane-system/aws-creds`
-Secret should contain AWS Access Key.
+For example, the AWS provider will require a `ProviderConfig` similar to the
+one below to be able to perform reconciliation. `crossplane-system/aws-creds`
+The Secret should contain AWS Access Key.
 
 ```yaml
 apiVersion: aws.crossplane.io/v1beta1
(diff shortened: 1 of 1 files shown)


kubasobon:
@giantswarm/team-honeybadger  It's a draft, because no decisions are taken yet, but PTAL and reflect on the questions.

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+  - `locks.pkg.crossplane.io`
+  - `providerrevisions.pkg.crossplane.io`
+  - `providers.pkg.crossplane.io`
+  - `storeconfigs.secrets.crossplane.io`
+- Other Kubernetes resources
+
+### Stage 2 - Crossplane Providers
+
+Infrastructure is managed by Providers. Creating a
+`providers.pkg.crossplane.io` resource will cause Crossplane operators to
```
uvegla, line 35: From what I have seen the providers can be installed as a configuration as well, for example in the official docs:

- https://crossplane.io/docs/v1.9/getting-started/install-configure.html#install-configuration-package

but the AWS Jet Provider is installed as:

- https://github.com/crossplane-contrib/provider-jet-aws#getting-started

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Stage 3 - Provider Configuration
+
+Just deploying Providers is not enough for them to be functional. They still
+need to be configured with passwords, secrets, access keys, API keys, and the
+like. Each Provider comes with its configuration CRD, usually
+`providerconfigs.<provider-name>.crossplane.io`.
+
+For example, the AWS provider will require a `ProviderConfig` similar to the
+one below to be able to perform reconciliation. `crossplane-system/aws-creds`
```
uvegla, line 51: Just FTR, the provider config can be / is different per provider, e.g. for AWS Jet:

```
apiVersion: aws.jet.crossplane.io/v1alpha1
kind: ProviderConfig
```

https://doc.crds.dev/github.com/crossplane-contrib/provider-jet-aws/aws.jet.crossplane.io/ProviderConfig/v1alpha1@v0.5.0-preview

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+providers](https://github.com/orgs/crossplane-contrib/repositories). Clients
+are free to pick the ones they like for us to install. The offering is "managed".
+
+#### Customers can request any Provider they like
+
+Customers can select any of the Providers they like to have. It means
+Giant Swarm tests and vets the Provider to be able to support the customer and
+"manage" the deployment.
+
+#### Customers can install any Provider they like
```
uvegla, line 129: To do this we only need to give access to a [handful of resources](https://github.com/crossplane/crossplane/tree/master/cluster/crds), which we need to do anyway I think.

This gives a lot of space and freedom for the customer I think and makes it easier for us to manage Crossplane. With managing the providers too, even a fixed subset - which constrains the customer - we need to manage a map of provider configs specific to the given providers. I think this does not mean managing the secrets, but they basically do not work without the secrets. I don't think GS should manage the secrets so it makes no sense and just extra work - so to say - to mange the provider configs as they basically add value only that a single ~20 line long yaml is not needed to be managed by the customer. And worst on top of that is that it will bring the same mistake we committed with e.g. `flux-app` that we manage the app, the CRDs and the CRs using those CRDs in the same Helm Chart.

So in short I feel like just managing crossplane is the best way and give customers access to those CRDs, their cluster admins can install whatever provider they need and manage the 2 simple configs (provider and secret) for it. (Actually maybe maybe they want to manage more, cos it could be useful when you want to give different resources different permissions, I think this is one great benefit of the creds / perms being in. the provider config in Crossplane, decoupled and you just tell which one to use when you want to use it). Customers having access to the provider logs they can debug issues too.

The only drawback I see here is that since whatever provider can be installed, the usual security risks and the stability of the MC because of this. But since we want to give more access / control over the MC to customers it may not be an issue (or issue for another day).

crossplane/README.md
kubasobon: Using CLI or supplying a Helm chart value does the same - it creates a `Provider` CR. Am I missing something?

crossplane/README.md
kubasobon: Yup, every provider's config kind is `ProviderConfig` but the `apiVersion`s differ.

crossplane/README.md
uvegla: Just thinking that there are multiplel ways to install packages, hence the `install configuration` and `install provider` commands for `kubectl crossplane`. From what I can tell, with `configuration` you can set constraints and multiple providers e.g: https://crossplane.io/docs/v1.9/getting-started/create-configuration.html#build-and-push-the-configuration compared to a raw provider: https://github.com/crossplane-contrib/provider-aws/blob/master/package/crossplane.yaml

kubasobon:
@piontec Would you kindly take a look here? Thanks :)

crossplane/README.md
piontec: I think it works by creating a Provider object no matter which way you request it.

crossplane/README.md
piontec: We can't allow customers to install any provider right now, at least not without some security tuning, that we might not have time for right now. If we allow the customers to install any provider, it's enough to install `provider-kubernetes` to extract any object in the cluster. So, to get started and complete our MVP on time, I suggest to limit the list of  providers and block installation by customers with kyverno

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+In this case, Giant Swarm is responsible for all three stages of deployment,
+including Provider Configuration. This means we need to store sensitive data
+(AWS Access Key, GCP Account Keyfile, Azure Service Principal Keys, etc.),
+support their encryption and decryption, and deliver them to Management
+Clusters.
+
+#### Giant Swarm manages Crossplane, the customer provides the keys
+
+In this case, Giant Swarm is responsible for stages one and two. Our goal is to
+deliver Crossplane and a set of Providers. It is up to the customer to create a
```
piontec, line 98: Yeap, that's what I think makes the most sense

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+We deploy a provider that best matches the Cloud Provider's infrastructure:
+- AWS <https://github.com/crossplane-contrib/provider-aws>
+- Azure <https://github.com/crossplane-contrib/provider-azure>
+- CAPA <https://github.com/crossplane-contrib/provider-aws>?
+- GCP <https://github.com/crossplane/provider-gc>
+- KVM is not supported; obviously, it's the customers' infrastructure
+- OpenStack TBD (<https://github.com/crossplane-contrib/provider-terraform>???)
+
+No other providers are available, even upon request. The offering is "managed".
```
piontec, line 115: I believe we have to start with that, otherwise we're not gonna make the security right in time (nor should we, assuming we're switching towards customer-with-MC-admin pattern). This is the simplest and safest solution we can come up with in a limited time we have.

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Customers can pick and choose from a well-defined list of vetted Providers
```
piontec, line 117: I don't think we have time to evaluate implications of installing all the different providers

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+We offer a curated selection of [all Crossplane
+providers](https://github.com/orgs/crossplane-contrib/repositories). Clients
+are free to pick the ones they like for us to install. The offering is "managed".
+
+#### Customers can request any Provider they like
```
piontec, line 123: Avoid like hellfire ;P

crossplane/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+Customers can select any of the Providers they like to have. It means
+Giant Swarm tests and vets the Provider to be able to support the customer and
+"manage" the deployment.
+
+#### Customers can install any Provider they like
+
+Customers are free to create and configure `Provider` resources however and
```
piontec, line 131: Yes, but only after we solve the shared-secrets problem and swich into customers-are-MC-admins mode

kubasobon:
@giantswarm/team-honeybadger PTAL. I've added decision points and a note about MVP.

uvegla (approved):

kubasobon:
@giantswarm/team-honeybadger Please take a final look :)

piontec (approved):

PR #49, opened 2022-11-03
Add RFC for Container Registry Configuration

erkanerol:
This PR adds an RFC regarding how we want to manage container registry configuration for our clusters.

A reference implementation in cluster app: https://github.com/giantswarm/cluster-cloud-director/pull/79
Configuration of management clusters during bootstrap: https://github.com/giantswarm/mc-bootstrap/pull/429

### DIFF
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -143,7 +143,6 @@ We don't want to add another responsility to `cluster-apps-operator`.
 
 `6.a` is selected.
 
-
 ### 7. Where to put the credentials
 
 #### 7.a Customer's Git Repository
@@ -160,8 +159,7 @@ We can use catalog configurations (See <mc-name>/appcatalog folder in `installat
 
 #### Decision
 
-**NOT DECIDED YET**
-
+7.c is selected.
 
 ### 8. Configuration interface
 
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -141,7 +141,7 @@ We don't want to add another responsility to `cluster-apps-operator`.
 
 #### Decision
 
-**NOT DECIDED YET**
+`6.a` is selected.
 
 
 ### 7. Where to put the credentials
@@ -153,6 +153,10 @@ We can put the secret into customers' git repositories. Customers will access to
 #### 7.b Management Cluster Fleet
 
 We can put the secret into our `management-clusters-fleet` repo. We will need to give reference to that secret in customers' git repository. Management of these secrets can be so costly when we need to create the secret in each organization namespace.
+
+#### 7.c Catalog configuration
+
+We can use catalog configurations (See <mc-name>/appcatalog folder in `installations` repo) to provide a MC specific configuration to all charts in the cluster.
 
 #### Decision
 
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -27,14 +27,15 @@ explicit k8s.gcr.io 12
 
 - In vintage, we configure [containerd for docker registry mirrors](https://github.com/giantswarm/giantnetes-terraform/blob/13700c6d1b1adf8d65fba8b1b37eccf31e1ce4f3/templates/files/conf/containerd-config.toml#L37). 
   
-## Possible solutions
+## Design proposals
 
+### 1. How to solve ImagePullErrors
 
-### ImagePullSecret
+#### 1.a ImagePullSecret
 
 We can set `imagePullSecrets` for all pods with an admission hook for authentication but Kubernetes doesn't have any mechanism like registry mirrors. That is why it is not a viable option.
 
-### FairwindsOps/saffire
+#### 1.b FairwindsOps/saffire
 
 There is an open-source project to patch pods in `imagePullErrors` error state: https://github.com/FairwindsOps/saffire
 
@@ -56,7 +57,7 @@ spec:
 This mechanism doesn't work for static pods and it doesn't prevent the issue in advance. Also, it has no built-in mechanism for authentication. We can extend it or combine with the image pull secret solution but it sounds too complicated and dirty.
 
 
-### Containerd Configuration
+#### 1.c Containerd Configuration
 
 As we do in vintage clusters, we can configure containerd configuration as below. This seems the best approach.
 
@@ -73,25 +74,102 @@ username = "giantswarm-user-account"
 password = "my_token_from_docker_hub"
 ```
 
-## Registry usage
+#### Decision
 
-### Which registry should we use? 
+`1.c` is selected.
+We will use containerd configuration. 
+
+### 2. Authentication
+
+#### 2.a Using unauthenticated accounts
+
+Using only public images without authentication is an option to avoid complexity of securing credentials. However, registry providers have pull/rate limits per account. We will be limited by total of [free account limit](https://docs.docker.com/docker-hub/download-rate-limit/) of Docker Hub (100 pulls per 6 hours per IP address) and [tier limit](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-skus) of Azure Container Registry (3000 ReadOps per minute for standard). Anyone can pull images from our registry and spend our limits, which makes us vulnerable. 
+
+#### 2.b Using authenticated accounts
+
+We can pass credentials to `containerd` configuration to be not limited by registry provider limits.
+
+#### Decision
+
+`2.b` is selected.
+We will use authenticated accounts.
+
+
+### 3. Registries
 
 The infranet page (See `Registry Mirrors`) states this as a desired configuration:
   - `docker.io` as primary public registry since it has a privilege in docker daemon.
   - Our own Azure Container Registry as secondary registry because of the security concerns
 
-Using only public images without authentication is also on the table. In that case, we will be limited by total of [free account limit](https://docs.docker.com/docker-hub/download-rate-limit/) of Docker Hub (100 pulls per 6 hours per IP address) and [tier limit](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-skus) of Azure Container Registry (3000 ReadOps per minute for standard). However, anyone can pull images from our registry and spend our limits, which makes us vulnerable. That is why we must use authenticated users.  
+Since we use `containerd`, docker has not a privelege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
 
-### Open Points
+### 4. Accounts
 
-## How are we going to configure containerd in cluster app?
+- 4.a Giantswarm Account + Single account
+- 4.b Giantswarm Account + Per MC including its WCs
+- 4.c Giantswarm Account + Per WC
+- 4.b Customer Accounts
 
-We need to provide a configation interface in our `cluster-$provider` apps for the users so that they can provide their credentials. Note that we are the users of our management clusters.
+#### Decision
 
-Options:
-1. We can ask users to create whole containerd configuration as a secret.
-2. We can ask users to provide list of registries and credentials as a secret and render containerd configuration with them.
+`4.b` is selected.
+
+Since it is a part of the platform itself, we are going to use GiantSwarm accounts in containerd configuration.
+As we do for other operators, we will follow "per MC" approach here to.
+
+### 5. How to provide credentials to WCs
+
+#### 5.a cluster-apps-operator
+
+We can propogate credentials from MC to WC by using `cluster-apps-operator`. It can create a secret per WC with the credentials.
+
+#### 5.b Optional interface + Gitops
+
+We can delegate the responsibility of creating credentials secret to the creator of `cluster-$provider` apps. GitOps can be useful to template the secrets per WC.
+
+#### Decision
+
+`5.b` is selected.
+
+We don't want to add another responsility to `cluster-apps-operator`. 
+
+
+### 6. Default configuration for WCs
+
+- 6.a: Configuring WCs by default
+- 6.b: Not configuring WCs by default
+
+#### Decision
+
+**NOT DECIDED YET**
+
+
+### 7. Where to put the credentials
+
+#### 7.a Customer's Git Repository
+
+We can put the secret into customers' git repositories. Customers will access to our credentials.
+
+#### 7.b Management Cluster Fleet
+
+We can put the secret into our `management-clusters-fleet` repo. We will need to give reference to that secret in customers' git repository. Management of these secrets can be so costly when we need to create the secret in each organization namespace.
+
+#### Decision
+
+**NOT DECIDED YET**
+
+
+### 8. Configuration interface
+
+How will be the configuration interface in `cluster-$provider` apps?
+
+#### 8.a Full containerd configuration
+
+We can implement a full transitive configuration interface so that users can provide a full containerd configuration, including registy credentials.
+
+#### 8.b Only registry credentials in a structred way
+
+We can define a configuration interface like below and render containerd configuration in `cluster-$provider` app chart.
 
 ```
 registries:
@@ -104,20 +182,8 @@ registries:
   ...
 ```
 
-> Note that there can be more than one feature we will support that affects containerd configuration like proxy settings. In this case, we should ensure that the option 1 will not lead conflicts.
+#### Decision
 
-### Workload Clusters
+`8.b` is selected.
 
-- Will we configure workload clusters by default?
-- Will we use our credentials for workload clusters by default?
-
-According to the discussion in Kaas Sync, we will configure management clusters and all workload clusters in the management cluster will inherit the configuration from their MCs.
-
-### Configuration Details
-
-- Where will be the source of truth for the container registry configuration? `config` repo or `management-cluster-fleet`?
-- How will WCs consume the credentials in their charts? 
-
-### How are going to manage credentials per cluster?
-
-Will we create a user per management cluster?
+8.a seems risky. Also, there can be other configurations (e.g. proxy) that touch containerd configuration too. 8.a can be error prone.
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -108,7 +108,7 @@ Since we use `containerd`, docker has not a privilege anymore but there is no ne
 - 4.a Giantswarm Account + Single account
 - 4.b Giantswarm Account + Per MC including its WCs
 - 4.c Giantswarm Account + Per WC
-- 4.b Customer Accounts
+- 4.d Customer Accounts
 
 #### Decision
 
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -18,7 +18,7 @@ explicit k8s.gcr.io 12
 
 - We don't configure our Kubernetes clusters for container registries via a central configuration. We mostly rely on public registries and anonymous users. Some applications consume `.dockerConfigJson` from config repo but it is for only `quay.io`.
 
-- We have an infranet page (See `Registry Mirrors`) that states some decisions about which registries we must use. Shortly,
+- We have an intranet page (See `Registry Mirrors`) that states some decisions about which registries we must use. Shortly,
   - We need registry mirrors for high availability.
   - We will not run our own registy.
   - We will use `docker.io` as primary public registry since it has a privilege in docker daemon.
@@ -97,11 +97,11 @@ We will use authenticated accounts.
 
 ### 3. Registries
 
-The infranet page (See `Registry Mirrors`) states this as a desired configuration:
+The intranet page (See `Registry Mirrors`) states this as a desired configuration:
   - `docker.io` as primary public registry since it has a privilege in docker daemon.
   - Our own Azure Container Registry as secondary registry because of the security concerns
 
-Since we use `containerd`, docker has not a privelege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
+Since we use `containerd`, docker has not a privilege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
 
 ### 4. Accounts
 
@@ -114,7 +114,7 @@ Since we use `containerd`, docker has not a privelege anymore but there is no ne
 
 `4.b` is selected.
 
-Since it is a part of the platform itself, we are going to use GiantSwarm accounts in containerd configuration.
+Since it is a part of the platform itself, we are going to use Giant Swarm accounts in containerd configuration.
 As we do for other operators, we will follow "per MC" approach here to.
 
 ### 5. How to provide credentials to WCs
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -140,7 +140,7 @@ We can put the secret into customers' git repositories and use GitOps templates 
 
 #### 6.c Management Cluster Fleet
 
-We can put the secret into our `management-clusters-fleet` repo. We will need to give reference to that secret in customers' git repository.
+We can put the secret into our `management-clusters-fleet` repo and refer that secret in `extraConfigs` in cluster apps. We can use GitOps templates to inject this configuration in all cluster apps.
 
 #### 6.d Catalog configuration
 
@@ -148,12 +148,11 @@ We can use catalog configurations (See <mc-name>/appcatalog folder in `installat
 
 #### Decision
 
-`6.d` is selected.
+`6.c` is selected.
 
 - Regarding 6.a, we don't want to add another responsility to `cluster-apps-operator`. 
 - Regarding 6.b, we want to manage the credentials in our side and to have a full control over them.
-- Regarding 6.c, management of the secrets can be so costly when we need to create the secret in each organization namespace.
-- Regarding 6.d, it is not the ideal solution we want to have since catalog configurations are hard to track. Nevertheless, we decided to start with this. Since the configurations are passed by the app platform, we can change this in the future without changing anything in cluster-$provider apps and users' inputs.
+- Regarding 6.d, people think catalog configurations are hard to track since there is no references in app CRs.
 
 ### 7. Configuration interface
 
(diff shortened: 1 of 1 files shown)


erkanerol:
@gianfranco-l I learned that this is also related to Harbor. @MarcelMue recommended me to ask help from @giantswarm/team-honeybadger to complete this work. Can you help me?

gianfranco-l:
we @giantswarm/team-honeybadger will discuss this during planning tomorrow and we'll get back to you

piontec:
@kubasobon this is really close to our discussions about repositories layout, but from a csutomer/client point of view. Can you PTAL here?

gianfranco-l:
@giantswarm/team-honeybadger will have a look at this and review it :)

kubasobon (commented):
I concur that `Containerd Configuration` looks like the most promising solution. What are your plans for open points? Discussing in a SIG, or a meeting?

container-registry-configuration/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Open Points
+
+## How are we going to configure containerd in cluster app?
+
+We need to provide a configation interface in our `cluster-$provider` apps for the users so that they can provide their credentials. Note that we are the users of our management clusters.
+
+Options:
+1. We can ask users to create whole containerd configuration as a secret.
+2. We can ask users to provide list of registries and credentials as a secret and render containerd configuration with them.
```
AverageMarcus, line 94: Why would we do this rather than making use of `imagePullSecrets` where the customer then has complete control over what uses their creds?

erkanerol: We need credentials for control-plane elements too. We need to put them into container configuration for static pods.  Should we use separate 2 credentials for control-plane&user workloads? 

Note: I am going to update this RFC soon according to the results of yesterday's kaas sync.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+### 4. Accounts
+
+- 4.a Giantswarm Account + Single account
+- 4.b Giantswarm Account + Per MC including its WCs
+- 4.c Giantswarm Account + Per WC
+- 4.b Customer Accounts
+
+#### Decision
+
+`4.b` is selected.
```
AverageMarcus, line 115: When was this decided? My understanding was we were going to use the same approach we do now with vintage and use out own read-only credentials.

Customers can then add their own using `imagePullSecrets`

erkanerol: That is my understanding from the discussions. I may be wrong.

What do you plan to do 
- for static pods of WCs like api-server?
- for our non-static pods in WCs like csi-controller-plugin?

AverageMarcus: I still think we should be setting the containerd config but it should be static using our credentials. No need to expose it to customer. There's no requests for it and no reason to expose a potential way for customers to destroy their cluster.

erkanerol: This section is not related to whether customers will see our credentials or not. Section 7 is related to that topic. 

Regarding this section, when we put the credentials to containerd configuration of WC nodes, it will be used for all pods of WC. Are you OK with it?

Are you OK with "per MC" approach in account usage? There will be like `giantswarm-hyperion` account and it will be used in hyperion and it's all WCs.

AverageMarcus: That's a change from how we do it in vintage.

In vintage we have a single set of credentials used _everywhere_ to handle authenticated pulls. Those credentials don't have access to anything secret and don't have write permission. Those credentials are baked into the node images and not available to be configured by the customer anywhere.

I thought for this specific issue (avoiding rate limiting) we would be following the same approach. 

I don't see any need to over complicate things with multiple credentials, multiple levels of config, out-of-sync config due to releases not being deployed everywhere yet, etc. It'll make debugging harder for (in my opinion) no benefit.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+Docker Hub has [download rate limit](https://docs.docker.com/docker-hub/download-rate-limit/) and it is pretty strict. To use Docker Hub without authentication is not a livable option for us. We started to hit the limit in big CAPI clusters during upgrades.
+
+## Current status
+
+- We use images from different registries. Here is the summary of registry usage of a test MC.
+```
+explicit docker.io  40
+implicit docker.io  2
+explicit quay.io    26
+explicit k8s.gcr.io 12
```
pipo02mix, line 16: is this per node basis or whole cluster? (some images will be the same)

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+registries:
+- name: docker
+  endpoint: https://registry-1.docker.io
+  username:
+  password:
+- name: azurecr
+  endpoint: giantswarm.azurecr.io
+  ...
+```
```
pipo02mix, line 183: Even (you decide)[https://github.com/giantswarm/rfc/pull/49/files#r1027860367] use the same credentials for all accounts I would still allow customers to add their own mirrors. It is a requirement by Panamax that we implemented in KVM via k8s initiator app

container-registry-configuration/README.md
erkanerol: It is for the whole MC cluster. Images are unique.

container-registry-configuration/README.md
erkanerol: Do you propose embedding credentials to node images in the image builder? If we use the same credentials everywhere, a customer's usage will affect other customers since there are still limits for premium accounts. Shouldn't we isolate tenants?

AverageMarcus: Any idea how vintage avoids that today? Is it an issue we're already seeing?

container-registry-configuration/README.md
JosephSalisbury: as a driveby note, it'd be good to retag any images from `k8s.gcr.io` to our own registries

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+explicit docker.io  40
+implicit docker.io  2
+explicit quay.io    26
+explicit k8s.gcr.io 12
+```
+
+- We don't configure our Kubernetes clusters for container registries via a central configuration. We mostly rely on public registries and anonymous users. Some applications consume `.dockerConfigJson` from config repo but it is for only `quay.io`.
+
+- We have an infranet page (See `Registry Mirrors`) that states some decisions about which registries we must use. Shortly,
```
JosephSalisbury, line 21:
```suggestion
- We have an intranet page (See `Registry Mirrors`) that states some decisions about which registries we must use. Shortly,
```

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+#### Decision
+
+`2.b` is selected.
+We will use authenticated accounts.
+
+
+### 3. Registries
+
+The infranet page (See `Registry Mirrors`) states this as a desired configuration:
```
JosephSalisbury, line 100:
```suggestion
The intranet page (See `Registry Mirrors`) states this as a desired configuration:
```

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+  - `docker.io` as primary public registry since it has a privilege in docker daemon.
+  - Our own Azure Container Registry as secondary registry because of the security concerns
+
+Since we use `containerd`, docker has not a privelege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
```
JosephSalisbury, line 104:
```suggestion
Since we use `containerd`, docker has not a privilege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
```

okay cool, ignore above

container-registry-configuration/README.md
JosephSalisbury: we currently don't isolate between tenants. this is possibly something we should start doing

we'd have to investigate the docker account setup a bit more, i forget the details here

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+- 4.a Giantswarm Account + Single account
+- 4.b Giantswarm Account + Per MC including its WCs
+- 4.c Giantswarm Account + Per WC
+- 4.b Customer Accounts
+
+#### Decision
+
+`4.b` is selected.
+
+Since it is a part of the platform itself, we are going to use GiantSwarm accounts in containerd configuration.
```
JosephSalisbury, line 117:
```suggestion
Since it is a part of the platform itself, we are going to use Giant Swarm accounts in containerd configuration.
```

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+#### Decision
+
+**NOT DECIDED YET**
+
+
+### 7. Where to put the credentials
+
+#### 7.a Customer's Git Repository
+
+We can put the secret into customers' git repositories. Customers will access to our credentials.
```
JosephSalisbury, line 151: As a note, customers technically have access to these secrets now in vintage

container-registry-configuration/README.md
JosephSalisbury: iirc, there's some flexibility here, i doubt docker care if we create a bunch of accounts and link them all to our plan

JosephSalisbury (commented):
drive by, not blocking, love ya <3

container-registry-configuration/README.md
bavarianbidi: As long as we don't disallow the usage of `hostPath` mounts (or `kubectl debug node/<nodeName>`) customers are always able to see all sensitive information on the host filesystem.

container-registry-configuration/README.md
puja108: > as a driveby note, it'd be good to retag any images from k8s.gcr.io to our own registries

also because they won't work in China, we had to do that already

container-registry-configuration/README.md
puja108: let's not mix the baking into node topics with the per MC or not per MC topic.

on the account topic: we currently use one account for all MCs, for Docker it would be the same if we stay that way vs multiple accounts. only benefit of per MC accounts would be that if they get expensive we can make those costs more transparent to the customers, which might be cool, ideally customer should not really re-use our accounts to be honest.

mid term: we should have harbor in the MC and cache/proxy all images through there, which would also significantly reduce requests.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+### 6. Default configuration for WCs
+
+- 6.a: Configuring WCs by default
+- 6.b: Not configuring WCs by default
+
+#### Decision
+
+**NOT DECIDED YET**
```
puja108, line 144: if we do not configure them by default we create unstable/unreliable clusters. In vintage we configure everything be default AFAIK. What would speak against that?

container-registry-configuration/README.md
puja108: still jsut for rotation purposes alone, having it outside our jurisdiction feels wrong, why would be put one of our secrets into sth owned by a customer? no matter the security implications

container-registry-configuration/README.md
erkanerol: To be able to do it by default, we need to create the credential secret in every organization.

erkanerol:
After all the discussions and comments above, my understanding is that 

- We should configure all WCs by default.
- We shouldn't put our secrets into customers' git repos.
- We shouldn't embed the secret creation logic into cluster-apps-operator. We should use GitOps.

I don't know how can we do it. We need to create the secret in each org to be able to give a reference in KubeAdmConfig. We cannot manage those secrets in management-cluster-fleet  for each org in customers' repo. 

@puja108 @kopiczko @alex-dabija  is there a way to do all these at the same time? 

Maybe we should use an operator to create the secret and let users give the reference in the app configuration. 
Or we should embed the credentials into node images as Marcus proposed above.

container-registry-configuration/README.md
AverageMarcus: We don't need to bake them into the image, they can be part of the cluster app and pulling a secret from the config repo (or wherever)

container-registry-configuration/README.md
AverageMarcus: Just a note: `k8s.gcr.io` is the old kubernetes registry and shouldn't be used anymore anyway.

bavarianbidi:
> Maybe we should use an operator to create the secret and let users give the reference in the app configuration. Or we should embed the credentials into node images as Marcus proposed above.

Maybe CAPIs runtime hooks might be a solution for that.
If the `Cluster CR` has a hook configured, CAPI stops reconciliation till the hook has done his job (e.g. create a secret for the specific cluster).
I've described a possible usage in [Improving Workload Cluster bootstrap process #24742](https://github.com/giantswarm/giantswarm/issues/24742#issuecomment-1324667451)

erkanerol:
I just noticed we can refer to a secret in another namespace in App CRs. Therefore, we can follow this mechanism:
1. Create a common secret in MC, managed by management-cluster-fleet
2. While creating clusters in customers' Git repos, give a reference to that secret
```
    secret:
      name: "container-credentials"
      namespace: "giantswarm"
```
3. Consume the structured (e.g. array or map) credentials in helm templates and render the containerd configuration as a secret in the same namespace
4. Refer to that secret in KubeAdmConfg.

That sounds reasonable to me now.

container-registry-configuration/README.md
cornelius-keller: I was just today in a meeting with Panamax where they said that it is a requirement in the future that all images are pulled through their own registry mirror (artifactory).

AverageMarcus: Requirement in the future is totally fine. That's a function for Harbour.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+  - Our own Azure Container Registry as secondary registry because of the security concerns
+
+Since we use `containerd`, docker has not a privilege anymore but there is no need to change this order at the moment. We will follow this configuration for CAPI clusters too.
+
+### 4. Accounts
+
+- 4.a Giantswarm Account + Single account
+- 4.b Giantswarm Account + Per MC including its WCs
+- 4.c Giantswarm Account + Per WC
+- 4.b Customer Accounts
```
alex-dabija, line 111: This needs to be 4.d instead of 4.b .

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+The purpose of this RFC is to specify how we configure Kubernetes clusters for container registries.
+
+## Problem Statement
+
+Docker Hub has [download rate limit](https://docs.docker.com/docker-hub/download-rate-limit/) and it is pretty strict. To use Docker Hub without authentication is not a livable option for us. We started to hit the limit in big CAPI clusters during upgrades.
+
+## Current status
+
+- We use images from different registries. Here is the summary of registry usage of a test MC.
```
AndiDog, line 11: I want to mention _pull through caching_ since it's not listed in this PR.

AWS ECR, as the only larger cloud registry, [supports](https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache.html) it already. When Docker rate limits became enforced, I experienced the same problems in large business Kubernetes installations. It was worked around by

- [AWS public ECR images](https://gallery.ecr.aws/) wherever available (major ones like Ubuntu, Flux, etc. seem available) because AWS yet has no rate limit or at least a per-account limit that doesn't totally kill productivity (example: `MY_ACCOUNT_ID.dkr.ecr.eu-central-1.amazonaws.com/ecr-public/docker/library/ubuntu:20.04` instead of `ubuntu:20.04`)
- rewriting images after enabling the pass-through cache (replace `docker.io/...` with AWS ECR image, for instance, which is quite some effort if not automated)

It's a _valid workaround_ to buy time and avoid live issues. I fully agree to buying into container registries such as Harbor which can work around the limitations from a centrally-configured place.

Anecdotal, minor story: replacing `image: abc` with `image: def` seems not uncommon. In addition to the above case, when I tried IPv6-only clusters for my private purposes, I noticed that lots of registries (e.g. `ghcr.io`) and websites (e.g. GitHub release binaries) aren't IPv6-reachable and I therefore patched a few of my Helm chart installations, for example like `redis: { image: { repository: registry.ipv6.docker.com/library/redis } }`. A container registry should have support for such patches.

container-registry-configuration/README.md
alex-dabija: I think we should get into the habit of isolating the MCs from each other (especially with the push for CAPA & CAPZ private clusters) and this includes having separate secrets per MC, even for the registry. I know in this case it's not a big deal because we don't have anything secret in the registry, but a decision like `MCs are isolated from each other and centralized features are an exception` would have avoided the conversation entirely.

I also think the registry and its configuration needs to be externalized (not backed in an image or hard-coded) in order to keep our product flexible. This would allow us to quickly respond to customer requests, debug problems & test features.

container-registry-configuration/README.md
alex-dabija: I think the clusters need to have the registry secret configured by default for the same reason Puja mentioned. The propagation of secrets across multiple namespaces could be handled by something like the ClusterSecret CR. There's an operator for this: https://github.com/zakkg3/ClusterSecret (but, I don't know how reliable it is).

puja108: As discussed in last KaaS sync and with my background as having been part of the team designing the app interface, I'd really suggest to use the systems that are in place in the interface that honeybadger is providing us with the app platform. An MC-wide default is supposed to go into the catalog config/secret, it can be overridden by cluster with the cluster secret (and/or user secret), but that should be an optional override if needed.

I know there was visibility concerns on this but that is feedback for honeybadger that you can bring there to discuss.

Building operators and workaround to get functionality that is already there sounds like busy work to me. We can improve on what we have, but let's use it first to see if it creates issues and then we can see how to solve those issues. Visibility is a UX issue that does not put the solution in question IMHO.

container-registry-configuration/README.md
alex-dabija: It's our registry, which implies that we should be in control of the secret to rotate it.

container-registry-configuration/README.md
alex-dabija: I didn't mean that we need to build a new operator (there is already one). I would just like to have the behavior of the ClusterSecret CR I mentioned. It's great if the app platform is able to do it and that we don't have to do anything else (no new code, no additional deployments).

In the end, I would like to see the following:
- default registry config (secret + url) per MC;
- this default is made available to all WCs (probably the responsibility of the app platform);
- `cluster-$provider` chart has default reference to the registry config (secret + url);
- `cluster-$provider` has the option to have a different registry config passed in (secret + url);
- `cluster-$provider` renders the `containerd` config;

The registry config can be a list of registries.

The `cluster-$provider` should also have the ability to have additional registries configured by the customer (appended to our defaults). Probably not needed now, but the design should not prevent it.
I can also see that in some private environments customers would want to have full control over the list of registries, but this requires more discussions.

Why? The defaults are needed in order to offer a good user experience out of the box. The configurability is needed to enable us to respond to customer requests and to allow us the flexibility needed in testing.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,189 @@
...
+#### 8.b Only registry credentials in a structred way
+
+We can define a configuration interface like below and render containerd configuration in `cluster-$provider` app chart.
+
+```
+registries:
+- name: docker
+  endpoint: https://registry-1.docker.io
+  username:
+  password:
```
alex-dabija, line 179: Are these properties passed in via the `values.yaml` file to the `cluster-$provider` chart? It's problamatic if that's the case and I think that a reference to a Kubernetes secrets needs to be used instead.

erkanerol: We will render a k8s secret and then give a reference in CAPI CRs. Example:  https://github.com/giantswarm/cluster-openstack/pull/123

As far as I see, it is already implemented in CAPA. https://github.com/giantswarm/cluster-aws/pull/160

container-registry-configuration/README.md
webwurst: The new one is `registry.k8s.io`

container-registry-configuration/README.md
uvegla: I am all for managing them ourselves.

container-registry-configuration/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+We can put the secret into our `management-clusters-fleet` repo. We will need to give reference to that secret in customers' git repository. Management of these secrets can be so costly when we need to create the secret in each organization namespace.
+
+#### 7.c Catalog configuration
+
+We can use catalog configurations (See <mc-name>/appcatalog folder in `installations` repo) to provide a MC specific configuration to all charts in the cluster.
+
+#### Decision
+
+7.c is selected.
```
piontec, line 162: Important comment about this solution. Ii totally makes sense short term, but in a long term we might (just might - nothing's set in stone) deprecate support for catalog configs. Why? 
* they are supper-hidden currently; there's no indication in the App CR that a config from a not very related Catalog CR is being used. Yes, we can indicate this in App's status, but still:
* from an unaware user's perspective, they work like magic, in the unreliable, hard to debug way: you install two apps, one gets the config the other doesn't - why? (they come from different catalogs). You move and app between catalogs (or just stop using the `-test` catalog and stuff breaks, even though app's config is exactly the same

In the long run, I feel it would be much better to provide a different source of cross-app config, like a CM/Secret name `cluster-default` (just an idea) that is included in every app's `extraConfigs` by default, with low priority.

Let me know what you think.

erkanerol: We picked the catalog configs since we don't have any other mechanism now. Personally, I am not happy with the solution. Now, I am trying to implement it but I need to use `opctl encrypt` and `opsctl decrypt` for catalog configuration in the installations repo, which will be something new for us. 

Regarding `cluster-default` secret, it should be created in every org namespace automatically. There is no operator for it yet.

erkanerol:
@alex-dabija Reference implementation is here https://github.com/giantswarm/cluster-cloud-director/pull/79

container-registry-configuration/README.md
```diff
@@ -0,0 +1,185 @@
...
+We can use catalog configurations (See <mc-name>/appcatalog folder in `installations` repo) to provide a MC specific configuration to all charts in the cluster.
+
+#### Decision
+
+`6.d` is selected.
+
+- Regarding 6.a, we don't want to add another responsility to `cluster-apps-operator`. 
+- Regarding 6.b, we want to manage the credentials in our side and to have a full control over them.
+- Regarding 6.c, management of the secrets can be so costly when we need to create the secret in each organization namespace.
+- Regarding 6.d, it is not the ideal solution we want to have since catalog configurations are hard to track. Nevertheless, we decided to start with this. Since the configurations are passed by the app platform, we can change this in the future without changing anything in cluster-$provider apps and users' inputs.
```
fiunchinho, line 156: I think this is not a good idea. The `installations` repo is something we should stop using, and it should've been removed by now. Adding sops secret to it means having yet another repository with sops secrets. I feel the situation with the `installations` repo and `cluster-apps-operator` is similar, in the sense that both are things we want to stop using. I think using `cluster-apps-operator` is similar to using the `installations` repo, but better, because at least it has been updated recently, and it's something many people know it exists (there may be people that don't even know about the `installations` repo).

But I wouldn't use either. I think `6.c` is the approach that it's more aligned with what we want. We can create the secret from `management-clusters-fleet` and reference it from the `App` CR `extraConfigs` field. This field allows to reference secrets in a different namespace, so we wouldn't need to copy/sync this secret between namespaces/organizations. Are there any downsides of following this approach?

erkanerol: Note that this is my PR regarding this https://github.com/giantswarm/mc-bootstrap/pull/404

@giantswarm/team-honeybadger Jose said cross-namespace secret reference is possible. Is it a feature or a bug?

fiunchinho: I haven't tried it, but that's what I understand from the docs

image

erkanerol: There won't be any SOPS related stuff in installation repository. There will be only an additional reference to a secret in app catalog. Example: https://github.com/giantswarm/installations/blob/gloria_auto_branch/gloria/appcatalog/cluster-appcatalog-values.yaml#L10

If @giantswarm/team-honeybadger is OK to use this way (extraConfigs + cross-namespace reference), I can improve the PR in mc-bootstrap.

piontec: Yes, this should work, and that's what I was advertising from the start: in the future, we want all the config sources to go to this list of `extraConfigs` and don't have any "mysterious" sources, like config injected suddenly from a Catalog. I like this option better as well,but to improve usability, there should be a logic that can set some config entries in `extraConfigs` automatically, ie. if there's a ConfigMap named 'ClusterDefault*', it is always set in the list - we don't have it yet.

fiunchinho: @piontec can you elaborate on what you mean by "logic that can set some config entries"? Would that be required to set the container registry configuration? As I can see it, we can set it directly in the `cluster-$provider` apps.

erkanerol: > I like this option better as well,but to improve usability, there should be a logic that can set some config entries in `extraConfigs` automatically

Do you mean we+customers have to set `extraConfigs` section in every cluster app? If it is so, it is not a big deal. We use GitOps to template cluster based on a base. 

I was not aware of that there is already a solution in Honeybadger's portfolio for cross-namespace secret reference. I learned it from Jose. If you suggest this solution, I am OK to implement this. Pawel will be happier :D

erkanerol: I talked with @piontec again and he confirmed this way. I updated the RFC and I am updating my PR.

Huge thanks for contributing the effort.

erkanerol: Up-to-date PR https://github.com/giantswarm/mc-bootstrap/pull/429

erkanerol:
In the end, we achieved a consensus and implemented initial versions for CAPVCD and CAPA. Merging this RFC.

PR #50, opened 2022-11-10
document current proxy implementation

bavarianbidi:


proxy-support/README.md
```diff
@@ -0,0 +1,272 @@
...
+  - [Tasks](#tasks)
+  - [PRs](#prs)
+<!-- /toc -->
+
+## Glossary
+
+- Management Cluster (`MC`): 
+- Workload Cluster (`WC`): 
+- Installation: An installation defines the combination of a `MC` with all the provisioned `WCs` from there.
+- App Platform:
```
pipo02mix, line 51: Here you missed the acronym

proxy-support/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+### workload on top of Kubernetes
+
+Every pod within a Kubernetes cluster needs a valid proxy configuration if the pod needs to access external systems.
+
+On a `MC` there are a lot of components which needs access to external systems:
+
+- `flux` needs access to `github.com`
+- the entire monitoring and alerting stack needs access to `opsgenie`, `grafana`, ...
+- App platform components needs access to an OCI artifact system (e.g. `github.com`, `azure`)
```
pipo02mix, line 89:
```suggestion
- `Flux` needs access to `github.com`
- The entire monitoring and alerting stack needs access to `opsgenie`, `grafana`, ...
- App platform components need access to an OCI artifact system (e.g. `github.com`, `azure`)
```

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+        key: containerdProxy   
+```
+
+By doing this, `ClusterAPI` is waiting for the `secret` to get created. Once the `secret` got created, it will be used by creating the machines (and injecting this data via `cloud-init` or `ignition`).
+
+As `cluster-apps-operator` takes care of creating `deployment/chart-operator` in the WC with the values from `(secret|configmap)/<clusterName>-cluster-values` as input, `chart-operator` in every WC automatically get a valid proxy configuration.
+
+## To be discussed
+
+- Customers are able to see the proxy configuration (especially the secret part) within a `WC`. Is this a security issue?
```
bavarianbidi, line 230: @giantswarm/team-shield i'm interested in your opinion about this

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+- Should the [`cluster-apps-operator`] known installation wide proxy configuration propagated into WCs for [`kyverno-policies-connectivity`]?
+- Should a customer defined Proxy configuration used for `containerd`, `cert-manager` and `chart-operator` as well or only for customer workload?
+
+## Current known limitations and issues:
+
+- If customers using a service mesh (e.g. `linkerd`) and the mesh is spawn over multiple WCs, the current implementation provides that all mesh specific member values are defined in the global proxy configuration.
+- If a customer want a specific proxy configuration for a workload cluster, this proxy configuration must be currently defined in multiple places.
```
bavarianbidi, line 237: At the moment it's technically possible to overwrite the proxyConfiguration which is required for the initial cluster bootstraping (proxy configuration in `containerd` and `certManager`) but this has to be done on multiple places (`secret` for `containerd` must be created, `values.yaml` for `certManager` must be configured)

We could think about extending `cluster-apps-operator` by propagating a custom proxy configuration to the workload clusters if the workload cluster specific configuration got created in a specific way (tbd e.g. a `secret` or `configmap` with a fixed naming pattern) beforehand.

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+- The entire monitoring and alerting stack needs access to `opsgenie`, `grafana`, ...
+- App platform components need access to an OCI artifact system (e.g. `github.com`, `azure`)
+
+As we already use `kyverno` in our `MC`, we can use the [`kyverno-policies-connectivity`] chart to inject the proxy configuration into all pods.
+
+> Please note: 
+> - due multiple valid reasons, `kyverno` doesn't mutate objects in `namespace/kube-system` per default.
+> - static pods are managed by the `kubelet`. For these pods it's not possible to use `kyverno` for mutation. Therefore it's required to inject the proxy configuration during the machine creation phase (could be achieved in `CAPI` based clusters via `kubeadm`)
+
+#### special workload
```
erkanerol, line 97: I think this should be H3 header. It is parent of the ones below.

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+> Note: the interpretation of the `no_proxy` variable is very language/tool specific and not every language supports e.g. network ranges.
+> a very details blog post was written by GitLab about [proxy variables].
+
+## Implementation
+
+### Management Cluster
+
+#### during the initial bootstrap
+
+During the initial bootstrap phase of the `MC`, there isn't the entire App Platform running. For that reason it's required to define the proxy configuration multiple times.
```
erkanerol, line 131: Can list the components here? I know it is not a static list and we may change it in the future but it may be good to know in this phase.

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### after all controllers/operators are in place
+
+After the initial bootstrap is done and all GiantSwarm specific controllers and operators are in place to treat the `MC` as a `WC`, [`cluster-apps-operator`] will take over the handling of the installation specific proxy configuration.
+
+### Workload Clusters
+
+The current implemented workflow can be visualized like this:
+
+![](data/workload_cluster_creation.svg)
```
erkanerol, line 141: This is my understanding too.

proxy-support/README.md
```diff
@@ -0,0 +1,273 @@
...
+  API:
+    clusterIPRange: 10.96.0.0/12
+  DNS:
+    IP: 10.96.0.10
+proxy:
+  noProxy: cluster.local,100.64.0.0/13,100.96.0.0/11,178.170.32.59,172.16.0.0/16,internal.corpdomain.net,example.com,svc,127.0.0.1,localhost
+```
+
+[`cluster-apps-operator`] also creates a `secret` which contains some infrastructure specific values ([for `openstack` and `vcd`](https://github.com/giantswarm/cluster-apps-operator/blob/master/service/controller/resource/clustersecret/desired.go#L42-L65)).
+The `secret` for each WC is called `<clusterName>-cluster-values`.
```
erkanerol, line 171: Did we split the proxy configuration into 2 places? I think we can put the `noProxy` values into the secret too. Noone will use the noProxy without others and it is not a problem to put a non-sensitive data into a secret. As far as I understand, it will not require any change in other components, only in cluster-apps-operator.

bavarianbidi: yes, currently the configuration is split into `secret/<clustername>-cluster-values` and `configmap/<clustername>-cluster-values`.

Also the configuration for `cluster-apps-operator` is currently done in the `config`-repo in the files:
- `installations/<INSTALLATION_NAME>/apps/cluster-apps-operator/configmap-values.yaml.patch`
- `installations/<INSTALLATION_NAME>/apps/cluster-apps-operator/secret-values.yaml.patch`

bavarianbidi: I've updated the `cluster-apps-operator` PR to safe the `noProxy` configuration in the `secret/<clustername>-cluster-values`. the `configmap/<clustername>-cluster-values` will be unchanged (compared to the current released version of `cluster-apps-operator`)

proxy-support/README.md
bavarianbidi: as the objects below are H5 i think H4 is correct.

proxy-support/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+### workload on top of Kubernetes
+
+Every pod within a Kubernetes cluster needs a valid proxy configuration if the pod needs to access external systems.
+
+On a `MC` there are a lot of components which needs access to external systems:
+
+- `Flux` needs access to `github.com`
+- The entire monitoring and alerting stack needs access to `opsgenie`, `grafana`, ...
+- App platform components need access to an OCI artifact system (e.g. `github.com`, `azure`)
```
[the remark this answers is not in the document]
bavarianbidi: > Dex needs access to the identity provider

Dex is getting deployed via the app-platform and is running in `namespace/giantswarm` when the kyverno policy is already in place.

proxy-support/README.md
```diff
@@ -0,0 +1,267 @@
...
+> - static pods are managed by the `kubelet`. For these pods it's not possible to use `kyverno` for mutation. Therefore it's required to inject the proxy configuration during the machine creation phase (could be achieved in `CAPI` based clusters via `kubeadm`)
+
+#### special workload
+
+There are some components, which are part of the default set on top of a cluster. These components can't get configured via `kyverno` as they are either in `namespace/kube-system` (for `cert-manager`) or must exist before we apply applications via the app platform (for `chart-operator`).
+
+##### `cert-manager`
+
+As `cert-manager` is needed in every cluster (`MC` and `WC`) and it's running in `namespace/kube-system`. `cert-manager` interact with external `ACME` services (`lets encrypt` at the moment).
+The proxy configuration must be explicitly defined in the `pod.spec`.
```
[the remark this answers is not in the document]
bavarianbidi: `kyverno` create certificates by their own and doesn't depend on `cert-manager`.

kopiczko:
I would like to rise my concern about using `cluster-apps-operator` to create and inject proxy secret. IMO this is unfortunate this operator exists at all and we should reduce the it's usage to the bare minimum as it injects values to all charts using cluster values. By increasing the number of values in cluster values CM/Secret we are creating the same issues we had with `installations` repo, i.e. values which we don't know where and how they are used and are scared to ever remove them because we don't know the consequences. Cluster values is also pretty magical concept. It's really difficult to realise they are in use making whole platform more indirect, implicit and difficult to understand.

I would prefer customers to inject proxy values to every single WC that is supposed to run behind a proxy. I understand that may lead to problems when clusters are not created correctly because a customer forgot about proxy configuration. But so they can for a number of different reasons. We can also have a validation for that if this becomes annoying. Besides that it better fits GitOps style of configuration.

Configuring the MC itself is not a problem as we are starting to manage `cluster-<provider>` chart in `management-clusters-fleet` repository.

proxy-support/README.md
puja108: the weird thing to me is that the proxy does not seem to differentiate between proxy admins (people who should see config) and proxy users (people who just need to use the proxy). I guess login they want so that no one else can use the proxy, who should maybe not have external access? otherwise it would feel wrong

proxy-support/README.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+By doing this, `ClusterAPI` is waiting for the `secret` to get created. Once the `secret` got created, it will be used by creating the machines (and injecting this data via `cloud-init` or `ignition`).
+
+As `cluster-apps-operator` takes care of creating `deployment/chart-operator` in the WC with the values from `(secret|configmap)/<clusterName>-cluster-values` as input, `chart-operator` in every WC automatically get a valid proxy configuration.
+
+## To be discussed
+
+- Customers are able to see the proxy configuration (especially the secret part) within a `WC`. Is this a security issue?
+- Should the [`cluster-apps-operator`] known installation wide proxy configuration propagated into WCs for [`kyverno-policies-connectivity`]?
```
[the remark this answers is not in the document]
puja108: see also discussion today in KaaS sync

bavarianbidi:
> I would like to rise my concern about using `cluster-apps-operator` to create and inject proxy secret. IMO this is unfortunate this operator exists at all and we should reduce the it's usage to the bare minimum as it injects values to all charts using cluster values. [...]

I didn't know that the existing of `cluster-apps-operator` is not wanted at all. To me it sounds that we have to elaborate why `cluster-apps-operator` exists and why we should evaluate alternative implementations. As `cluster-apps-operator` belongs to honeybadger but we (as provider teams) aren't happy with the existence of this component i would propose to create at least an issue for team honeybadger to discuss the future of this component. WDYT @kopiczko? (i can take over the issue creation but i need your help in the discussion).

bavarianbidi:
> I would prefer customers to inject proxy values to every single WC that is supposed to run behind a proxy. I understand that may lead to problems when clusters are not created correctly [...]

IMO both use-cases are valid. Some customers don't want to take care of any proxy configuration and maybe some others would define the proxy configuration in different components very different.
But beside the general configuration (proxy server + username/password) the `noProxy` value is something which is very cluster specific and not all values are known during the initial cluster creation (e.g. the k8s apiserver address). And therefore i've decided to compute most of these values as from a customer point of view (maybe) not every customer want to take care of these.

But i guess, we didn't know our customers very well (especially for this use-case).

proxy-support/README.md
stone-z: Shield discussed this ([thread](https://gigantic.slack.com/archives/C02FL4EAADD/p1668625800984809)).

In general, we think it's necessary to expose proxy information to workloads which need it, so it has to work this way right now. We also aren't limited by this approach in the future, since using Kyverno policies it would also be possible to refine which workloads should be mutated to include the proxy information (if, for example, some pods are not trusted with that information, or if different proxy information should be injected into different pods).

As Puja mentioned, there is one caveat: I understood that any credentials contained in the proxy variables were required for a workload to use a proxy (i.e. the proxy requires authentication). However, Puja understood it differently, and if that is NOT the case, and the credentials are somehow "admin" credentials for actually administering the proxy, then we should probably skip straight to splitting the policies and only injecting those specific credentials where needed.

bavarianbidi: The current implementation doesn't fullfil this need, if the proxy users shouldn't see any kind of proxy configuration. 

If this is something which should be abstracted for the users and they shouldn't see this configuration we have to think about some `transparent proxy` solutions (by implementing some iptables rules + a installation specific proxy server which act as "man in the middle" with the only purpose of having corporate proxy credentials stored).

stone-z: > The current implementation doesn't fullfil this need, if the proxy users shouldn't see any kind of proxy configuration.

Right, I meant if a workload did not need to use the proxy and should not see the proxy info, not that it needed to use the proxy but was not trusted. I agree that would be a separate problem

kopiczko:
> I didn't know that the existing of `cluster-apps-operator` is not wanted at all.

This is only my opinion.

puja108:
As to the `cluster-apps-operator` topic: This is more of an architecture/engineering topic. We want to keep teams independent and create good interfaces between them. AFAIK the operator was intended as a way for Honeybadger to bootstrap their app platform (similar to what PMO is to Atlas), it should not be an interface where others add business logic. The interface is app CRs and the way they can be configured with CMs and Secrets (and Flux functionality in a way now). This way the team has the freedom to change and even get rid of its implementation without other teams having to move their functionality to other places and blocking the team.

proxy-support/README.md
```diff
@@ -0,0 +1,267 @@
...
+> - due multiple valid reasons, `kyverno` doesn't mutate objects in `namespace/kube-system` per default.
+> - static pods are managed by the `kubelet`. For these pods it's not possible to use `kyverno` for mutation. Therefore it's required to inject the proxy configuration during the machine creation phase (could be achieved in `CAPI` based clusters via `kubeadm`)
+
+#### special workload
+
+There are some components, which are part of the default set on top of a cluster. These components can't get configured via `kyverno` as they are either in `namespace/kube-system` (for `cert-manager`) or must exist before we apply applications via the app platform (for `chart-operator`).
+
+##### `cert-manager`
+
+As `cert-manager` is needed in every cluster (`MC` and `WC`) and it's running in `namespace/kube-system`. `cert-manager` interact with external `ACME` services (`lets encrypt` at the moment).
```
QuentinBisson, line 103: Did kyverno not move to the kyverno namespace? Is there still a reason preventing it from running in the kube-system namespace?

stone-z: Kyverno has moved and is now running in the `kyverno` namespace by default. 

However, to support cases where customers run very small, single-master clusters, product decided to exclude `kube-system` by default so that the cluster still can recover from the loss of all nodes. The intent is currently to re-enable Kyverno in `kube-system` once we've found a way to still support recovery from total node loss. Here is an issue from the implementation which references the product discussion https://github.com/giantswarm/giantswarm/issues/24707

QuentinBisson: That is sad news :( Do we have some documentation explaining how to configure it in the pod spec then?

stone-z: Configure the proxy? Or configure Kyverno to also apply in `kube-system`?

QuentinBisson: Sorry, I meant the proxy :)

bavarianbidi: will create an issue and prepare an initial documentation

calvix (approved):

kubasobon (approved):
No blocking remarks from me

kopiczko:
I still don't love we use `cluster-apps-operator` for injecting proxy values. We could take the same approach we took with containerd configuration which is setting a secret from `giantswarm` ns in `App.spec.extraConfigs` for each cluster App. I understand if someone "forgets" to set it cluster won't work.

mcharriere:
hi there, I'm working on https://github.com/giantswarm/giantswarm/issues/25235 and we are considering deprecating the `proxy` values in favor of the `env` value directly. 

My first instinct was to set the necessary bits in the `default-apps-*`, basically translating from one structure to the other. But unfortunately the `proxy` configuration is not available to the default-app. 

I'm wondering if it would be possible to inject the proxy values into the default-app and if so, how?
IMO the benefits of this approaches are:
 - do not rely on the `cluster-values` CM for default apps (which circles back to the `cluster-apps-operator` discussion). 
 - be explicit about proxy configuration
 - the interface between the provider and the app lives in the configmap configuring the app, not inside the app

Regarding the implementation, I'm not sure how the customers are creating the default-apps CR or if we _could_ have some sort of `installation` CM that is added to every default-app automatically. any ideas?

bavarianbidi:
In favor of https://github.com/giantswarm/giantswarm/pull/27149 (to document the current status) i think about closing this PR soon.

For some of the open discussions github issues got created.
Will collect them all and try to handover most of the things to sig architecture (when it make sense)

erkanerol (approved):
I have read carefully and spent some time to think what we can add more. I couldn't find anything. Looks good to me.

bavarianbidi:
For all open questions/opinions, dedicated follow up issues got created:

- [ ] https://github.com/giantswarm/giantswarm/issues/26812 (+ [comment](https://github.com/giantswarm/giantswarm/issues/26812#issuecomment-1591231332))
- [ ] https://github.com/giantswarm/giantswarm/issues/24784
- [ ] https://github.com/giantswarm/roadmap/issues/1854 (+ [comment](https://github.com/giantswarm/roadmap/issues/1854#issuecomment-1591241058))
- [ ] https://github.com/giantswarm/roadmap/issues/1853 (+ [comment](https://github.com/giantswarm/roadmap/issues/1853#issuecomment-1591127917))

As the handbook also contains the current [proxy documentation](https://handbook.giantswarm.io/docs/product/connectivity/proxy/) i will close this RFC for now.

As all follow up improvements are addressed and they can be discussed in multiple different formats/SIG i will close this RFC.

PR #52, opened 2022-11-11
Update automatic upgrades rfc with CAPI news

pipo02mix:
Towards https://github.com/giantswarm/thg/issues/113

### DIFF
--- a/automatic-cluster-upgrades/README.md
+++ b/automatic-cluster-upgrades/README.md
@@ -8,7 +8,7 @@
 
 - As a Giant Swarm customer, and also as a Giant Swarm maintainer, we would like to have workload clusters running in the latest version automatically, and not worry about patching security vulnerabilities or cluster components running out of updates (e. g. Kubernetes release end of life).
 
-- As a customer admin, I would like to define upgrade windows, to ensure that upgrades happen when disruption is most permittable and impact on production traffic is minimal.
+- As a customer admin, I would like to define upgrade windows, to ensure that upgrades happen when disruption is most suitable and impact on production traffic is minimal.
 
 - As a customer admin, I would like to mark a workload cluster as "frozen" in order to block upgrades for a certain time period.
 
@@ -18,6 +18,8 @@
 
 - As a customer, and also as Giant Swarm, we want to capture the result of an upgrade. Based on the result we can alert on a certain condition, track the history for certain cluster(s) or automate response based on the end state.
 
+- As a customer, and also as Giant Swarm, we want to make the automatic upgrades compatible to GitOps, so the state of the infrastructure changes are tracked in Git.
+
 ## Reasoning
 
 Our customers need to plan every upgrade on their side and manage the upgrades for every workload cluster release. Each customer has their own requirements regarding when they need to schedule upgrades (freeze periods, maintenance windows, etc.). At the same time, cloud native projects and Giant Swarm are continuously providing new releases. The preparation and execution of upgrades is time and labor intensive, while the fact that they are executed manually does not add much value to stakeholders.
@@ -26,9 +28,9 @@ Giant Swarm designed a versioning scheme (based on Semver), which includes all c
 
 ## Current state
 
-Right now the automatic upgrades are carried by Solution Engineers agreeing with the customer on upgrade windows and environments selected. The idea is to automate all this behaviour and information under a Kubernetes operator(s) and Custom Resource(s).
+Right now the automatic upgrades are carried by Solution Engineers (AEs and SAs) agreeing with the customer on upgrade windows and environments selected. The idea is to automate all this behaviour and information under a Kubernetes operator(s) and Custom Resource(s).
 
-The Solution Engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on the cluster version field on the cluster Custom Resource (CR).
+The Solution Engineer plans cluster upgrades based on cluster type (environment, team,...) and maintenance windows offered by the customers. The upgrade is just a change on cluster-<provider> App manifest of the cluster.
 
 Based on the fact Giant Swarm is moving to Cluster API this is story let us define a new greenfield scenario where we can design a model that can be more generic and not focus only in Giant Swarm platform. The idea would be contribute as much as possible with upstream due the fact most of this functionality would be useful for everyone that manages clusters.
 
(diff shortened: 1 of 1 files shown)


automatic-cluster-upgrades/README.md
```diff
@@ -122,22 +128,18 @@ This would allow upgrades to happen between 07:00 UTC and 17:00 UTC on Monday-Fr
 
 #### Cluster upgrade executor
 
-There will be an operator that based on the `Cluster Upgrade` CRs will initiate the upgrades changing the labels on the specific CR(s), it could include Apps at some point though ideally would detach apps totally from the cluster and apps metadata and App Platform enforce the requirements. 
+There will be an operator that based on the `Cluster Upgrade` CRs will trigger the upgrades changing the labels on the specific CR(s) when date matches or it will create the pull requests in Git repo to trigger the upgrade.
```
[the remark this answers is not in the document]
pipo02mix: yeah that is just a proposal. I am happy the team that will implement this modify based on experience or requirements. Your approach sounds good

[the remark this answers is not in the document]
pipo02mix: I do think this label of abstraction let us not tie the cluster upgrade automation to GitOps and leave the option to implement a different executor that changes cluster CR fields directly or something else

automatic-cluster-upgrades/README.md
```diff
@@ -72,25 +74,29 @@ Example resource:
...
-    kind: Cluster
-    name: foo01
+  matchers:
+  - name: cluster_id
+    value: mgab6
+    isRegex: false
+  - name: dev_clusters
+    key: stage
+    value: dev
+    isRegex: true
```
pipo02mix, line 87: @piontec moving the discussion here. This is a proposal but I got your point of making this object tracked by Git so the GitOps workflow is not broken. 

Two things:

1) We could change `ClusterUpgrade` object to be tied to a single cluster (1:1 relation). That way the `status` would reflect the result of the upgrade and can help us to track upgrades, and at same time can be added to Git repo as part of GitOps workflow

2) The Cluster Upgrade controller instead of creating `ClusterUpgrade` resource in MCs can do PRs to your Git repo in case of GitOps enabled. Otherwise can default to MC API

mproffitt: I like the simplicity of a 1:1 relationship between `ClusterUpgrade` object and `Cluster` however I also feel there is scope for requirements gathering on this point - how do customers *want* to implement upgrades? Is a 1:1 mapping desired or would they want to upgrade by `Organization` - I think there needs to be a balance here between simplicity and desirability here.

mproffitt: Irrespective of `GitOps` process, I also think that we need to track the current deployed version for an individual cluster, especially so in larger customers where many versions might be in play at any one time. Tracking of deployed versions would be important in the event of DR or cluster rebuilds where deploying an upgraded version might have detrimental impact on customers ability to operate their workloads.

pipo02mix: > I like the simplicity of a 1:1 relationship between ClusterUpgrade object and Cluster however I also feel there is scope for requirements gathering on this point - how do customers want to implement upgrades? Is a 1:1 mapping desired or would they want to upgrade by Organization - I think there needs to be a balance here between simplicity and desirability.

I do think it can be tackled with higher abstraction. For example, in GitOps we will charge a base and it will run all attached clusters, and it is defined by the customer 💪 . In other scenarios, we can rely on a higher entity but having the `ClusterUpgrade` entity tied to a single cluster will help in most cases IMO

> Irrespective of GitOps process, I also think that we need to track the current deployed version for an individual cluster, especially so in larger customers where many versions might be in play at any one time. Tracking of deployed versions would be important in the event of DR or cluster rebuilds where deploying an upgraded version might have detrimental impact on customers ability to operate their workloads.

But the versions (k8s/OS/CAPI) are defined in Cluster API CRs, so we will not lose them. Having anyway a `ClusterUpgrade` for each upgrade done can help in tracking changes sure (we can use `status` field) but we can find another approach and store upgrades in an independent DB in case we find using MC etcd is too much.

mproffitt (approved):

piontec:
The problem I see with the proposed approach and CR is that it is very strict and inflexible: it strictly specifies which cluster, exactly when, to which specific version should be upgraded. This means that I can't configure an upgrade "policy": a set of rules about how to upgrade my clusters instead of a specific instruction about how to do a single upgrade. 

Let's try to define a policy of upgrades that also allows tracking a specific version we're running right now. Modelling a bit after flux'es auto upgrades, WDYT about this:

```yaml
apiVersion: 
kind: ClusterUpgrade
metadata:
  name: upgrade-to-cluster-openstack-v1.20
  namespace: default
spec:
  targetClusterSelector:
  - name: cluster_id
    value: mgab6
  policy:
    release:
      apiVersion: application.giantswarm.io/v1alpha1
      kind: Release
      semver:
        regex: 1\.2\..*-rc.*   # exclusive with below
        range: >=1.0.0 <2.0.0
    timeWindow:
      cron: 5 3 * * * *
  deployedVersion: 1.2.1
status:
  triggeredAt: 2021-03-10T10:00:00Z
  state: finished
  outcome: success
  finishedAt: 2021-03-10T10:31:10Z
  fromRelease: v1.16.0
  toRelease: v1.20.0
```

The approach above allows to "define a policy and forget about it". Still, it also requires some means of discovering of new versions matching the criteria are available.

piontec:
> `ClusterPolicy` would be the one created to define the policy used for cluster-scheduler-controller to generate `ClusterUpgrades`

I like this idea. Just keep in mind it has a relatioship to how gitops works. In general, we want the state in git to fully describe the current configuration, so even if we have upgrade policies, we still need to know what exact version is deployed right now (think about disaster recovery).

This means we have to either:
- make a commit to the repo every time a `ClusterUpgrade` is created and completed/deleted
- include the current version in `ClusterUpgradePolicy` and commit it with a new version once upgrade is completed.

I think I like the 2nd approach better. Still, fi it works like this, do we need `ClusterPolicy` at all? Maybe just logs will be enough?

pipo02mix:
Yeah the idea suggested [here](https://github.com/giantswarm/rfc/pull/52/files#diff-15f56470ef844d066b992b78d06419f5da8df81465023d8b52a2bbc9960d5135R131) was to have different flavours of `cluster-executor-controller` based on customer preferences, so in case of GitOps instead of changes directly on cluster  CRs, it does the PRs to right repo

pipo02mix:
I am merging this now with the hope on a future team will take when implementation/design has to be done

PR #53, opened 2022-11-16
Labelling workload cluster's apps

gusevda:
The purpose of this RFC is to propose a labelling scheme to safely identify a workload cluster's defining apps.

Towards https://github.com/giantswarm/roadmap/issues/1599, https://github.com/giantswarm/roadmap/issues/1559.

### DIFF
--- a/labelling-wc-apps/README.md
+++ b/labelling-wc-apps/README.md
@@ -5,11 +5,11 @@ The purpose of this RFC is to propose a labelling scheme to safely identify a wo
 This is required for:
 
 - deleting a workload cluster on CAPI;
-- displaying a workload cluster's defining apps in Happa.
+- displaying a workload cluster's defining apps in a client like the web UI (happa).
 
 ## Deleting a workload cluster on CAPI
 
-To delete a workload cluster on CAPI we need to identify corresponding cluster app.
+To delete a workload cluster on CAPI, we need to identify the corresponding cluster app.
 
 ### Current state
 
@@ -17,15 +17,15 @@ On CAPI we are relying on the resource names (in the org namespace) for cluster 
 
 ### Proposal
 
-Require `giantswarm.io/cluster` to be present in all app resources related to a workload cluster, including cluster app.
+Require the `giantswarm.io/cluster` label to be present in all App resources related to a workload cluster, including the cluster app.
 
-Introduce a new `giantswarm.io/cluster-app` label to safly mark cluster app CRs.
+Introduce a new `giantswarm.io/cluster-app` label to safely mark cluster app resources (the App CR).
 
 Refactor places where it's assumed that cluster app doesn't have `giantswarm.io/cluster` label to rely on a new `giantswarm.io/cluster-app` label instead.
 
 ## Displaying a workload cluster's defining apps in Happa
 
-In Happa, we display workload cluster's apps in two separate lists - apps installed by a user and default (preinstalled) apps. To achieve that, we need:
+In Happa, we intend to display a workload cluster's apps in two separate lists - apps installed by a user and default (preinstalled) apps. To achieve that, we need:
 
 - fetch all App CRs related to a workload cluster. On CAPI this list should not contain cluster app App CR.
 - split the App CRs into default apps and user installed apps.
@@ -49,6 +49,6 @@ On CAPI, proposal from the "Deleting a workload cluster on CAPI" will also suit 
 
 #### For splitting cluster's apps into two lists:
 
-We could eather introduce a new label to mark default apps bundle e.g. `giantswarm.io/installed-by-default`, or we can set `giantswarm.io/managed-by` label of a default apps bundle to something specific (right now it's not set).
+We could either introduce a new label to mark default apps bundle, e.g. `giantswarm.io/installed-by-default`, or we can set the `giantswarm.io/managed-by` label of a default apps bundle to something specific (right now it's not set).
 
-Require that `giantswarm.io/managed-by` label is correctly set for app bundle's child apps. It should point to a parent app.
+Require that `giantswarm.io/managed-by` label is correctly set for app bundle's child apps. It should point to the parent app.
--- a/labelling-wc-apps/README.md
+++ b/labelling-wc-apps/README.md
@@ -1,0 +1,54 @@
+# Labelling workload cluster's apps
+
+The purpose of this RFC is to propose a labelling scheme to safely identify a workload cluster's defining apps.
+
+This is required for:
+
+- deleting a workload cluster on CAPI;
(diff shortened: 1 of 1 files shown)


labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
+# Labelling workload cluster's apps
+
+The purpose of this RFC is to propose a labelling scheme to safely identify a workload cluster's defining apps.
+
+This is required for:
+
+- deleting a workload cluster on CAPI;
+- displaying a workload cluster's defining apps in Happa.
```
marians, line 8:
```suggestion
- displaying a workload cluster's defining apps in a client like the web UI (happa).
```

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Deleting a workload cluster on CAPI
+
+To delete a workload cluster on CAPI we need to identify corresponding cluster app.
```
marians, line 12:
```suggestion
To delete a workload cluster on CAPI, we need to identify the corresponding cluster app.
```

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Current state
+
+On CAPI we are relying on the resource names (in the org namespace) for cluster app selection which is not recommended. There is no reference between cluster CR and cluster app CR - unlike other cluster's app resources, cluster app CR doesn't have `giantswarm.io/cluster` label. It cannot be set at the moment, since it will break cluster deletion.
+
+### Proposal
+
+Require `giantswarm.io/cluster` to be present in all app resources related to a workload cluster, including cluster app.
```
marians, line 20:
```suggestion
Require the `giantswarm.io/cluster` label to be present in all App resources related to a workload cluster, including the cluster app.
```

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Introduce a new `giantswarm.io/cluster-app` label to safly mark cluster app CRs.
+
+Refactor places where it's assumed that cluster app doesn't have `giantswarm.io/cluster` label to rely on a new `giantswarm.io/cluster-app` label instead.
+
+## Displaying a workload cluster's defining apps in Happa
+
+In Happa, we display workload cluster's apps in two separate lists - apps installed by a user and default (preinstalled) apps. To achieve that, we need:
+
+- fetch all App CRs related to a workload cluster. On CAPI this list should not contain cluster app App CR.
```
marians, line 30: Why should this list not include the cluster app?

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### For cluster's apps fetching:
+
+On vintage, let's make `giantswarm.io/cluster` label to be required for all app resources related to a workload cluster.
+
+On CAPI, proposal from the "Deleting a workload cluster on CAPI" will also suit this use case. We will fetch all cluster's app resources relying on the `giantswarm.io/cluster` label and the `giantswarm.io/cluster-app` label can be used to filter out cluster app from a list.
+
+#### For splitting cluster's apps into two lists:
+
+We could eather introduce a new label to mark default apps bundle e.g. `giantswarm.io/installed-by-default`, or we can set `giantswarm.io/managed-by` label of a default apps bundle to something specific (right now it's not set).
```
marians, line 52:
```suggestion
We could either introduce a new label to mark default apps bundle, e.g. `giantswarm.io/installed-by-default`, or we can set the `giantswarm.io/managed-by` label of a default apps bundle to something specific (right now it's not set).
```

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Require that `giantswarm.io/managed-by` label is correctly set for app bundle's child apps. It should point to a parent app.
```
marians, line 54:
```suggestion
Require that `giantswarm.io/managed-by` label is correctly set for app bundle's child apps. It should point to the parent app.
```

marians (approved):
Cosmetic changes suggested. Content wise I have nothing to add/change.

labelling-wc-apps/README.md
gusevda: I should rephrase it. It's more like 
> Currently we don't display cluster app in the UI, so we should either fetch all the apps except cluster app, or we need a way to filter it out later.

marians: Now I get it. Yep, would be good to clarify that.

vxav:
It should be enough to add a label in the `default-app-<provider>` charts to identify default apps.
e.g. https://github.com/giantswarm/default-apps-cloud-director/blob/main/helm/default-apps-cloud-director/templates/_helpers.tpl#L19

labelling-wc-apps/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+To delete a workload cluster on CAPI, we need to identify the corresponding cluster app.
+
+### Current state
+
+On CAPI we are relying on the resource names (in the org namespace) for cluster app selection which is not recommended. There is no reference between cluster CR and cluster app CR - unlike other cluster's app resources, cluster app CR doesn't have `giantswarm.io/cluster` label. It cannot be set at the moment, since it will break cluster deletion.
+
+### Proposal
+
+Require the `giantswarm.io/cluster` label to be present in all App resources related to a workload cluster, including the cluster app.
```
fiunchinho, line 20: Just to add some context, I believe that we don't add the label to the cluster-app because `cluster-apps-operator` adds a finalizer to the Cluster CR once it starts reconciling a cluster, which is removed from the Cluster CR once all the apps labeled `giantswarm.io/cluster` are deleted. But the cluster-app won't be removed until the Cluster CR is removed. And that won't happen until the finalizer is removed. And that won't happen until the cluster-app is removed. So we have a chicken - egg kind of problem.

How do you propose to rewrite the deletion logic so that it works after adding the `giantswarm.io/cluster` label?

gusevda: I think we can be more explicit about not deleting the cluster-app. At the moment it kinda happens as a side effect, it's assumed that you have a knowledge that the cluster-app doesn't have a `giantswarm.io/cluster` label. Maybe instead of selecting the apps labeled with `giantswarm.io/cluster` label, we can select the apps labeled with `giantswarm.io/cluster` and that don't have `application.giantswarm.io/cluster-app` label which I suggest to add.

kubasobon (approved):

fiunchinho:
What's the current state of this RFC? What do we do with it?

gusevda:
> What's the current state of this RFC? What do we do with it?

I think we need to discuss the priority of it in the Honeybadger team, I will raise the discussion.

PR #54, opened 2022-11-24
New RFC for CAPI Kubernetes/OS version management 

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/23341#issuecomment-1315766253

### DIFF
--- a/capi-matrix/README.md
+++ b/capi-matrix/README.md
@@ -2,27 +2,29 @@
 
 ## Introduction
 
-We have a cluster-<provider> chart for every provider that we support. It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
+We have a cluster-<provider> chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
 
-In the values of these charts we define the Kubernetes version of the cluster and the image that machine will use to run. These images contain the Kubernetes components with a specific version. 
+In the values of these charts, among all possible parameters we define the Kubernetes version of the cluster and the image that machines will use to run. These images contain the Kubernetes components with a specific version. 
 
-One of the goal in this RFC is decouple Kubernetes version from image names so customer does not need to know which image correspond to which version. At the same time we might want to provide new OS version when new CVE appears without necessarily change the Kubernetes version. Finally having a list of kubernetes version certified as ready to be used benefit to disallow users to create cluster in flawy versions.
+This RFC born out of necessity to automate the cluster upgrades. For that reason team rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
 
 ## User stories
 
-- As a customer, I want to not worry about the kubernetes version and rely on Giant Swarm default version provided.
+- As a customer, I don`t want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
 
-- As Giant Swarm staff, I don't want to expend time upgrading cluster manually.
+- As a customer, I would like to choose the kubernetes version of my cluster without think which OS AMI correspond to that version.
+
+- As Giant Swarm staff, I want to have the capacity to push new Kubernetes/OS versions and trigger the upgrade automation.
 
 - As Giant Swarm staff, I would like to forbid a specific Kubernetes/OS version to be used in case of security constraints.
 
-- As customer, I want to have the option to have a cluster pinned to a version.
+- As Giant Swarm staff, I would like to release a new OS version with the same Kubernetes version and roll all customer clusters when a security breach is detected.
 
 ## Solution
 
-- Lets have a cluster-<provider> chart release for every Kubernetes version is released. That way our automation can roll upgrades automatically when new version is published
+The solution proposed is to create a cluster-<provider> chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
 
-- Lets include a matrix between Kubernetes version and images names in the chart.
+But we would like to have a list of images attached to every Kubernetes version like:
 
 ```yaml
 kubernetesVersionImages:
@@ -33,7 +35,7 @@ kubernetesVersionImages:
   v1.23.5: "ubuntu-2004-kube-v1.23.5"
 ```
 
-This will help users to have to think about image names and I same time we can release a new OS version automatically like
+So the customers don't have to think about image names and we can remove Kubernetes/OS versions that are flawed. For example we can change OS version keeping the Kubernetes version the same:
 
 ```yaml
 kubernetesVersionImages:
@@ -41,10 +43,10 @@ kubernetesVersionImages:
   ...
 ```
 
-And customers upgrading to this new version will roll machines with the newer OS with same Kubernetes version.
+Every time there is a new Kubernetes version or OS version we will (automatically) create a PR to cluster-<provider> charts including the version and potentially changing the default version (if our tests run successfully).
 
-Every time there is a new K8s version or OS version we will (automatically) create a PR to cluster-<provider> charts including the version and potentially changing the default version (if our tests run successfully).
-
-We can also remove Kubernetes/OS version that we found are flawy or problematic in some way.
+At the same time if we found a Kubernetes/OS version with a security issue, we can remove the Kubernetes/OS version and enforce customers to upgrade clusters.
 
 We have start with [Cluster API OpenStack](https://github.com/giantswarm/cluster-openstack/pull/125).
+
+__Bonus__: If cluster-<provider> chart is used for companies out of Giant Swarm they can use their own matrix.
--- a/capi-matrix/README.md
+++ b/capi-matrix/README.md
@@ -1,16 +1,16 @@
-# RFC 0015 - CAPI matrix per provider
+# CAPI matrix per provider
 
 ## Introduction
 
-We have a cluster-<provider> chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
+We have a cluster-PROVIDER chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
 
 In the values of these charts, among all possible parameters we define the Kubernetes version of the cluster and the image that machines will use to run. These images contain the Kubernetes components with a specific version. 
 
-This RFC born out of necessity to automate the cluster upgrades. For that reason team rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
+This RFC born out of necessity to automate the cluster upgrades. For that reason, team Rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
 
 ## User stories
 
-- As a customer, I don`t want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
+- As a customer, I don't want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
 
 - As a customer, I would like to choose the kubernetes version of my cluster without think which OS AMI correspond to that version.
 
@@ -22,7 +22,7 @@ This RFC born out of necessity to automate the cluster upgrades. For that reason
 
 ## Solution
 
-The solution proposed is to create a cluster-<provider> chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
+The solution proposed is to create a cluster-PROVIDER chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
 
 But we would like to have a list of images attached to every Kubernetes version like:
 
@@ -43,10 +43,10 @@ kubernetesVersionImages:
   ...
 ```
 
-Every time there is a new Kubernetes version or OS version we will (automatically) create a PR to cluster-<provider> charts including the version and potentially changing the default version (if our tests run successfully).
+Every time there is a new Kubernetes version or OS version we will (automatically) create a PR to cluster-PROVIDER charts including the version and potentially changing the default version (if our tests run successfully).
 
 At the same time if we found a Kubernetes/OS version with a security issue, we can remove the Kubernetes/OS version and enforce customers to upgrade clusters.
 
 We have start with [Cluster API OpenStack](https://github.com/giantswarm/cluster-openstack/pull/125).
 
-__Bonus__: If cluster-<provider> chart is used for companies out of Giant Swarm they can use their own matrix.
+__Bonus__: If cluster-PROVIDER chart is used for companies out of Giant Swarm they can use their own matrix.
--- a/capi-matrix/README.md
+++ b/capi-matrix/README.md
@@ -10,9 +10,7 @@ This RFC born out of necessity to automate the cluster upgrades. For that reason
 
 ## User stories
 
-- As a customer, I don't want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
-
-- As a customer, I would like to choose the kubernetes version of my cluster without think which OS AMI correspond to that version.
+- As a customer, I would like to choose between the two supported minor Kubernetes versions without having to think about choosing an OS or machine image.
 
 - As Giant Swarm staff, I want to have the capacity to push new Kubernetes/OS versions and trigger the upgrade automation.
 
--- a/capi-matrix/README.md
+++ b/capi-matrix/README.md
@@ -1,0 +1,50 @@
+# RFC 0015 - CAPI matrix per provider
+
+## Introduction
+
+We have a cluster-<provider> chart for every provider that we support. It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
+
+In the values of these charts we define the Kubernetes version of the cluster and the image that machine will use to run. These images contain the Kubernetes components with a specific version. 
+
+One of the goal in this RFC is decouple Kubernetes version from image names so customer does not need to know which image correspond to which version. At the same time we might want to provide new OS version when new CVE appears without necessarily change the Kubernetes version. Finally having a list of kubernetes version certified as ready to be used benefit to disallow users to create cluster in flawy versions.
+
+## User stories
+
+- As a customer, I want to not worry about the kubernetes version and rely on Giant Swarm default version provided.
+
+- As Giant Swarm staff, I don't want to expend time upgrading cluster manually.
+
+- As Giant Swarm staff, I would like to forbid a specific Kubernetes/OS version to be used in case of security constraints.
+
+- As customer, I want to have the option to have a cluster pinned to a version.
+
+## Solution
+
+- Lets have a cluster-<provider> chart release for every Kubernetes version is released. That way our automation can roll upgrades automatically when new version is published
+
+- Lets include a matrix between Kubernetes version and images names in the chart.
+
+```yaml
+kubernetesVersionImages:
+  v1.22.9: "ubuntu-2004-kube-v1.22.9"
+  v1.22.10: "ubuntu-2004-kube-v1.22.10"
+  v1.23.3: "ubuntu-2004-kube-v1.23.3"
+  v1.23.4: "ubuntu-2004-kube-v1.23.4"
+  v1.23.5: "ubuntu-2004-kube-v1.23.5"
+```
+
+This will help users to have to think about image names and I same time we can release a new OS version automatically like
+
+```yaml
+kubernetesVersionImages:
+  v1.22.9: "ubuntu-2005-kube-v1.22.9"
+  ...
+```
+
+And customers upgrading to this new version will roll machines with the newer OS with same Kubernetes version.
+
+Every time there is a new K8s version or OS version we will (automatically) create a PR to cluster-<provider> charts including the version and potentially changing the default version (if our tests run successfully).
+
+We can also remove Kubernetes/OS version that we found are flawy or problematic in some way.
+
+We have start with [Cluster API OpenStack](https://github.com/giantswarm/cluster-openstack/pull/125).
--- a/capi-matrix/README.md
+++ b/capi-matrix/README.md
@@ -51,6 +51,22 @@ We have start with [Cluster API OpenStack](https://github.com/giantswarm/cluster
 
 __Bonus__: In favour of making the providers apps not tied to Giant Swarm, adding the conversion image table helps enabling other vendors to adopt it and customize it.
 
+## How the conversion image table is updated?
+
+- We have not defined this process. But one possible scenario would be:
+
+1) Kubernetes/Kinvolk release a new OS/k8s version
+
+2) Our image builder job is triggered and creates the AMIs for the different providers. 
+
+3) Other job can be triggered after and create pull requests to the cluster-PROVIDER repos adding the new k8s/OS version.
+
+4) Our future CI pipeline can test this new version for each provider and check it runs correctly (conformance tests or even custom tests). 
+
+5) In case CI succeed the default version and conversion image table is changed to the new one with the approval of the owning team.
+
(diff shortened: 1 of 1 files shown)


capi-matrix/README.md
```diff
@@ -0,0 +1,50 @@
+# RFC 0015 - CAPI matrix per provider
+
+## Introduction
+
+We have a cluster-<provider> chart for every provider that we support. It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
+
+In the values of these charts we define the Kubernetes version of the cluster and the image that machine will use to run. These images contain the Kubernetes components with a specific version. 
+
+One of the goal in this RFC is decouple Kubernetes version from image names so customer does not need to know which image correspond to which version. At the same time we might want to provide new OS version when new CVE appears without necessarily change the Kubernetes version. Finally having a list of kubernetes version certified as ready to be used benefit to disallow users to create cluster in flawy versions.
```
AverageMarcus, line 9: > flawy

Typo?

capi-matrix/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## User stories
+
+- As a customer, I want to not worry about the kubernetes version and rely on Giant Swarm default version provided.
+
+- As Giant Swarm staff, I don't want to expend time upgrading cluster manually.
+
+- As Giant Swarm staff, I would like to forbid a specific Kubernetes/OS version to be used in case of security constraints.
+
+- As customer, I want to have the option to have a cluster pinned to a version.
```
AverageMarcus, line 19: Can we include reasons here? 

And what boundaries? 
Will it be pin to a specific kubernetes/OS combo? Pinned to a patch? Pinned to a major?
Will we let them pin for 6 months? 1 year?

capi-matrix/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Solution
+
+- Lets have a cluster-<provider> chart release for every Kubernetes version is released. That way our automation can roll upgrades automatically when new version is published
```
AverageMarcus, line 23: I think this is the wrong approach.

I _still_ think we need an abstraction on top of cluster apps that the user deals with. That layer should include this kind of pinning.

capi-matrix/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Lets include a matrix between Kubernetes version and images names in the chart.
+
+```yaml
+kubernetesVersionImages:
```
AverageMarcus, line 28: Is this useful to the customer? How will that know what to pick? Why would we encourage the use of old (unsupported) Kubernetes versions?

AverageMarcus (commented):
Additional comments:

* How will OS patches be handled in this case?
* How will containerd updates be handled?
* How will changes to core/default applications be handled when they have Kubernetes version-specific changes? (e.g. Cilium, CoreDNS, etcd)
* What happens when our customers are running a version with a known CVE?

In general, I really don't like this idea. We seem to have completely abandoned the plan to have rolling updates which, when we started CAPI work, was billed as one of the key features.

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
+# RFC 0015 - CAPI matrix per provider
```
marians, line 1: We don't do numbering of RFCs, at least based on the main README.

```suggestion
# CAPI matrix per provider
```

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Introduction
+
+We have a cluster-<provider> chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation.
```
marians, line 5: "it is our UI to the customer" sounds like this is THE user interface, as if there are no others.

I guess you'd like to express something like: The cluster app is our technical interface to configure clusters.

In terms of terminology, I would replace "chart" with "app".

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
+# RFC 0015 - CAPI matrix per provider
+
+## Introduction
+
+We have a cluster-<provider> chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
+
+In the values of these charts, among all possible parameters we define the Kubernetes version of the cluster and the image that machines will use to run. These images contain the Kubernetes components with a specific version. 
+
+This RFC born out of necessity to automate the cluster upgrades. For that reason team rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
```
marians, line 9:
```suggestion
This RFC born out of necessity to automate the cluster upgrades. For that reason, team Rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
```

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## User stories
+
+- As a customer, I don`t want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
```
marians, line 13:
```suggestion
- As a customer, I don't want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
```

capi-matrix/README.md
marians, line 5: The term `<provider>` disappears when rendered as HTML. Changing the formatting.

```suggestion
We have a cluster-PROVIDER chart for every provider that we support (it is our UI to the customer). It has a set of templates that allow our customers to create a cluster app with all necessary resources required by a Cluster API implementation. 
```

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+- As Giant Swarm staff, I want to have the capacity to push new Kubernetes/OS versions and trigger the upgrade automation.
+
+- As Giant Swarm staff, I would like to forbid a specific Kubernetes/OS version to be used in case of security constraints.
+
+- As Giant Swarm staff, I would like to release a new OS version with the same Kubernetes version and roll all customer clusters when a security breach is detected.
+
+## Solution
+
+The solution proposed is to create a cluster-<provider> chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
```
marians, line 25:
```suggestion
The solution proposed is to create a cluster-PROVIDER chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
```

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+So the customers don't have to think about image names and we can remove Kubernetes/OS versions that are flawed. For example we can change OS version keeping the Kubernetes version the same:
+
+```yaml
+kubernetesVersionImages:
+  v1.22.9: "ubuntu-2005-kube-v1.22.9"
+  ...
+```
+
+Every time there is a new Kubernetes version or OS version we will (automatically) create a PR to cluster-<provider> charts including the version and potentially changing the default version (if our tests run successfully).
```
marians, line 46:
```suggestion
Every time there is a new Kubernetes version or OS version we will (automatically) create a PR to cluster-PROVIDER charts including the version and potentially changing the default version (if our tests run successfully).
```

capi-matrix/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+At the same time if we found a Kubernetes/OS version with a security issue, we can remove the Kubernetes/OS version and enforce customers to upgrade clusters.
+
+We have start with [Cluster API OpenStack](https://github.com/giantswarm/cluster-openstack/pull/125).
+
+__Bonus__: If cluster-<provider> chart is used for companies out of Giant Swarm they can use their own matrix.
```
marians, line 52:
```suggestion
__Bonus__: If cluster-PROVIDER chart is used for companies out of Giant Swarm they can use their own matrix.
```

marians:
For clarity: this is about the base image used for every node, control plane and worker, correct? I don't see this explained anywhere.

marians:
I don't understand the proposed solution yet.

One confusing bit may be the term "matrix". I don't see a matrix here. Is the term appropriate? For me, a matrix would mean a table, where one dimension is the OS version and the other one is the Kubernetes version, and each combination should create a valid result.

In CircleCI, for example, a [matrix job](https://circleci.com/docs/using-matrix-jobs/) means that all parameter values get combined when running a job.

```yaml
workflows:
  all-tests:
    jobs:
      - test:
          matrix:
            parameters:
              os: [linux, macos]
              node-version: ["14.17.6", "16.9.0"]
```

So `os=linux` is tested with both `node-version=14.17.6` and `node-version=16.9.0` etc.

pipo02mix:
In general, I really don't like this idea. We seem to have completely abandoned the plan to have rolling updates which, when we started CAPI work, was billed as one of the key features.

> How will OS patches be handled in this case?

In the same way, you push a new cluster-<provider> release with the new OS. 

For example, in the image matrix you move from
```
  v1.23.5: "ubuntu-2004-kube-v1.23.5"
```
to 
```
  v1.23.5: "ubuntu-2005-kube-v1.23.5"
```

>    How will containerd updates be handled?

I am not touching here. Just wanted to do a minimal change that unblocks automatic upgrades. Not sure if containerd is baked in the image or not. 

>    How will changes to core/default applications be handled when they have Kubernetes version-specific changes? (e.g. Cilium, CoreDNS, etcd)

Not in the scope of this RFC. RIght now we don't enforce a default app version and k8s version anyway

>     What happens when our customers are running a version with a known CVE?

Yeah this is the real reason of this RFC. WHen you have a flawed OS, let's say ubuntu-2004 and we will release ubuntu 2005 with the fix, then we will move from 
```
  v1.23.5: "ubuntu-2004-kube-v1.23.5"
```
to 
```
  v1.23.5: "ubuntu-2005-kube-v1.23.5"
```
And that way customers with automatic upgrades set will have the nodes rolled

pipo02mix:
> In general, I really don't like this idea. We seem to have completely abandoned the plan to have rolling updates which, when we started CAPI work, was billed as one of the key features.

This is not a very constructive comment. Since we started the CAPI implementation we create a WG to talk about upgrades and we produced [this RFC](https://github.com/giantswarm/rfc/tree/main/automatic-cluster-upgrades). Now we have created a [PR to update it based on CAPI learnings and GitOps](https://github.com/giantswarm/rfc/pull/52). The whole idea of this change is to make possible roll upgrades automatically. Can you explain to me what is stopping to do that?

capi-matrix/README.md
pipo02mix: sure

capi-matrix/README.md
pipo02mix: We proposed a meta app in the past and I think we discussed it in Kass sync. There were some drawbacks and it was discarded for now. What is the problem of using a kuberneteVersion default in this layer? BTW CAPG and CAPA is doing too

capi-matrix/README.md
pipo02mix: No this is not useful for the end customer (maybe there is a corner case) but maybe for other implementators like I describe at bottom of the doc. I have doubted to embedded in the _helpers but setting it values allow us to change it in real live if a critical issue appears

capi-matrix/README.md
pipo02mix: yeah good point. Adding it

pipo02mix:
> I don't understand the proposed solution yet.

> One confusing bit may be the term "matrix". I don't see a matrix here. Is the term appropriate? For me, a matrix would mean a table, where one dimension is the OS version and the other one is the Kubernetes version, and each combination should create a valid result.

Yeah it is not a matrix but a table or dictionary. It translates the Kubernetes version to the image AMI reference/name. I am explaining this better

AverageMarcus:
> Since we started the CAPI implementation we created a WG to talk about upgrades and we produced [this RFC](https://github.com/giantswarm/rfc/tree/main/automatic-cluster-upgrades?rgh-link-date=2022-11-25T11%3A59%3A25Z).

It seems I had completely missed this at the time and was completely unaware. Apologies.

---

The example you included in your comment of updating `v1.23.5: "ubuntu-2004-kube-v1.23.5"` to `v1.23.5: "ubuntu-2005-kube-v1.23.5"` isn't how ubuntu updates work. `ubuntu-2204` is the release name. Patches for CVEs are performed by updating apps in the image, the version of the OS doesn't change. The name would end up being the same unless we start timestamping them. (Which we could do, just makes things more difficult / error prone)

capi-matrix/README.md
pipo02mix: but this is not limiting or changing anything about your approach. It just makes easier the customer life, defaulting the image name based on the kubernetes version (at the same time allow us to control which versions are available to the customer)

pipo02mix:
> The example you included in your comment of updating v1.23.5: "ubuntu-2004-kube-v1.23.5" to v1.23.5: "ubuntu-2005-kube-v1.23.5" isn't how ubuntu updates work. ubuntu-2204 is the release name. Patches for CVEs are performed by updating apps in the image, the version of the OS doesn't change. The name would end up being the same unless we start timestamping them. (Which we could do, just makes things more difficult / error-prone)

But we are moving to Flatcar, right? I hope flatcar use different image name for each version. 

Anyway how we would do it currently with ubuntu if the image name does not change?

pipo02mix:
I have just updated the PR with more info and a new section describing how the table is updated. Thanks, Marian and Marcus for the comments. I have renamed matrix to table because it was confusing. In the end it is just a conversion table

AverageMarcus:
Currently we just build a new image for each new version of Kubernetes released. If we need to patch an old version we overwrite an existing AMI but that won't cause any running instances to be updated until they're replaced.

pipo02mix:
> Currently we just build a new image for each new version of Kubernetes released. If we need to patch an old version we overwrite an existing AMI but that won't cause any running instances to be updated until they're replaced.

I don't think it will change with this approach. If you keep the k8s -> image relation and not change the image name then end result will be the same (machines will not be rolled till they are not replaced). But this will be up to the team, we can decide to include in name a timestamp. I don't think here we enforce any behavior in relation of AMI building/naming

capi-matrix/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+From this moment Cluster Upgrade automation can start scheduling the workload cluster upgrades based on customer constraints. More info about this process discussed [here](https://github.com/giantswarm/rfc/pull/52).
+
+## Open questions
+
+- Every time we release a new provider app with a newer Kubernetes version we can make it default, but at the same time we might want to wait and run not edge version but a previous one. So we will need a bit of process deciding when we change the default version.
+
+## Out of scope
+
+- How containerd version is managed.
```
JosephSalisbury, line 76: if i understand correctly, we would manage containerd versions with this rfc, as they're tied to os versions

capi-matrix/README.md
```diff
@@ -0,0 +1,78 @@
...
+- As Giant Swarm staff, I would like to release a new OS version with the same Kubernetes version and roll all customer clusters when a security breach is detected.
+
+## Solution
+
+The solution proposed is to create a cluster-PROVIDER chart release for every Kubernetes/OS version is released. That way will help our automation to roll upgrades automatically when new version is published.
+
+But we would like to have a list of images attached to every Kubernetes version like:
+
+```yaml
+kubernetesVersionImages:
```
JosephSalisbury, line 30: dumb question: are these names provider specific? or would we have the same names across providers?

capi-matrix/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+In the values of these charts, among all possible parameters we define the Kubernetes version of the cluster and the machine image (AMI) that will be used to run the VMs. These images contain the Kubernetes components with a specific version. There are usually three AMIs defined in every provider: bastion, control-plane and node pools.
+
+This RFC born out of necessity to automate the cluster upgrades. For that reason, team Rocket has discussed and found a solution adding a Kubernetes and OS version in each chart release. Today AWS, GCP anf Open Stack providers pinned the Kubernetes/OS version in the chart. Here we propose an extension to give more flexibility and control to Giant Swarm and the customers.
+
+## User stories
+
+- As a customer, I don't want to worry about the kubernetes version or OS version used by my cluster machines. Just rely on the Giant Swarm default and tested version.
+
+- As a customer, I would like to choose the kubernetes version of my cluster without think which OS AMI correspond to that version.
```
puja108, line 15: this story directly opposes the first story above

capi-matrix/README.md
puja108: I still don't really get why there needs to be a list of versions, if every new OS/K8s version is anyway a new chart release/version. Wouldn't each chart release then only need a single image?

I thought we pin an OS/K8s combination per chart release. This way the upgrade is moving to the next chart version, which is something that is easy to automate also later based on upstream automation instead of having to rely on value fields.

Only exception I'd see and that I'd like to support is having different images for CP and worker nodes, cause sometimes we might just want to update the CP nodes (CVE in K8s CP) without having to roll all workers.

puja108 (commented):
On a more general note and in relation to Marcus' first comment:

We need to be careful with current iterations. The goal should still be a fully continuous rolling integration (I do not say release on purpose here) process. We already decided when we were discussing automatic upgrades last year, that we want a continuous release process with lots of minimal changes, so each update is just more or less an atomic change, no big releases anymore and no real releases towards the customer, but a promise of a continuous service that works and updates smoothly (longterm goal).
Also, from the ownership side, we will have huge issues next year pulling ownership of provider-independent things like K8s and OS, but also other smaller things, out of the current provider teams. I will soon present the ideas towards what we aim for with the future KaaS structure and we need to be careful not to build ourselves into a corner.

Not saying this proposal is against that, but let's evaluate it in such a light AND have it aligned between providers.

In the end an app/chart is our interface towards the customer and we need to be careful what we expose in that interface. From a product perspective we said that (for the beginning), we do not want to give all the freedom there is. We rather align with what we offer right now and customers are used to and happy with, with the exception of a few things where we already had customer requests and use cases (e.g. per cluster settings like OIDC).

cc @alex-dabija @cornelius-keller @gawertm @Rotfuks

capi-matrix/README.md
pipo02mix: then yes, but I meant here don't want to bring containerd version to the mix. I assume the containerd version will be compatible with the k8s components running on the OS

capi-matrix/README.md
pipo02mix: IIRC the names are produced by our image builder CI for each OS/k8s version published. Am I right @kopiczko ?

capi-matrix/README.md
pipo02mix: yeah it is on purpose. Unfortunately, not every customer has the same mindset. Some of them want not to worry at all and some want to have finer control. I can add as `advance` customer WDYT?

pipo02mix:
> We need to be careful with current iterations. The goal should still be a fully continuous rolling integration (I do not say release on purpose here) process. We already decided when we were discussing automatic upgrades last year, that we want a continuous release process with lots of minimal changes, so each update is just more or less an atomic change, no big releases anymore and no real releases towards the customer, but a promise of a continuous service that works and updates smoothly (longterm goal).

Yeah but underneath there will be releases, right? otherwise how we will change k8s or OS versions? We can hide the releases by building good automatic-upgrades automation so we can rely on it to release frequent and small changesets. 

I don't think the addition here is so drastic. IMO it is reasonable to have a k8s version conversation table to avoid the user (or us in case of providing the automation) knowing about the image name. BUT how this table is maintained and how we manage the automatic upgrades roll on top is where we need to put the effort into avoiding past mistakes.

My 2 cents: Lets discuss the automatic upgrades or upgrade process in general in different RFCs. Here we can focus on the proposed use cases and understand if this addition would have drawbacks rolling changes

capi-matrix/README.md
AverageMarcus: Correct, the names are chosen by us.

capi-matrix/README.md
puja108: My issue with this is that we have not decided to give customers such freedom for now. The current product officially supports 2 Kubernetes minor versions at each point in time. Patch versions are not something we want to give freedom on, I would also very much question the use case here, a patch version in K8s should only fix problems and CVEs, there should be nothing speaking against it.

Furthermore, even within the 2 versions our product offering is currently:
- new features and minor changes only get into the latest
- the older version only gets patch updates

On the OS side we do not even give that choice and I would also like to keep that product decision unless there's a concrete use cases that makes sense here.

Exceptions to the above I know of, but I'd like to keep those to exceptions and not make them a feature.

TLDR I would reduce it to `As a customer, I would like to choose between the two supported minor Kubernetes versions without having to think about choosing an OS or machine image.`

capi-matrix/README.md
pipo02mix: But our idea is to moved Flatcar in all providers, right?

capi-matrix/README.md
pipo02mix, line 15:
```suggestion
- As a customer, I would like to choose between the two supported minor Kubernetes versions without having to think about choosing an OS or machine image.
```

capi-matrix/README.md
teemow: What Marcus is saying is that there should be just one version available. So neither the customer should be able to pin any version nor we provide any choice. I'd like to get there too. 

But how would we solve freeze periods then? Do we have confidence that we won't break anything and everything is just kept up to date whatsoever? Or do we still need to be able to stay at a certain version? And will we be able to test the upgrade that happens after the freeze period? This might be a bigger jump.

capi-matrix/README.md
teemow: This would make the lifecycle of kubernetes independent from other features within the cluster chart. Eg independent of bastions, private clusters etc. But some other parts in the chart are still bound to kubernetes too. So if we want to have a team responsible for kubernetes itself and other teams responsible for infrastructure integration in a specific infrastructure like aws or openstack, how can we split responsibility within the same chart? Does the kubernetes team then still need to do releases of each provider chart? And is that a desired design? Are we even able to separate this completely? We need to make sure we separate the configuration for k8s itself and the provider one. The k8s team still needs to run tests on all providers and they need to have changes in their own chart dependent on which provider. But without the need to update all charts that are maintained by the infrastructure teams. They need to be able to roll out their own changes to all providers.

Another idea (not sure if discussed already) would be to allow customers to select something like beta, stable, long-term moving targets. This way teams can release their changes and the release gets tested further along and promoted through the stages.

teemow:
@pipo02mix regarding your last comment. I think the design is important and the decisions will have consequences. As @puja108 mentioned the interface for future teams is important. Is the matrix the best solution to separate the OS lifecycle (and partly also k8s) from the provider specific chart? Imo this would mean that a team responsible for the OS needs to do PRs in all provider charts. This is an anti-pattern imo which we already have in several places and we keep accumulating toil for the engineers.

I wrote some more here: 
 * https://github.com/giantswarm/rfc/pull/54#discussion_r1038162567
 * https://github.com/giantswarm/rfc/pull/54#discussion_r1038179218

With this taken into account I would also change this:

> As a customer, I would like to choose between the two supported minor Kubernetes versions without having to think about choosing an OS or machine image.

to:

> As a customer, I would like to choose between different release channels of kubernetes versions (eg ranging from alpha, beta, stable to LTS)

I'd like to bring in some clarity here as this seems to be stuck. Please invite to a meeting with @alex-dabija @cornelius-keller @gawertm @puja108 @JosephSalisbury and me to get consensus and take a decision how to move forward.

capi-matrix/README.md
kopiczko: I agree it should be hidden from the customer. Did we ever had a critical issue requiring us to change the OS so quickly we wouldn't be able to change a few lines in the chart and release it?

capi-matrix/README.md
pipo02mix: IMO forcing a version per cluster_PROVIDER release limits us not the customer. Let me explain it. In our whole history we have been managing clusters and apps, and in many times across all customers we have some requirements, blockers,... that block us to upgrade to certain versions of an app or a k8s (even OS when cgroups).

This is going to happen no matter how good the cluster upgrade mechanics are. You also mention other conditions like frozen periods or software dependencies (openet ehem ehem) that force us to offer that option.

IMO the problem we had in the past is the manual toil of schedule, discuss, manual preparation, ... which required us to do an upgrade. This is what we need to solve and for that reason [the automate cluster upgrade rfc](https://github.com/giantswarm/rfc/tree/main/automatic-cluster-upgrades) was created. 

I don't say that we need to put the list of supported versions in cluster-PROVIDER charts, we could find a better place, but we will need that list somewhere. Leaving the chart without the default k8s/OS version and defining it in another place for all providers is another option. But we need to clarify the vision of how clusters/app should be updated.

capi-matrix/README.md
pipo02mix: No I don't think so. It can be hidden of course, I have a discussion with someone else and making it part of values seemed logical for other possible users (not GS). 

And Timo I expected that if there is a K8s team releasing new version there is an automated process that creates the release for each provider (anyway so far is something we do manually, so this is not getting worse). Having multiple release channels could be a solution, but I would not make it hard to select a specific version of k8s/OS since it will could impact us in the future having to make workaround to pin versions.

PR #55, opened 2022-11-25
Add RFC for schema requirements in cluster apps

marians:
Towards https://github.com/giantswarm/roadmap/issues/1733

This RFC proposes standards (with different levels of strictness) that should at some point be met by our CAPI cluster apps (the apps defining workload clusters).


cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,199 @@
...
+Example:
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  ...
+}
+```
+
+The example below shows an excerpt of a JSON Schema which uses draft 2020-12.
```
Rotfuks, line 51: The example is above, not below, right? :)

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+All properties that can be used in values MUST be covered in the schema explicitly.
+
+To enforce this and to disable the use of any undefined properties, the keyword `additionalProperties` SHOULD be set to `false` on all object schemas.
+
+The `patternProperties` keyword MUST NOT be used.
+
+### R3: Array item schema must be defined {#r3}
+
+The items schema for all array properties must be defined using the `items` keyword.
```
Rotfuks, line 63: this MUST should also be all upper case to match the others

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+I haven't gotten to these yet, or I'm not sure about them.
+
+- Each property must declare the `type`. Not sure this is required by JSON schema.
+
+- Use of the `default` keyword. As per JSON Schema documentation, it is for annotation only. It is not meant to automatically fill in missing values for validation.
+
+- How to specify whether a property can be modified or not. `"readOnly": true` mit be the right one for that.
+
+- Use of the `deprecated` keyword to phase out properties. Documentation says: "The deprecated keyword is a boolean that indicates that the instance value the keyword applies to should not be used and may be removed in the future."
```
Rotfuks, line 195: I would definitely add this as mandatory. Deprecation notice as the most minimal form of versioning makes contracts for common schemas way easier. Not just removing things but first deprecating it definitely builds important trust into the schemas

cluster-app-values-schema/README.md
marians: Yes

cluster-app-values-schema/README.md
marians: In the meantime I learned that Helm does not make use of the `deprecated` property of JSON schema at all. There is a little risk that it might mean something different at some point in the Helm world than in the Giant Swarm world, once they start making use of it. Should that worry us?

Context: https://github.com/helm/helm/issues/10732

Rotfuks: I don't think there is that many use cases for that word besides actually deprecating things, everything else would be expectation breaking. So I wouldn't worry too much - or at least not yet :D.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+The example above shows an excerpt of a JSON Schema which uses draft 2020-12.
+
+### R2: Schema must explicitly cover all allowed properties {#r2}
+
+All properties that can be used in values MUST be covered in the schema explicitly.
+
+To enforce this and to disable the use of any undefined properties, the keyword `additionalProperties` SHOULD be set to `false` on all object schemas.
+
+The `patternProperties` keyword MUST NOT be used.
```
mogottsch, line 59: Why do we prohibit the use of `patternProperties`?
If there exists an object that allows arbitrary properties like [here](https://github.com/giantswarm/cluster-cloud-director/blob/main/helm/cluster-cloud-director/values.schema.json#L269), then it would be more specific to use `patternProperties` to constrain how these keys look like. E.g. a regex that enforces all lowercase characters.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,237 @@
...
+- **JSON Schema**: declarative language that allows to annotate and validate documents that are compatible with JSON.
+- The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html)
+
+## Requirements
+
+Requirements carry identifiers with prefix `R` and a unique number, for easy referencing.
+
+### R1: JSON Schema dialect must be specified {#r1}
+
+Each cluster app's values schema file MUST specify the schema dialect (also called draft) used via the `$schema` keyword on the top level, with the draft URI as a value.
```
mogottsch, line 40: The meaning and possible values of specific keywords changed over the course of different drafts. Supporting all drafts would increase the complexity of our tooling, e.g. `schemalint`. I would suggests that we initially only support the newest draft (2020-12) and add support for newer drafts as they are released.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+All properties that can be used in values MUST be covered in the schema explicitly.
+
+To enforce this and to disable the use of any undefined properties, the keyword `additionalProperties` SHOULD be set to `false` on all object schemas.
+
+The `patternProperties` keyword MUST NOT be used.
+
+### R3: Array item schema must be defined {#r3}
+
+The items schema for all array properties MUST be defined using the `items` keyword.
```
mogottsch, line 63: Do we allow the use of `prefixItems` ([understandable reference](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation), [rigorous reference](https://json-schema.org/draft/2020-12/json-schema-core.html#name-prefixitems))?
It has potential use cases, but probably not for us.
As it increases the complexity of our tooling, I would prohibit its use until someone actually wants it.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+- [R1: JSON Schema dialect must be specified](#r1)
+- [R2: Schema must explicitly cover all allowed properties](#r2)
+- [R3: Array item schema must be defined](#r3)
+- [R4: Properties must have a title](#r4)
+- [R5: Properties should have descriptions](#r5)
+- [R6: Properties should provide examples](#r6)
+- [R7: Constrain values as much as possible](#r7)
+- [R8: Required properties must be marked as such](#r8)
+
```
mogottsch, line 17: R9 is missing here.

cluster-app-values-schema/README.md
marians: A valid concern. I guess if we support `additionalProperties`, then we must also support `patternProperties`.

marians: Removing the line for now.

cluster-app-values-schema/README.md
marians: OK

cluster-app-values-schema/README.md
marians: Good catch. To be discussed. Depends on whether we can support it in the UI.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,205 @@
...
+Properties of type `string`:
+
+- Use of the keywords `minLength` and `maxLength` to specify the valid length.
+- If applicable, use the `pattern` keyword to specify a regular expression for validation. This is useful for example to restrict the value to ASCII characters, or to lowercase etc.
+- Use the `format` keyword to restrict the property to a common string format, like a URL or an email address. Watch out to use only formats which are available in the JSON schema dialect used (see also R1).
+- Use `enum` if only certain known values can be valid.
+
+Numeric properties (type `number`, `integer`):
+
+- Restrict the values range using `minimum` or `exclusiveMaximum` and `maximum` or `exclusiveMaximum`.
```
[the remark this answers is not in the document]
marians: Fixed

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+Best practices:
+
+- Do not repeat the property name or title in the description.
+- Write descriptions between 50 and 200 characters long.
+- Use simple language.
+- Use no formatting syntax, nor hyperlinks.
+- Use no line breaks.
```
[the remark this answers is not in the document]
marians: Changed

marians:
Updated the draft based on suggestions. There is a new requirement R10 dealing with the `deprecated` key.

cluster-app-values-schema/README.md
gusevda: `draft-2020-12` hasn't been fully tested with `react-jsonschema-form` - [docs](https://react-jsonschema-form.readthedocs.io/en/latest/usage/validation/#ajvclass). Maybe it's better to ask schema maintainers to use a previous draft - `draft-2019-09`?
For example, `prefixItems` property that was introduced in `draft-2020-12` is not supported by the react library, so if we want such functionality, we would need to use `additionalItems` property from the `draft-2019-09`.
https://json-schema.org/draft/2020-12/release-notes.html#changes-to-items-and-additionalitems

mogottsch: I agree. If the library does not support `2020-12`, then we shouldn't use it.
Even though, it surprises me that they don't fully support it, as it's been out for over two years. Hopefully, they are working towards it.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,283 @@
...
+  "maximum": 200,
+  "multipleOf": 10
+}
+```
+
+### R8: Required properties {#r8}
+
+If a property is required to be set by the creator of a cluster, it MUST be included in the `required` keyword.
+
+Properties that have default values defined in the app's `values.yaml` file MUST NOT be set as required, as here the default values are applied in case no user value is given.
```
mogottsch, line 184: Regarding the discussion after the JSON schema intro lunch talk:
It seems to me that people are bothered by the fact that they have to maintain `values.yaml` and `values.schema.json`. We could work towards a solution that generates `values.yaml` from `values.schema.json`.
In that case we should mention that we use `default` for that purpose.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+}
+```
+
+Best practices:
+
+- Ideally, provide one example.
+- Provide more than one example if you want to highlight that very different values are accepted.
+- Provide no more than five examples
```
mogottsch, line 145: Should this be a MUST/SHOULD rather than a best practice?

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,283 @@
...
+Property schema SHOULD explicitly restrict values as much as possible.
+
+There are several ways this requirement can/has to be fulfilled, depending on the property type and the use case.
+
+Properties of type `string`:
+
+- Use of the keywords `minLength` and `maxLength` to specify the valid length.
+- If applicable, use the `pattern` keyword to specify a regular expression for validation. This is useful for example to restrict the value to ASCII characters, or to lowercase etc.
+- Use the `format` keyword to restrict the property to a common string format, like a URL or an email address. Watch out to use only formats which are available in the JSON schema dialect used (see also R1).
+- Use `enum` if only certain known values can be valid.
```
mogottsch, line 162: Can we make our requirements here more specific?
E.g.
Properties of type `string`...

- ...SHOULD use `minLength` and `maxLength` to specify the valid length, unless `format` or `enum` is already used.
-  ...SHOULD have one of the following properties: `pattern`, `format` or `enum`.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+Numeric properties (type `number`, `integer`):
+
+- Restrict the values range using `minimum` or `exclusiveMinimum` and `maximum` or `exclusiveMaximum`.
+- Use `multipleOf` if the value has to be a multiple of a certain number.
```
mogottsch, line 167: Same here.

I suggest SHOULD for the first bullet point and MAY for the second.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Requirements
+
+Requirements carry identifiers with prefix `R` and a unique number, for easy referencing.
+
+### R1: JSON Schema dialect (draft 2019-09) must be specified {#r1}
+
+Each cluster app's values schema file MUST specify the schema dialect (also called draft) used via the `$schema` keyword on the top level, with the draft URI as a value.
+
+The draft URI MUST be `https://json-schema.org/draft/2019-09/schema`.
```
mogottsch, line 45: It is possible to reference external schemas with different draft versions. This could be problematic if we rely on a certain draft version in happa and in schemalint.
I suggest we restrict external schemas usage to those with the correct draft.

marians: It's indeed getting complicated if a referenced schema uses a different draft than the main one. Does the JSON schema spec mention anything about how to handle this?

mogottsch: I tried to pinpoint this in the specification, but I didn't find it. However, the author of the JSON schema library we use said that [here](https://github.com/santhosh-tekuri/jsonschema/issues/103#issuecomment-1408736354) and I definitely trust him to know the specification.

cluster-app-values-schema/README.md
marians: This is decided: we don't allow `prefixItems`. Resolving this conversation.

cluster-app-values-schema/README.md
marians: Will clarify this.

cluster-app-values-schema/README.md
marians: Clarification coming with the next commit.

cluster-app-values-schema/README.md
marians: Actually I removed the mention of `multipleOf` because it is a feature that can simply be used, but we don't have anything to say about it.

cluster-app-values-schema/README.md
marians: With the latest restrictions we discussed (e.g. forbid `prefixItems`), does it still make sense to require 2019-09? Or can we go back to 2020-12?

mogottsch: Yes, if we make sure that we are indifferent to all the changes between 2019-09 and 2020-12.
That would be:
- forbid `prefixItems` (2020)
- forbid `additionalItems` (2019)
- forbid `items` with tuple values (2019)
- forbid `$dynamicRef` and `$dynamicAnchor` (2020)
- forbid `$recursiveRef` and `$recursiveAnchor` (2019)
As far as I know, we already wanted to forbid all of these. So we could go back to 2020-12.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,319 @@
...
+- [R3: Array item schema must be defined](#r3)
+- [R4: Properties must have a title](#r4)
+- [R5: Properties should have descriptions](#r5)
+- [R6: Properties should provide examples](#r6)
+- [R7: Constrain values as much as possible](#r7)
+- [R8: Required properties must be marked as such](#r8)
+- [R9: Avoid `anyOf` and `oneOf`](#r9)
+- [R10: Use `deprecated` to phase out properties](#r10)
+- [R11: Provide valid string values where possible](#r11)
+- [R13: Avoid recursion](#r13)
```
mogottsch, line 21: We skipped R12

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,373 @@
...
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  ...
+}
+```
+
+### R2: A single type must be declared {#r2}
+
+For each property, including the root level schema, the `type` keyword MUST be present, and there MUST be a single value for the `type` keyword.
```
mogottsch, line 64: I think we don't need `type`, when `$ref` is specified. For example:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "address": {
      "$ref": "address.json"
    },
  "type": "object"
}

```
 Here I would expect that we don't require a type for `address`.

marians: Yes, true. Do you think this should be mentioned? To me that's covered by the definition of `$ref` in JSON Schema.

mogottsch: This is important for the implementation of this rule in schemalint. JSON Schema does not forbid using `type` next to `$ref`. However, I think leaving out `type` when `$ref` is present is so intuitive, that we don't necessarily need to mention it here.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,373 @@
...
+    {"const": "IfNotPresent", "title": "Pull image if not present in the node"},
+    {"const": "Always", "title": "Always pull up-to-date image"},
+    {"const": "Never", "title": "Never pull image, use image available locally"}
+  ]
+}
+```
+
+### R13: Avoid recursion {#r13}
+
+The JSON Schema keywords `dynamicRef`, `dynamicAnchor`, and `recursiveRef` MUST NOT be used.
```
mogottsch, line 318: The title says to avoid recursion. Do we forbid recursion in general or just these keywords?
I know that we talked about these keywords, because they make implementation in schemalint more cumbersome. But I guess for happa self-referencing recursion is impossible to display properly in the UI.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,373 @@
...
+
+The JSON Schema keywords `unevaluatedProperties` , `unevaluatedItems` MUST NOT be used.
+
+### R16: Array items and tuple validation {#r16}
+
+All array items MUST be of the same type.
+
+The JSON Schema keywords `contains`, `additionalItems` and `prefixItems` MUST NOT be used.
+
+### R17: Common schema structure {#r17}
```
mogottsch, line 334: Do we prohibit other properties at the root level than those described below?

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,373 @@
...
+| Property name | Property type | Description |
+|-|-|-|
+| `metadata` | object | Descriptive settings like name, description, cluster labels (e. g. service priority), owner organization |
+| `connectivity` | object | Settings related to connectivity and networking, and defining how the cluster reaches the outside world and how it can be reached. |
+| `controlPlane` | object | Configuration of the cluster's control plane, Etcd, API server and more. |
+| `nodePools` | array | Configuration of node pools (groups of worker nodes), regardless of their implentation flavour. |
+
+In addition, the cluster app values schema SHOULD offer the following root level properties:
+
+| Property name | title | description |
```
mogottsch, line 349:
```suggestion
| Property name | Property type | Description |
```

cluster-app-values-schema/README.md
marians: That's the idea, eventually.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,384 @@
...
+}
+```
+
+### R3: Explicitly cover all allowed properties {#r3}
+
+All properties that can be used in values MUST be covered in the schema explicitly.
+
+To enforce this and to disable the use of any undefined properties, the keyword `additionalProperties` MUST be set to `false` on the schema's root object.
+
+For objects deeper in the schema hierarchy, keyword `additionalProperties` SHOULD be set to `false`, too.
```
mogottsch, line 90: I think here we need to adapt. `additionalProperties` can also be set to a schema.
It should be always set and never to `true`.

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,383 @@
...
+        "type": "string",
+        "pattern": "^[a-z0-9]{5,10}$",
+        "examples": ["devel001"],
+        ...
+    }
+  }
+}
+```
+
+An example can give users easy-to-understand guideance on how to use a property. Ideally, examples SHOULD be valid cases, fulfilling the contstraints of the property (if given).
```
mogottsch, line 172:
```suggestion
An example can give users easy-to-understand guidance on how to use a property. Ideally, examples SHOULD be valid cases, fulfilling the constraints of the property (if given).
```

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,383 @@
...
+|-|-|-|
+| `internal` | object | Settings which are not supposed to be configured by end users, and which won't be exposed via user interfaces. Also experimental features that undergo schema changes. |
+| `providerSpecific` | object | Configuration specific to the infrastructure provider. |
+
+For compatibility reasons, the schema MAY have the following properties in the root level:
+
+- `managementCluster`
+- `baseDomain`
+- `provider`
+- `cluster-shared`
```
AndiDog, line 360: We can link to https://github.com/giantswarm/cluster-shared/tree/main/helm/cluster-shared as explanation

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,383 @@
...
+
+- How to specify whether a property can be modified or not. `"readOnly": true` mit be the right one for that.
+
+- Clarify use of `Not`.
+
+- dependentRequired, dependentSchemas: to be evaluated.
+
+- contentEncoding in combination with contentSchema: to be defined.
+  - see "username:password" in base 64 example
+
```
AndiDog, line 380: What about "How to verify/format JSON schema files" (in your editor)

cluster-app-values-schema/README.md
```diff
@@ -0,0 +1,383 @@
...
+For compatibility reasons, the schema MAY have the following properties in the root level:
+
+- `managementCluster`
+- `baseDomain`
+- `provider`
+- `cluster-shared`
+
+The schema MUST NOT define any other properties on the root level, in addition to the ones mentioned above.
+
```
mogottsch, line 363: As we forbid empty objects, I suggest adding that to the RFC ([related issue](https://github.com/giantswarm/roadmap/issues/2110)).

```suggestion

### R18: Avoid empty objects {#r18}

All objects MUST specify at least one property through at least one of the following keywords: `properties`, `additionalProperties`, `patternProperties`.

```

cluster-app-values-schema/README.md
AndiDog: Okay, that was a beginner question I guess 😆 – `additionalProperties` is for when the object keys can be dynamic, such as in `nodePools: { abc: ..., def: ... }`.

mogottsch: Exactly. `additionalProperties` can be used if the keys can be arbitrary and there is also `patternProperties`, which uses a regex pattern to restrict the keys. This one is also used in our schemas, AFAIK.

PR #56, opened 2022-11-30
Updating RFC for configs merging

ljakimczuk:
Towards: https://github.com/giantswarm/gitops-template/pull/79

### DIFF
--- a/merging-configmaps-gitops/README.md
+++ b/merging-configmaps-gitops/README.md
@@ -74,7 +74,7 @@ spec:
 or should be left empty to be later populated by either `app-admission-controller`, upon submission, or `app-operator`, upon
 reconciliation. Using this field to provide other values may result in the mis-configuration of certain apps.**
 
-The values coming from these fields are merged based on the priority, see the [App Platform configuration](https://docs.giantswarm.io/developer-platform/app-platform/app-configuration/). Both, the `.spec.config` and `.spec.userConfig`,
+The values coming from these fields are merged based on their priority, see the [App Platform configuration](https://docs.giantswarm.io/developer-platform/app-platform/app-configuration/). Both, the `.spec.config` and `.spec.userConfig`,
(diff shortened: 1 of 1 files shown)


merging-configmaps-gitops/README.md
```diff
@@ -45,37 +45,50 @@ Full context of the PR can be found here: https://github.com/giantswarm/workload
...
       name: flux01-userconfig
       namespace: org-some
 ```
-This functionality means that values in `config` will be overlaid by values from `userConfig`.
-So far we have only been using this functionality to supply Giant Swarm offered configuration through `config` and encouraged end users to only use the `userConfig` field.
 
-The proposal is to use `config` for cases in gitops where we can not merge configmaps through `kustomize`.
+**The `.spec.config` should be used with caution. It should either be set explicitly to the `<CLUSTER_NAME>-cluster-values`,
+or should be left empty to be later populated by either `app-admission-controller`, upon submission, or `app-operator`, upon
+reconciliation. Using this field to provide other values may result in underconfiguring certain apps.**
```
mproffitt, line 75:
```suggestion
reconciliation. Using this field to provide other values may result in the mis-configuration of certain apps.**
```

merging-configmaps-gitops/README.md
```diff
@@ -45,37 +45,50 @@ Full context of the PR can be found here: https://github.com/giantswarm/workload
...
+
+The values coming from these fields are merged based on the priority, see the [App Platform configuration](https://docs.giantswarm.io/developer-platform/app-platform/app-configuration/). Both, the `.spec.config` and `.spec.userConfig`,
```
mproffitt, line 77:
```suggestion
The values coming from these fields are merged based on their priority, see the [App Platform configuration](https://docs.giantswarm.io/developer-platform/app-platform/app-configuration/). Both, the `.spec.config` and `.spec.userConfig`,
```

mproffitt (changes requested):
Just minor changes to wording here

mproffitt (approved):

PR #58, opened 2022-12-13
add logging-infrastructure RFC

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/311

Following up on recent discussions around our logging architecture, here is an RFC which summarize the centralized vs distributed approach pro and cons, and the next steps forward.

Feel free to comment, voice your opinion or add anything missing.

### DIFF
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -44,11 +44,17 @@ So logs are stored and accessible in a single and central place.
 
 There is a hard requirement from some customer where no data should leave their installation (due to legal concerns).
 In order to comply with this requirement we need to adopt a distributed setup.
-We currently have some concerns towards the feasibility of implementing Loki on all the different providers we support, due to storage requirements.
 
 ## Next steps
 
-- Build a POC with a Loki running inside an installation starting with 1 provider.
+### Requirements
+
+- object storage: in order for data to be stored at the installation level, Loki requires access to an object storage solution
+                  this is currently unknown and needs to be evaluated on a per-provider basis
+
+### Steps
+
+- Build a POC following the distributed setup with 1 provider.
 - Re-evaluate and see how we proceed with other providers and other customer requirements.
 
 ## Open questions
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -12,8 +12,7 @@ There were some recent discussions where we learned about other requirements lea
 
 ### Distributed setup
 
-* 1 Loki instance for each installation
-* Loki is hosted at the installation level
+1 Loki instance for each installation, each instance is hosted at the installation level
 
 #### Pros
 
@@ -27,8 +26,7 @@ There were some recent discussions where we learned about other requirements lea
 
 ### Centralized setup
 
-* Single central Loki instance
-* Loki is hosted by GiantSwarm
+Single central Loki instance, which is hosted by GiantSwarm
 
 #### Pros
 
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -1,0 +1,47 @@
+# Logging infrastructure
+
+As a Giant Swarm engineer, I want to be able to access a history of logs for components managing our platform in order to be to both investigate ongoing operational issues and provide details for incident reports.
+
+## Proposals
+
+High level architecture proposal to implement a logging solution.
+
+### Distributed (per installation) setup
+
+Having 1 Loki instance per installation
+
+### Pros
+- Satisfy customer requirement towards data privacy (data stays in customer hands)
+- Cost re-use our current billing system
+- Keep our Platform as a Product model (self-sufficient installation can keep running alone)
+- Resilient setup
+### Cons
+- Operational cost (many Lokis instance to handle)
+- Heterogeneous Object storage (different provider, different requirements, different limitations)
+- No global view (aka single pane of glass, no cross-installation queries)
+
+## Centralized setup
+
+Having a single Loki instance hosted by GiantSwarm
+
+### Pros
+- Global view (aka single pane of glass, cross-installation queries)
+- Anomalies detections (multi installation correlated data)
+- Operational cost (1 Loki instance to handle)
+
+### Cons
+- Data privacy (data stored in GiantSwarm account)
+- Cost needs a new billing system and business model
+- Single point of failure
+
+## Decision
+
+<tbd>
+
+## Requirements
+
+- object storage provisioning
+
+### Next steps
+- how to automate object storage provisioning ?
+- can we provide a global view on a distributed setup ? (maybe on a data subset)
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -19,7 +19,6 @@ So logs are stored and accessible on a per installation basis.
 - More resilient setup towards failure
 - Keep our Platform as a Product model (self-sufficient installation can keep running alone)
 
-
 #### Cons
 
 - Heterogeneous Object storage (different provider, different requirements, different limitations)
@@ -38,6 +37,8 @@ So logs are stored and accessible in a single and central place.
 #### Cons
 
 - Cost needs a new billing system and business model
+- Customer does not own their data (i.e. lose logs access when leaving GiantSwarm)
+- Requires good connectivty from installations to Loki (i.e. edge cluster might be a problem)
 
 ## Conclusion
 
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -32,7 +32,7 @@ So logs are stored and accessible in a single and central place.
 #### Pros
 
 - Global view (aka single pane of glass, cross-installation queries)
-- Anomalies detections (multi installation correlated data)
+- Anomalies detections (i.e. detect issues earlier via error/trends measures accross installations/regions/providers)
 
 #### Cons
 
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -1,47 +1,56 @@
 # Logging infrastructure
 
-As a Giant Swarm engineer, I want to be able to access a history of logs for components managing our platform in order to be to both investigate ongoing operational issues and provide details for incident reports.
+As a Giant Swarm engineer, I want to be able to access a history of logs for components managing our platform in order to both investigate ongoing operational issues and provide details for incident reports.
 
-## Proposals
+## Context
 
 High level architecture proposal to implement a logging solution.
 
-### Distributed (per installation) setup
+Atlas started working on a logging infrastructure implementation using [Loki](https://grafana.com/docs/loki/latest/).
+So far we worked on a POC towards having a centralized setup.
+There were some recent discussions where we learned about other requirements leaning towards a distributed setup, here are the pro and cons details of those setups.
 
-Having 1 Loki instance per installation
+### Distributed setup
 
-### Pros
-- Satisfy customer requirement towards data privacy (data stays in customer hands)
-- Cost re-use our current billing system
+1 Loki instance for each installation
+Loki is hosted at the installation level
+
+#### Pros
+
+- Satisfy customer requirement towards data isolation (i.e. legal, or data privacy requirements)
 - Keep our Platform as a Product model (self-sufficient installation can keep running alone)
-- Resilient setup
-### Cons
-- Operational cost (many Lokis instance to handle)
+- More resilient setup towards failure
+
+#### Cons
+
 - Heterogeneous Object storage (different provider, different requirements, different limitations)
-- No global view (aka single pane of glass, no cross-installation queries)
 
-## Centralized setup
+### Centralized setup
 
-Having a single Loki instance hosted by GiantSwarm
+Single central Loki instance
+Loki is hosted by GiantSwarm
 
-### Pros
+#### Pros
+
 - Global view (aka single pane of glass, cross-installation queries)
-- Anomalies detections (multi installation correlated data)
 - Operational cost (1 Loki instance to handle)
+- Anomalies detections (multi installation correlated data)
 
-### Cons
-- Data privacy (data stored in GiantSwarm account)
+#### Cons
+
 - Cost needs a new billing system and business model
-- Single point of failure
 
-## Decision
+## Conclusion
 
-<tbd>
+There is a hard requirement from some customer where no data should leave their installation (due to legal concerns). In order to comply with this we need to adopt a distributed setup.
+We currently have some concerns towards the feasibility of implementing Loki on all the different providers we support, due to storage requirements.
 
-## Requirements
+## Next steps
 
-- object storage provisioning
+- Build a POC with a Loki running inside an installation starting with 1 provider.
+  - crossplane and work done on harbour might help us for storage
+- Re-evaluate and see how we proceed with other providers and other customer requirements.
 
-### Next steps
-- how to automate object storage provisioning ?
-- can we provide a global view on a distributed setup ? (maybe on a data subset)
+## Open questions
+
+- Can we provide a global view on a distributed setup ? (maybe on a data subset)
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -49,8 +49,8 @@ In order to comply with this requirement we need to adopt a distributed setup.
 
 ### Requirements
 
-- object storage: in order for data to be stored at the installation level, Loki requires access to an object storage solution
-                  this is currently unknown and needs to be evaluated on a per-provider basis
+- object storage: in order for data to be stored at the installation level, Loki requires access to an object storage solution.
+                  This is currently unknown and needs to be evaluated on a per-provider basis
 
 ### Steps
 
--- a/logging-infrastructure/README.md
+++ b/logging-infrastructure/README.md
@@ -12,8 +12,8 @@ There were some recent discussions where we learned about other requirements lea
 
 ### Distributed setup
 
-1 Loki instance for each installation
-Loki is hosted at the installation level
+* 1 Loki instance for each installation
+* Loki is hosted at the installation level
 
 #### Pros
 
@@ -27,8 +27,8 @@ Loki is hosted at the installation level
 
 ### Centralized setup
 
-Single central Loki instance
-Loki is hosted by GiantSwarm
+* Single central Loki instance
+* Loki is hosted by GiantSwarm
 
 #### Pros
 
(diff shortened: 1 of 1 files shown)


QuantumEnigmaa (approved):
Very descriptive, I love it !

logging-infrastructure/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+#### Pros
+
+- Global view (aka single pane of glass, cross-installation queries)
+- Operational cost (1 Loki instance to handle)
+- Anomalies detections (multi installation correlated data)
+
+#### Cons
+
+- Cost needs a new billing system and business model
```
hervenicol, line 40: customer leaving GS loses his data
requires good connectivity from customer cluster to Loki (thinking edge clusters)

hervenicol (commented):
Maybe add a "pre-requisites" part?

With:
* object storage
* Crossplane for object storage provisioning automation

hervenicol (approved):
LGTM

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
+# Logging infrastructure
+
+As a Giant Swarm engineer, I want to be able to access a history of logs for components managing our platform in order to both investigate ongoing operational issues and provide details for incident reports.
```
puja108, line 3: We might want to share this infra with customers, so the use case might be for both GS and customers. While they are similar the amount of data and its respective sensitivity with the customer use case is higher.

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+                  This is currently unknown and needs to be evaluated on a per-provider basis
+
+### Steps
+
+- Build a POC following the distributed setup with 1 provider.
+- Re-evaluate and see how we proceed with other providers and other customer requirements.
+
+## Open questions
+
+- Can we provide a global view on a distributed setup ?
```
puja108, line 62: similar to what we do with the prometheus remote write to a central instance but with lower cardinality and only a selection of data.

puja108 (approved):
generally LGTM, summarizes our last meeting well and adds some thoughts to complications

as for object storage, for now maybe we can focus on cloud installations and see if we can re-use the Crossplane work done by Honeybadger, which is also a nice first internal use case for combining app and cloud resources.

logging-infrastructure/README.md
JosephSalisbury: +1, with the whole of idea of being a developer platform, i think customers having access to logs is a requirement

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Context
+
+Atlas started working on a logging infrastructure implementation using [Loki](https://grafana.com/docs/loki/latest/).
+So far we worked on a POC towards having a centralized setup, where a central Loki instance hosted by GiantSwarm ingests logs from our installations.
+There were some recent discussions where we learned about other requirements leaning towards a distributed setup, here are the pro and cons details of those discussions.
+
+### Distributed setup
+
+What we call distributed setup, is when there is 1 Loki instance hosted on each installation.
```
JosephSalisbury, line 13: what's the scaling like with loki? would we we need to think about sharding along some domain lines?

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+#### Pros
+
+- Satisfy customer requirement towards data isolation (i.e. legal, or data privacy requirements)
+- More resilient setup towards failure
+- Keep our Platform as a Product model (self-sufficient installation can keep running alone)
+
+#### Cons
+
+- Heterogeneous Object storage (different provider, different requirements, different limitations)
```
JosephSalisbury, line 24: this is a contra because we lack object storage by default in the platform, right? not because we have to use object storage in a generic sense?

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+There is a hard requirement from some customer where no data should leave their installation (due to legal concerns).
+In order to comply with this requirement we need to adopt a distributed setup.
+
+## Next steps
+
+### Requirements
+
+- object storage: in order for data to be stored at the installation level, Loki requires access to an object storage solution.
+                  This is currently unknown and needs to be evaluated on a per-provider basis
```
JosephSalisbury, line 53: this is probably something we need to bring up in product in a more general sense, yeah

logging-infrastructure/README.md
JosephSalisbury: or have a look at some kind of federated querying solution

logging-infrastructure/README.md
hervenicol: No need to shard at application level, but we should use a different loki "tenant-id" for each cluster's promtails to limit blast radius in case one generates too many logs.

Basically:
* Data storage is S3 (or similar), scaling it is not our problem.
* app scales horizontally quite well, with separate scaling for read and write paths
* organizations or tenants are individually rate-limited, so it limits blast radius if one sends too much logs.

For instance, Grafana cloud is one common Loki instance for all customers.

Is it worth writing this in the RFC?

logging-infrastructure/README.md
hervenicol: :100: 

Object storage is great, that what makes these products so easy to use.
But this is the constraint that may prevent us from deploying it on all platforms.

logging-infrastructure/README.md
primeroz: FWIW i had setup in the past Loki in [microservice mode](https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/#microservices-mode) and it was handling about 15 shipping clusters at 40MB/s 90k/s(messages) with no much sweat since with this deployment mode it was easy to scale the read and write path independently . 
The write path _did need_ some resources to keep up

JosephSalisbury: Okay cool, that's what I reckoned, but just wanted to check. I don't think it's necessary to add to the RFC

logging-infrastructure/README.md
JosephSalisbury: ack

logging-infrastructure/README.md
hervenicol: AFAIK there's no official feature for that today.
To be investigated later IMO.

logging-infrastructure/README.md
puja108: We already need it for quite a few upcoming apps and customers most probably also have their use cases for object storage that we could make easier.

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Distributed setup
+
+What we call distributed setup, is when there is 1 Loki instance hosted on each installation.
+So logs are stored and accessible on a per installation basis.
+
+#### Pros
+
+- Satisfy customer requirement towards data isolation (i.e. legal, or data privacy requirements)
+- More resilient setup towards failure
```
QuentinBisson, line 19: I'm not really sure I agree with that one unless you have arguments. With a centralized setup, there is no reason why we cannot create a failover mechanism in another region (s3 backups and so on)

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+There were some recent discussions where we learned about other requirements leaning towards a distributed setup, here are the pro and cons details of those discussions.
+
+### Distributed setup
+
+What we call distributed setup, is when there is 1 Loki instance hosted on each installation.
+So logs are stored and accessible on a per installation basis.
+
+#### Pros
+
+- Satisfy customer requirement towards data isolation (i.e. legal, or data privacy requirements)
```
QuentinBisson, line 18: I would rephrase that because loki multitenancy satisfies this. The requirement would be more about the fact they don't want their logs outside of their machines, otherwise it can be removed

logging-infrastructure/README.md
QuentinBisson: I would even think distributed is less resilient (unless we implement the failover mechanism with Cross plane on all installations)

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+- More resilient setup towards failure
+- Keep our Platform as a Product model (self-sufficient installation can keep running alone)
```
QuentinBisson, line 20: if customers are leaving, they get to keep their Loki. We could probably work on a "migration" path to get the centralized logs in an installation Loki for them if they leave but that would simplify it

logging-infrastructure/README.md
QuentinBisson, line 18: FYI, I think this is the only real reason why we have to go distributed

logging-infrastructure/README.md
QuentinBisson, line 24: I definitely would add operations, because operating x lokis is definitely harder than 1, especially since they will need to still handle multi tenancy at least at the organization level

QuentinBisson, line 24: Resource usage might be a con (because this will probably require more resources)

logging-infrastructure/README.md
QuentinBisson, line 18: I'd add simplify sales because we have a hard time selling extra money for logging when they would pay for extra object storage/machine needs to host a Loki per installation

logging-infrastructure/README.md
QuentinBisson, line 24: This would be more expensive for customers (at least for the machines, not sure about object storage)

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+#### Cons
+
+- Heterogeneous Object storage (different provider, different requirements, different limitations)
+
```
QuentinBisson, line 25: Another con is that we lose the use case of checking logs accross all installations which would be useful for things like the app platform crossplanes and so on (e.g. error trends accross installations/regions/providers to detect issues earlier than if we check at the installation level)

QuentinBisson: cc @piontec as this is something you would have loved to have

logging-infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+### Centralized setup
+
+What we call centralized setup, is when there is a single central Loki instance hosted by GiantSwarm.
+So logs are stored and accessible in a single and central place.
+
+#### Pros
+
+- Global view (aka single pane of glass, cross-installation queries)
+- Operational cost (1 Loki instance to handle)
```
QuentinBisson, line 34: Compute cost also

logging-infrastructure/README.md
QuentinBisson: IMO, this makes no sense to me to have a subsect or a selection of the data because we would lose a lot of the logs values (context labels, exemplars, history) for no benefits. Also, this contradicts the fact customers do not want logs outside their installations.
The only way I see this going forward is to create metrics from logs and send those metrics to the central instance so we can try to get trends.

Grafana is definitely not working toward any kind of federation because they think 1 Loki is the way to go and i'd rather not go down the rabbit hole of hosting a meh solution that nobody uses like Promxy to fix that use case

puja108: customers worry about logs of their workloads and sensitive components, a subset to centralize for us could be k8s and some of our own components logs

logging-infrastructure/README.md
TheoBrigitte: added to the Centralized setup's cons section

logging-infrastructure/README.md
TheoBrigitte: This point is good to have but we have to first work on it in order to see if its doable and what solutions we have.

logging-infrastructure/README.md
TheoBrigitte: It is more resilient in the sense that multiple isolated Loki running on each installation are less likely to all fail at the same time than a single Loki instance. Maybe we should phrase this as `No single point of failure` ?

QuentinBisson: I would counter that having more loki makes risks of failure more likely but with a smaller blast radius

QuentinBisson: So smaller blast radius in case of failure makes sense to me instead of more resilient if you want, but in both solutions we need to think about the failover mechanism because if Loki is down in the middle of an incident, well this will be fun :D

logging-infrastructure/README.md
TheoBrigitte: Yeah good point. How about this ?
```suggestion
- Satisfy customer requirement towards network traffic isolation (i.e. no data leaving their machines or cloud account)
```

logging-infrastructure/README.md
TheoBrigitte: That's a workaround, but it does not change the point that if we introduce a centralized Loki we would move away from having everything in the installation.

logging-infrastructure/README.md
TheoBrigitte: Yes, that's what I mention later on in the conclusion.

logging-infrastructure/README.md
TheoBrigitte: I am not so sure about this, is there such a big difference between running multiple small instance vs a single large one ? I guess we are talking cpu and memory here, but also network traffic is easier when we are at the installation level.

logging-infrastructure/README.md
TheoBrigitte: I kind of added this as a con in the centralized setup under : `Cost needs a new billing system and business model`

logging-infrastructure/README.md
TheoBrigitte: I kind of already added those point under centralized setup pros:

- `Global view (aka single pane of glass, cross-installation queries)`
- `Anomalies detections (multi installation correlated data)`

logging-infrastructure/README.md
TheoBrigitte: I think this relate to your previous point about resource usage

logging-infrastructure/README.md
TheoBrigitte: also relate to your previous point about resource usage

logging-infrastructure/README.md
QuentinBisson: Like it

logging-infrastructure/README.md
QuentinBisson: Yes but:
1: self sufficiency makes sense when they leave, not when they use the product, around which we can build a workaround
2: It depends where you draw the line with the platform. Some see it as the installation level, I prefer to see the platorm at the customer level, but then, why do we have multiple installations for the same customer. This is mostly coming from a customer centric approach (at least that's how I see it)

logging-infrastructure/README.md
QuentinBisson: I'm not sure network traffic is because the traffic would go through the public internet anyway (i.e. ingresses) and we can probably create tunnels otherwise.

I do think this will require more resources because running the loki containers themselves (without taking into account data storage) consumes some resources (we need at least 3 replicas for the read and write path). This might be negligible for sure but having a centralized one would make us use reserved or even better spot instances (probably big) when customers could use bigger machines that would be half empty

QuentinBisson: But we can ignore this point

QuentinBisson: yep

logging-infrastructure/README.md
QuentinBisson: Sure thing but the comments bring more context, or at least I think it does :D

logging-infrastructure/README.md
puja108: Sure, the idea behind adding it is that considerations around the customer can already influence the architecture decision we make here (some of the customer constraints are mentioned below anyway) and we ideally would avoid re-architecting. Especially, if we are thinking about using the same system for customers and ourselves in a multi-tenant model.

logging-infrastructure/README.md
puja108: Not sure we really need failover as logs are not that critical (as long as they don't also cover audit logs) and we don't failover on metrics either, do we?

QuentinBisson: I'm not sure I follow. Why would logs not
be critical? If we provide Loki to customers and they use it for their log searching (especially during incidents), having it unavailable will have them lose trust. We currently do not provide a failover for metrics because customers are not using it (and because individual prometheis cost too much resources) mimir would fix that a bit.

TheoBrigitte: Reworded this section to `Smaller impact in case of failure`
And no we don't failover on metrics

logging-infrastructure/README.md
TheoBrigitte: I get your point, maybe I mixed too many concept in this one. Maybe I just remove this point as we already have the opposite argument in centralized setup as `Customer does not own their data (i.e. lose logs access when leaving GiantSwarm)`

logging-infrastructure/README.md
TheoBrigitte: I'll leave this out then to avoid repetition of arguments.

logging-infrastructure/README.md
TheoBrigitte: To me it feels hard to foresee what will be the differences in terms of resource usage, leaving this out.

logging-infrastructure/README.md
TheoBrigitte: I added part of your comment in a later section as
`Anomalies detections (i.e. detect issues earlier via error/trends measures accross installations/regions/providers)`

logging-infrastructure/README.md
TheoBrigitte: Let's avoid repeating arguments, it will just make this document harder to read IMO

logging-infrastructure/README.md
hervenicol: Regarding federated queries: talked with Loki people today, looks like they have plans for that in the future.
No guarantees though, but yet good news to know they're interested in that.

Regarding "remotewrite" for logs: this feature exists on grafana cloud, no plan to release it open-source for the moment.

logging-infrastructure/README.md
puja108: Honestly, I've rarely seen people HAing their metrics or logs setup, but maybe my view is too narrow. Backing up data, yes, but fully HA rare IMHO. I'm not saying it's not needed, but I'm not sure how high a priority something like that would be compared to other stories we might need working on.

marieroque (approved):

logging-infrastructure/README.md
TheoBrigitte: removed

TheoBrigitte:
@teemow would like to have a look at this before I merge it ?

TheoBrigitte:
I see 4 approvals a no major concerns so I merge this now, happy to discuss and improve this later on when needed.

PR #59, opened 2023-01-18
Add fork management for CAPI+infra controllers

erkanerol:
This RFC can be expanded as a generic policy for fork management of all open-source projects but I don't want to discuss it for months. That is why the scope is limited and it is only for CAPI + infra controllers.

I mostly stated what we hit and what we decided for CAPO and CAPVCD. As far as I see, CAPA/CAPG/CAPI are pretty similar. 

The critical parts:
- Are we OK with the current mechanisms especially repo+branching strategies?
- How are we going to implement the container building pipeline?

### DIFF
--- a/fork-management-for-infra-providers/README.md
+++ b/fork-management-for-infra-providers/README.md
@@ -1,0 +1,148 @@
+# Fork Management for CAPI+Infra controllers
+
+The purpose of this RFC is to specify how we will manage forked repositories for CAPI and infra providers such as CAPVCD, CAPA, CAPO etc.
+
+## Problem Statement
+
+We use open-source Kubernetes operators in our platform and we need to customize them at some points.
+
+- We deploy them via the app platform so we need to bundle them as apps.
+- We have our requirements regarding deployment/configuration specs (e.g. common labels, CRD management etc.) so we need to customize the manifests.
+- We sometimes need to maintain some diffs in the core code since
+  - We need hotfixes and we don't have time to wait for merges in upstream repositories.
+  - We have a pretty specific temporary/permanent requirement and it is not possible to add it to upstream repositories. We must fight for keeping this diff as small/short-term as possible. 
+
+We need to 
+- define a fork management policy including repo management and branching strategies
+- automate code syncs, image building and releases 
+
+## Current status
+
+### Repositories
+
+We have 2 repositories per operator
+1. Fork of the upstream repository 
+- Examples:
+  - https://github.com/giantswarm/cluster-api
+  - https://github.com/giantswarm/cluster-api-provider-aws
+  - https://github.com/giantswarm/cluster-api-provider-cloud-director
+  - https://github.com/giantswarm/cluster-api-provider-gcp
+
+1. A repository for bundling the operator as an app 
+- Naming: https://github.com/giantswarm/\<fork-name\>-app
+- Examples
+  - https://github.com/giantswarm/cluster-api-app
+  - https://github.com/giantswarm/cluster-api-provider-aws-app
+  - https://github.com/giantswarm/cluster-api-provider-cloud-director-app
+  - https://github.com/giantswarm/cluster-api-provider-gcp-app
+
+### Syncing fork repositories
+
+There is no common way.
+
+In CAPVCD, since we have some diffs between our fork and upstream repo, we have a mechanism as follows:
+- `main` branch of the repo is the same as the upstream one. We sync this branch by clicking `Sync fork` button in the UI.
+- `giantswarm-main` branch of the repo contains our changes. We manually rebase it on top of `main` branch when necessary and open a new PR each time. The developer is supposed to resolve conflicts during rebasing.
+ 
+### Bundling + Customization
+
+There are some differences between repositories but the general structure in app repositories is as follows:
+
+- We have Makefile + Bash scripts to
+  - Check out a specific commit/branch of fork/upstream repo. [Example](https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L14)
+  - Generate base manifests by running scripts of the core repo
+  - Customize base manifests with kustomization. [Example](https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L19)
+  - Update manifests under `helm` folder for the app. [Example](https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L28)
+
+When it is necessary to update the app content, developers are supposed to
+- run make targets manually to update manifests and values in the app repo
+- open a PR in the app repo
+
+### Image Management
+
+For some providers, we retag upstream images with retagger.
+For some providers like CAPVCD, we build/push images manually.
+
+## Design proposals
+
+### 1. Repo management
+
+#### 1.a Separate repositories for fork and app
+
+We can have two repositories per operator. One is only fork of upstream and the other one is for only bundling it as an app.
+
+#### 1.b One repository for fork and app
+
+We can maintain bundling and customization logic in the same repo.
+
+#### Decision
+
+- `1.a` is the current situation.
+- `1.b` will lead to many conflicts with the upstream repository and it will be hard to automate conflict resolution.
+
+`1.a` is decided.
+
+---
+
+### 2. Branching strategy in fork repo
+
+#### 2.a main vs giantswarm-main
+
+We can maintain two branches in the fork repo. 
+- `main` will be just the same as `main` of upstream repository. This branch can be synced automatically.
+- `giantswarm-main` will contain our changes if necessary. 
+
+The update flow will be
+- First update `main`
+- Then rebase `giantswarm-main`
+
+#### 2.b only main
+
+We can maintain only one `main` branch including our changes and use `git merge` to apply upstream changes.
+
+#### Decision
+
+`2.a` is decided.
+
+---
+
+### 3. Image building
+
+#### 3.a Not building images 
+
+It is not possible to rely on only retagged versions of containers provided by the upstream community since we want to be able to apply our changes to code base.
+
+#### 3.b Building/Pushing images in the fork repo
+
+We can add an image building/pushing mechanism to fork repository.
+
+- This will increase the diff between fork and upstram repo but it will be only some manifests under `.github` or `.circleci` folders.
+- This will allow developers to verify container building while developing features.
+
+#### 3.c Building/Pushing images in the app repo
+
+We can build/push images while syncing the app repo. That works pretty well while consuming upstream repo but it will make less sense when having our own fork. 
+
+#### Decision
+
+`3.b` is decided.
+
+---
+
+### 4. Where to build the images
+
+#### 4.a Github Actions
+
+#### 4.b CircleCI
+
+#### Decision
+
+Since we build/push images via CircleCI for our other repositories, `4.b` is decided.
+
+---
+
+### 5. Sync logic in the app repo
+
+Because of the decisions in sections 1 and 2, in the app repo, the sync mechanism should be able to consume any commit/branch from fork/upstream repo. 
+See https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L7
+


puja108:
double pinging @yulianedyalkova @AverageMarcus @cornelius-keller @alex-dabija @fiunchinho as you have most stake or prior experience with this from what I know

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+For some providers, we retag upstream images with retagger.
+For some providers like CAPVCD, we build/push images manually.
+
+## Design proposals
+
+### 1. Repo management
+
+#### 1.a Separate repositories for fork and app
+
+We can have two repositories per operator. One is only fork of upstream and the other one is for only bundling it as an app.
```
nprokopic, line 72: This should be the state even now, at least it is for cloud providers.

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### 1.b One repository for fork and app
+
+We can maintain bundling and customization logic in the same repo.
+
+#### Decision
+
+- `1.a` is the current situation.
+- `1.b` will lead to many conflicts with the upstream repository and it will be hard to automate conflict resolution.
```
nprokopic, line 81: To add few comments here.

1. Fork management where we "simply" want to maintain the controller code fork (CAPI, CAPA, CAPZ, etc) with our code changes to the upstream go code is already challenging enough.

2. Additionally, maintaining app repos which are packaging aforementioned controllers into helm charts is also challenging enough.

These two are two separate concerns and we should not mix them.

Having consistent, scalable and robust fork management (1) across different forks (for controller/operator code) for different providers will be challenging, as those are upstream projects that are maintained by different groups of people, so we will encounter differences across providers.

Having consistent, scalable and robust controller/operator packaging into apps (2) will be maybe even more challenging, as for now we have done this across different teams and in different ways, and even if same team was doing it, it wouldn't be simple, as there are differences in upstream projects. This, controller/operator packaging into apps, should be out of scope of this RFC.

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+---
+
+### 2. Branching strategy in fork repo
+
+#### 2.a main vs giantswarm-main
+
+We can maintain two branches in the fork repo. 
+- `main` will be just the same as `main` of upstream repository. This branch can be synced automatically.
+- `giantswarm-main` will contain our changes if necessary.
```
nprokopic, line 93: What is the value of having a branch which is upstream `main` with our changes, what would we use it for?

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+We use open-source Kubernetes operators in our platform and we need to customize them at some points.
+
+- We deploy them via the app platform so we need to bundle them as apps.
+- We have our requirements regarding deployment/configuration specs (e.g. common labels, CRD management etc.) so we need to customize the manifests.
+- We sometimes need to maintain some diffs in the core code since
+  - We need hotfixes and we don't have time to wait for merges in upstream repositories.
+  - We have a pretty specific temporary/permanent requirement and it is not possible to add it to upstream repositories. We must fight for keeping this diff as small/short-term as possible. 
+
+We need to 
+- define a fork management policy including repo management and branching strategies
```
vxav, line 16: Plus sync/rebase schedule/conditions?

fork-management-for-infra-providers/README.md
nprokopic: From upstream docs https://github.com/kubernetes-sigs/cluster-api/blob/main/CONTRIBUTING.md#branches

> Cluster API has two types of branches: the main branch and release-X branches.
>
> The main branch is where development happens. All the latest and greatest code, including breaking changes, happens on main.
>
> **The release-X branches contain stable, backwards compatible code.** On every major or minor release, a new branch is created. It is from these branches that minor and patch releases are tagged. In some cases, it may be necessary to open PRs for bugfixes directly against stable branches, but this should generally not be the case.

We should never deploy a version of controller that is coming from upstream main branch, because it is not considered stable, it is latest, greatest and often breaking code. Therefore we should really not care about maintaining our changes in some main branch in our fork (named `giantswarm-main` here), as it would be additional maintenance work.

The changes in our fork, when we really must implement them, should always come on top of stable upstream code, which means that we would do PRs to release branches directly.

If some custom change in our fork gets to be long(er) lived (we should not have many of those, but about that in another comment) because it takes more time to get it released in upstream, and there is a new release, meaning there is a new release branch, then those custom changes would be cherry-picked from old release branch to a newer release branch.

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### 4.b CircleCI
+
+#### Decision
+
+Since we build/push images via CircleCI for our other repositories, `4.b` is decided.
+
+---
+
+### 5. Sync logic in the app repo
```
nprokopic, line 144: Syncing upstream changes should be fully automated and require no human intervention, unless we have our changes and there are conflicts.

nprokopic:
Fork management is a technical details that should be as simple as possible and fully automated, otherwise it will be a maintenance burden and a waste of time as it has been for the past 2-3 years.

There are few very important things that are not mentioned, that are bigger concern, and that are affecting how we do fork management:
- What types of changes do we want in our forks?
- Do we always default to writing code in our forks first?

## 1. Before writing code

For every change that we need in upstream controller code, we first talk to upstream project maintainers, in Slack, in upstream office hours, in GitHub issues, PRs, etc.

The only exception being something like urgent security issues and obvious critical bugs.

## 2. The type of work that we do in our forks

Any work we do in our forks should be just temporary, something we maintain until it is merged and released in upstream, or until it is removed from our forks because we don't need it anymore.

This would be:
  - CircleCI and similar config
  - Bug fixes
  - Security fixes (kind of subset of the above, but worth mentioning separately)
  - Any kind of feature work that is also going to be merged in upstream (details below).
  - Very small feature work that will be removed in not too distant future because it is something we need temporarily (so it's fine-ish to maintain it for some time).

### Feature work that is also going to be merged in upstream

Where and how we do feature work is important. Here are some comments from my past experience in cloud KaaS, some dating back to Team Celestial as that is when we started forking upstream repos (yeah this is that old topic) and when we started using customized CAPI/CAPZ (API changes only, without modifying controllers as we didn't run them).

What we should do:

- Feature works happens on a branch that is branched off the latest release branch. The latest release branch contains the latest stable code that we are running, so we care about that the most.
- We create two pull requests that will enable us to get our work in upstream as fast as possible, while having it in our fork right away:
  - One from our feature branch to the release branch in our fork.
  - One from our feature branch to the main branch in upstream (because upstream is first merging into `main` for the purposes of the next release, and then potentially cherry picking into current or some older release branch).

With the above in mind, we don't have a need to maintain our changes in main branch of the fork (or some clone of upstream `main` like mentioned `giantswarm-main` with our changes added).

**What we should not do**:

- Major or any remotely bigger feature work in our fork that will not end up in upstream. This should really be a written down and documented as a big no.
  - If some feature cannot be accepted and implemented in upstream CAPI, then it should not be part of our fork because we treat our forks as a bit faster-to-release-and-deploy upstream. Instead, those features should be developed in a separate operator.
  - If that feature work requires CAPI/CAPx code changes (cannot be done just in a separate operator) because something in upstream code is conflicting with what we need, and at the same time we cannot eventually get those changes in upstream, then we should rethink our approach and look for a different solution, because CAPI/CAPx is the technical foundation of our product and we should not be fighting against it.
- Maintaining long-lived features that we have only in our fork and that are not part of the upstream. We surely can cherry-pick from old to new releases our custom changes, but if we do that for too long and too many times, at some point we will start getting more and more conflicts and it will become harder and burden to maintain.

## Conclusion

As describe above, when it comes to changes in our forks, all of them would be basically temporary, because those changes would be either merged in upstream or discarded entirely once they are not needed.

Upstream `main` branch is never released, it's a development branch. Since our forks exists for reasons different than regular development, and we would do custom changes in forks for the sake of velocity, we should not care about the `main` branch and we should just maintain custom changes on release branches.

Once newer upstream release gets the changes we need, we simply stop maintaining that change in our release branches and we don't cherry-pick to changes to he latest release branch because they are already there.

All custom changes in our forks should be considered as items in an Inbox, and we should strive to have Inbox Zero most of the time (with exception of CircleCI and similar config).

JosephSalisbury:
cc @JosephSalisbury

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+In CAPVCD, since we have some diffs between our fork and upstream repo, we have a mechanism as follows:
+- `main` branch of the repo is the same as the upstream one. We sync this branch by clicking `Sync fork` button in the UI.
+- `giantswarm-main` branch of the repo contains our changes. We manually rebase it on top of `main` branch when necessary and open a new PR each time. The developer is supposed to resolve conflicts during rebasing.
+ 
+### Bundling + Customization
+
+There are some differences between repositories but the general structure in app repositories is as follows:
+
+- We have Makefile + Bash scripts to
+  - Check out a specific commit/branch of fork/upstream repo. [Example](https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L14)
```
piontec, line 52: We have a similar problem when updating any external chart from upstream. We're adapting git to solve all our needs, including git's subtree where needed. That way we don't need any synchronization scripts, as every change we do is saved in git itself.

We have docs about how it works here: https://handbook.giantswarm.io/product/managed-apps/dev-experience/git-subtree/

I also did a few demos that show how it works in real life:
- https://drive.google.com/file/d/1T9j2-sk3CEdp1-7_Q00VHj3s11eK6_jG/view?usp=share_link
- https://drive.google.com/file/d/11BWAOIm6Mr04DRPf-E4RHf_y7QSee5f8/view?usp=share_link
- https://drive.google.com/file/d/1xb8MGhv5OiBgwL5KMCZCRG3cw59xYOtT/view?usp=share_link

piontec (commented):
I'm dropping info about how we work (and recommend to work) with upstream helm charts. As it's pure git, it applies to any repo.

Also, please note, that if there're decisions already, this doc is probably an ADR, not just RFC.

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+  - https://github.com/giantswarm/cluster-api-provider-gcp-app
+
+### Syncing fork repositories
+
+There is no common way.
+
+In CAPVCD, since we have some diffs between our fork and upstream repo, we have a mechanism as follows:
+- `main` branch of the repo is the same as the upstream one. We sync this branch by clicking `Sync fork` button in the UI.
+- `giantswarm-main` branch of the repo contains our changes. We manually rebase it on top of `main` branch when necessary and open a new PR each time. The developer is supposed to resolve conflicts during rebasing.
+
```
AverageMarcus, line 46: Could you please include examples of some of the other ways.

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+We have 2 repositories per operator
+1. Fork of the upstream repository 
+- Examples:
+  - https://github.com/giantswarm/cluster-api
+  - https://github.com/giantswarm/cluster-api-provider-aws
+  - https://github.com/giantswarm/cluster-api-provider-cloud-director
+  - https://github.com/giantswarm/cluster-api-provider-gcp
+
+1. A repository for bundling the operator as an app 
+- Naming: https://github.com/giantswarm/\<fork-name\>-app
```
AverageMarcus, line 32:
```suggestion
2. A repository for bundling the operator as an app 
- Naming: https://github.com/giantswarm/<fork-name>-app
```

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+  - Update manifests under `helm` folder for the app. [Example](https://github.com/giantswarm/cluster-api-provider-cloud-director-app/blob/adfc74115d9b7aeda60d885500122be3fb434d2e/Makefile.custom.mk#L28)
+
+When it is necessary to update the app content, developers are supposed to
+- run make targets manually to update manifests and values in the app repo
+- open a PR in the app repo
+
+### Image Management
+
+For some providers, we retag upstream images with retagger.
+For some providers like CAPVCD, we build/push images manually.
```
AverageMarcus, line 64: Why is CAPVCD different?

fork-management-for-infra-providers/README.md
```diff
@@ -0,0 +1,148 @@
...
+- First update `main`
+- Then rebase `giantswarm-main`
+
+#### 2.b only main
+
+We can maintain only one `main` branch including our changes and use `git merge` to apply upstream changes.
+
+#### Decision
+
+`2.a` is decided.
```
AverageMarcus, line 105: Who decided this? 

This wont work for CAPA, for example, as the `main` branch isn't a release branch. For CAPA we're based on the `release-1.5` branch until we have the bandwidth to upgrade to the next `release-xxx` branch but we would never release from a `main` branch.

AverageMarcus:
(Whoops, accidentally closed the PR 😅 )

I completely agree with all the comments that Nikola has added. They all resonate pretty well with my thoughts.

I just want to add a few additional / more specific comments:

* I don't think we should include notes about the `xxx-app` repos / charts in this. That's not really related to fork management and is already a process we have pretty well laid out. The only exception to that is around the CRDs from upstream but those aren't called out specifically in this.
* I think this doc is too focussed on technical process and doesn't include the community aspect of fork management. Arguably, the technical problem is easy compared to the community side. Similar to what Nikola said we need to make sure our process first starts out with upstream work. This needs to be **at least** an issue covering the problem in the upstream repo but ideally should also be a discussion directly with the project maintainers on either Slack or office hours. We should then follow the upstream contributing guides to submit our changes to upstream and only then should the process of including it in our fork take over.
* I think we need to include something about what to do when our fork isn't going to sync with upstream (because they disagree with an approach)
* I'd like to see the examples / current approaches acknowledge how other providers are being managed (CAPA, CAPG, CAPZ) as this document is currently too CAPVCD focussed.
* And as @piontec said "that if there're decisions already, this doc is probably an ADR, not just RFC." Including decisions in this document means we can't suggest alternatives or improvements as part of this process. My understanding is the decisions should come _after_ the RFC is discussed and is then transferred to an ADR.

fork-management-for-infra-providers/README.md
erkanerol: Since we have some diffs between our fork and the upstream repo.

fork-management-for-infra-providers/README.md
erkanerol: Yep. I stated it in the Decision section.

fork-management-for-infra-providers/README.md
cornelius-keller: In theory I would agree that our forks should always come out of an  upstream release branch. However this is against the experience we had with CAPO. Upstream was not releasing for many month. We even had several PRs merged into it, but it was not released for a very long time. So upstream `main` was already `giantswarm-main` in a way. 
I don`t think backporting all this changes to a released version would be worth the effort.

cornelius-keller: I think it should be up to the PEs to decide which branch is the most stable and best base for a fork.

AverageMarcus: But wouldn't those unreleased changed go into a `release-(n+1)` branch?

For example, in CAPA that have a `release-2.0` and a `release-1.5`. The `release-2.0` is what new releases are tagged off of, so `v2.0.2` being the latest from that branch, and the `release-1.5` is for any backported features with `v1.5.2` being the latest release from that branch. Once new breaking changes are planned a `release-2.1` (or whatever) branch will be created first. This is the approach I've seen for at least CAPI & CAPA and I think CAPZ. 

If the upstream processes change that much between providers we then need to call that out in this and plan accordingly.

fork-management-for-infra-providers/README.md
erkanerol: @piontec do we build images in fork repositories in this structure or do we consume the upstream images directly? I didn't see anything about image building.

fiunchinho: > We have a similar problem when updating any external chart from upstream. We're adapting git to solve all our needs, including git's subtree where needed. That way we don't need any synchronization scripts, as every change we do is saved in git itself.
> 
> We have docs about how it works here: https://handbook.giantswarm.io/product/managed-apps/dev-experience/git-subtree/
> 
> I also did a few demos that show how it works in real life:
> 
> * https://drive.google.com/file/d/1T9j2-sk3CEdp1-7_Q00VHj3s11eK6_jG/view?usp=share_link
> * https://drive.google.com/file/d/11BWAOIm6Mr04DRPf-E4RHf_y7QSee5f8/view?usp=share_link
> * https://drive.google.com/file/d/1xb8MGhv5OiBgwL5KMCZCRG3cw59xYOtT/view?usp=share_link

isn't that a different scenario? There are no upstream Helm charts for CAPx.

fork-management-for-infra-providers/README.md
nprokopic: > but we would never release from a main branch

We should never release from `main` branch in any provider, at least not according to upstream branching/versioning strategy where releases are coming only from `release-xxx` branches.

fork-management-for-infra-providers/README.md
nprokopic: > I think it should be up to the PEs to decide which branch is the most stable and best base for a fork.

This is documented in upstream projects, and unless some other projects are for some reason different, the stable branches used for releases are `release-xxx` branches, and `main` is a development branch.

nprokopic: > However this is against the experience we had with CAPO. Upstream was not releasing for many month. We even had several PRs merged into it, but it was not released for a very long time. So upstream main was already giantswarm-main in a way.

This sounds like an exception, as it seems that a large amount of work, features and fixes got accumulated in `main` branch that was not released at all, so backporting that to the older version would not make sense in that case, as new release should be released.

Relatively active upstream CAPI projects do not have this issue, I think this is in exception rather than a rule.

In this case we should push upstream to cut a release, and if the project is inactive/unmaintained that much in general to the point that we are always waiting too much for new minor release (to get those accumulated features released), we should get involved in upstream project more directly and get somebody from Giant Swarm to become reviewer/maintainer and drive the project forward. We already have folks that became reviewers/maintainers in other upstream projects.

nprokopic: > This sounds like an exception, as it seems that a large amount of work, features and fixes got accumulated in main branch that was not released at all, so backporting that to the older version would not make sense in that case, as new release should be released.

In this case, where new release has not been cut for that long, the `main` branch state is probably what the latest `release-xxx` branch should be, but the project is inactive/slow, so that work is not done.

erkanerol:
This PR is moved to https://github.com/giantswarm/giantswarm/pull/25779. Please contribute that PR.

PR #60, opened 2023-01-24
add logging-infrastructure/promtail-config-management

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/1903

Adding description and details regarding how we will handle the Promtail configuration.


logging-infrastructure/promtail-config-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Architecture
+
+Here is a schema describing how the Promtail configuration is managed :
+
+![promtail-config-management](./assets/promtail-config-management.svg)
+
+We have 3 distinct source of truth :
+
+* `scrape_config`
```
puja108, line 23: This looks to be your interface to other teams. Not sure if as part of this RFC or a separate one you want to suggest how this interface is defined, managed, and if it fulfills the use cases the other teams (and ideally at some point similarly customers) have. Interesting in that context would be if it's some kind of standard upstream interface, so we have a minimum amount of specificity and can onboard users quickly (and with upstream docs).

This might be a topic that can be brought to SIG architecture for example, as I think that's where we want to agree on inter-team interfaces.

TheoBrigitte: Yes good point, I had similar thoughts towards this afterwards. The `scrape_config` is the interface to plug-in application logs to our Loki. This need to be clearly defined. Will revisit this then.

QuantumEnigmaa (approved):
I think the graph alone explains most of the architecture and the md file is also well redacted :)

logging-infrastructure/promtail-config-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+As a Giant Swarm engineer, I want to be able to deploy configuration for ingesting logs to all our cluster in a unified way.
+
+## Context
+
+Our logging infrastructure will rely on [Promtail](https://grafana.com/docs/loki/latest/clients/promtail/) to ingest logs. Promtail act as agent which collect logs locally from within the cluster and ship them to an external component for storage and processing.
+
+After deploying Promtail we need to control 3 elements :
+
+- The `scrape_config`, which describe how Promtail will discover targets to collect logs from.
```
hervenicol, line 11:
```suggestion
- The `scrape_config`, which describes how Promtail will discover targets to collect logs from.
```

logging-infrastructure/promtail-config-management.md
```diff
@@ -0,0 +1,43 @@
...
+   - In the Grafana Loki datasource: to read logs from Loki
+
+* toggle flag
+
+  The toggle flag is defined in the giantwarm/config repository, which give us the ability to enable and disable log ingestion per installation.
+
+
+## Limitation
+
+* Using the same `scrape_config` for all cluster regardless of provider and version. This has the benefit for us to deploy update easily. Also this should not be a problem regarding target discovery, learning from Promtheus scrape config we should be able to define a clear interface.
```
hervenicol, line 41:
```suggestion
* Using the same `scrape_config` for all installations/clusters regardless of provider and version. This has the benefit for us to deploy update easily. Also this should not be a problem regarding target discovery, learning from Prometheus scrape config we should be able to define a clear interface.
```

hervenicol (approved):
Very nice!

logging-infrastructure/promtail-config-management.md
puja108: fine by me to have that separate, I only wanted to point it out as we should not be waiting too long with defining interfaces and thinking about the end user use cases that we can cover, as we need to iterate more with end users outside the team so you get feedback earlier.

QuentinBisson: Sure @puja108 my comment was mostly for Theo so that he does not get into too much detail and maybe postpone work on Promtail until we have a proper interface (which would eventually change anyway) defined :)

I do agree we need to think it through eventually and having an operator to manage the promtail config will help for this. We can think about some kind of CR that let the customer configure the applications they want to have their logs shipped (could be opt-in/out by default, maybe defined at the installation level).

Our first plan is to get the giant swarm logs first abd this would be in the defautl config. The controller would be able to override that config anyway the same way the prometheus-operator renders the promtheus configuratoin from all the servicemonitors on a cluster :)

PR #61, opened 2023-01-25
rfc about crossplane's access to infrastructure

piontec:
This is written from my perspective - and I don't have much knowledge about how we bootstrap and manage underlying infrastructure. As such, more ideas are very welcome :)

### DIFF
--- a/crossplane-infra-access/README.md
+++ b/crossplane-infra-access/README.md
@@ -47,3 +47,12 @@ To limit the problematic side of creating access tokens with cloud provider's IA
 
 This is exactly like 2., but we start with a limited set of service allowed (right now we know that we need S3, so we include S3 only). When more services are required, we add more permissions to already existing token.
 
+### 4. IRSA + Role claims
+
+According to [docs](https://github.com/upbound/provider-aws/blob/main/AUTHENTICATION.md) we should be able to mix IRSA and assumed Web Identity providers. In this solution, `Crossplane` controllers run with a ServiceAccount that is authenticated by IRSA and authorized in IAM to claim other Roles in IAM. Then, we and customers can create multiple `ProviderConfig`s CRs that reference different IAM roles that `aws-provider` controller is able to claim using IRSA. This seems like a preferred solution, but has to be tested.
+
+Problems/questions:
+- This doesn't solve a use case, where clients (software) access to AWS infrastructure created by Crossplane needs to be based on access keys (we have this issue with Harbor). This can be worked-around by creating an IAM role, user and access keys from Crossplane, in the same go as the needed infrastructure, but this seems to be a complex and error prone solution.
+- `aws-provider` controller still runs with a single SA (obviously). This means, that the IAM Role it runs with has to have permissions to claim all other needed IAM roles, including these created by a customer. And we don't want to edit our Role every time a customeror we adds another to-be-claimed Role.
+
+
--- a/crossplane-infra-access/README.md
+++ b/crossplane-infra-access/README.md
@@ -34,8 +34,7 @@ Specific questions in regard to infrastructure provider access tokens:
 
 ### 1. No default access token
 
-As it's hard to create a limited yet elastic set of access permissions to the infrastructure, we provide no default config. Every user that will need to do something will have to manage 
-access tokens on its own. Sample workflow: a user that needs access to S3, needs to create AWS Access Keys on each MC. Then, the user needs to encrypt them and add to the [management clusters fleet repo](https://github.com/giantswarm/management-clusters-fleet/) in `sops` encrypted form. As a result, access tokens are delivered to every MC and RBAC permissions to them can be handled easy as a common RBAC configuration delivered by Flux to every relevant MC.
+As it's hard to create a limited yet elastic set of access permissions to the infrastructure, we provide no default config. Every user that will need to do something will have to manage access tokens on its own. Sample workflow: a user that needs access to S3, needs to create AWS Access Keys on each MC. Then, the user needs to encrypt them and add to the [management clusters fleet repo](https://github.com/giantswarm/management-clusters-fleet/) in `sops` encrypted form. As a result, access tokens are delivered to every MC and RBAC permissions to them can be handled easily as a common RBAC configuration delivered by Flux to every relevant MC.
 
 Questions:
 - Is it possible/realistic to automate creation of these access keys (same set of permissions, but on different infra provider accounts)? What about different providers (AWS vs Azure vs ...)
--- a/crossplane-infra-access/README.md
+++ b/crossplane-infra-access/README.md
@@ -1,0 +1,50 @@
+# Crossplane infrastructure access
+
+## Intro
+
+Team Honeybadger is deploying Crossplane to MCs. Crossplane can be used to manage infrastructure on the cloud provider side by both us and out customers.
+
+In order to do that, Crossplane has to be able to authenticate and authorize against the infrastructure provider. The access model that Crossplane uses depends on the user. By default, Crossplane doesn't require nor provide any access info. The administrator of Crossplane can create as many access secrets as he/she wants and can grant access to them to other Kubernetes users using the standard RBAC configuration.
+
+Example:
+1. Crossplane admin logins to an AWS account. In IAM, it creates two roles:
+  - `s3-access` that grants permissions to use S3 only,
+  - `db-access` that allows usage of RDS,
+2. For both roles, the admin creates access keys. Then, access keys are saved as secrets in the kubernetes API.
+3. The admin creates 2 `ProviderConfig` custom resources, named `s3` and `rds`, referencing respective Secrets with AWS API keys. Access to `ProviderConfig` is restricted.
+4. The admin grants permissions with RBAC:
+  - he/she allows access to `rds` `ProviderConfig` to the `db-automation` ServiceAccount in the `devel` Namespace
+  - he/she allows access to the `s3` `ProviderConfig` to the `customer-admins` Group
+5. From now on, respective SA and users belonging to the group can request Crossplane to manage (CRUD) resources in AWS.
+
+## Problem
+
+We want to configure all our MCs so that it's easy to use Crossplane by our internal (Giant Swarm) users.
+
+Example:
+1. We want to deploy Harbor to all CAPA MCs. Harbor needs an S3 bucket to work. How can we pre-configure MCs, so that it is possible to create the necessary buckets as part of Harbor App deployment?
+
+Specific questions in regard to infrastructure provider access tokens:
+
+1. Is it possible to easily mass-create a set of access credentials across multiple/all MCs?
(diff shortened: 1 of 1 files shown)


crossplane-infra-access/README.md
```diff
@@ -0,0 +1,50 @@
+# Crossplane infrastructure access
+
+## Intro
+
+Team Honeybadger is deploying Crossplane to MCs. Crossplane can be used to manage infrastructure on the cloud provider side by both us and out customers.
+
+In order to do that, Crossplane has to be able to authenticate and authorize against the infrastructure provider. The access model that Crossplane uses depends on the user. By default, Crossplane doesn't require nor provide any access info. The administrator of Crossplane can create as many access secrets as he/she wants and can grant access to them to other Kubernetes users using the standard RBAC configuration.
+
+Example:
+1. Crossplane admin logins to an AWS account. In IAM, it creates two roles:
```
calvix, line 10: Who is Crossplane admin? a person or a tool/operator?

crossplane-infra-access/README.md
```diff
@@ -0,0 +1,50 @@
...
+  - `s3-access` that grants permissions to use S3 only,
+  - `db-access` that allows usage of RDS,
+2. For both roles, the admin creates access keys. Then, access keys are saved as secrets in the kubernetes API.
```
calvix, line 13: I think that we want to move away from any manually generated access keys as much as possible, as the general approach in AWS itself (the guidelines) and also our customer requirements are often that we do not use accesskeys but rather have roles and assume the role.

For that case we deeply IRSA which can leverage assuming specific roles based on the pod needs

calvix:
IMHO the general answer to it is to leverage IRSA  on the CAPA mcs

crossplane-infra-access/README.md
whites11: +1 no hardcoded secrets ever

crossplane-infra-access/README.md
uvegla: I think a person is meant here. Could be GS staff or customer admins too. We need to manage the scope of the access keys / tokens to avoid exposing full access in case of an issue with the secrets.

crossplane-infra-access/README.md
uvegla: + 1 for not hardcoding secrets / using access keys.

For reference, the community AWS provider's provider config (the thing used to let the provider controller access to the cloud provider for a given resource (as this can be different for every resource / composition / whatever)): https://doc.crds.dev/github.com/crossplane/provider-aws/aws.crossplane.io/ProviderConfig/v1beta1@v0.31.0

It can assume a given role / web identity, but a secret is needed that has some sort of credentials / tokens.

Something maybe that could regenerate this token periodically would be nice. As of now I don't know if IRSA could work here or now as I don't know it very well.

uvegla: IRSA could work, according to: https://github.com/crossplane-contrib/provider-aws/blob/bf4088121aa87877cb969c5de3e1b287d6f853ca/AUTHENTICATION.md#using-iam-roles-for-serviceaccounts

pipo02mix: but IIRC for assuming a role you have to be an AWS user (so credentials for origin account). Like our aws operator, it has user credentials for our base AWS account and it assumes the  roles in our customer accounts, right?

crossplane-infra-access/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Problem
+
+We want to configure all our MCs so that it's easy to use Crossplane by our internal (Giant Swarm) users.
+
+Example:
+1. We want to deploy Harbor to all CAPA MCs. Harbor needs an S3 bucket to work. How can we pre-configure MCs, so that it is possible to create the necessary buckets as part of Harbor App deployment?
+
+Specific questions in regard to infrastructure provider access tokens:
+
+1. Is it possible to easily mass-create a set of access credentials across multiple/all MCs?
```
pipo02mix, line 29: No, we need permissions to create IAM roles and policies. I would think more into a script or template that customer can use. For AWS, we have [TF script](https://github.com/giantswarm/giantswarm-aws-account-prerequisites) that helps with that

crossplane-infra-access/README.md
```diff
@@ -0,0 +1,50 @@
...
+2. What is a reasonable access scope of a default infrastructure access token?
+3. Should there be a default infrastructure access token on each MC at all?
```
pipo02mix, line 31: I think this is a good idea, define the scope based on use case. If we need certain access across some AWS services then have role with them and assume from our infrastructure provider SA

crossplane-infra-access/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Possible solutions
+
+### 1. No default access token
+
+As it's hard to create a limited yet elastic set of access permissions to the infrastructure, we provide no default config. Every user that will need to do something will have to manage 
+access tokens on its own. Sample workflow: a user that needs access to S3, needs to create AWS Access Keys on each MC. Then, the user needs to encrypt them and add to the [management clusters fleet repo](https://github.com/giantswarm/management-clusters-fleet/) in `sops` encrypted form. As a result, access tokens are delivered to every MC and RBAC permissions to them can be handled easy as a common RBAC configuration delivered by Flux to every relevant MC.
```
pipo02mix, line 38:
```suggestion
As it's hard to create a limited yet elastic set of access permissions to the infrastructure, we provide no default config. Every user that will need to do something will have to manage access tokens on its own. Sample workflow: a user that needs access to S3, needs to create AWS Access Keys on each MC. Then, the user needs to encrypt them and add to the [management clusters fleet repo](https://github.com/giantswarm/management-clusters-fleet/) in `sops` encrypted form. As a result, access tokens are delivered to every MC and RBAC permissions to them can be handled easily as a common RBAC configuration delivered by Flux to every relevant MC.
```

crossplane-infra-access/README.md
calvix: No,  we don't need, IRSA does it without the need for credentials, similar to what KIAM does

calvix: https://github.com/crossplane-contrib/provider-aws/blob/bf4088121aa87877cb969c5de3e1b287d6f853ca/AUTHENTICATION.md#using-iam-roles-for-serviceaccounts

looks like it should work

calvix: > However the problem must be solved for other cloud providers as well. We support AWS, GCP and Azure right now with Crossplane.

All providers have similar functionality based on the cloud roles so this should be similar solution across all providers

calvix: and FYI we have a customer who doesn't allow the use and manual creation of any static Access/Secret keys and we are only allowed to use role assuming so anyway  any approach not leveraging IRSA will not work there and will be denied

crossplane-infra-access/README.md
piontec: Yes, that's a person that is responsible for configuring crossplane for users, either internal or external

crossplane-infra-access/README.md
piontec: The problem is that IRSA is not a good solution as well: it can use only a single kubernetes Role. More precisely, we can bind the SA that `aws-provider` runs with with an IAM role, but it has to be a single SA and thus a single IAM role. We can't make fine-grained separate `ProviderConfigs`. This means that this IAM role we bind with IRSA needs to be almost "aws admin", or at least have all the required permissions for a given MC.

piontec:
> IMHO the general answer to it is to leverage IRSA on the CAPA mcs and maybe add have some operator to create the required roles for the crossplane/harbor for MCs

Like explained above, IRSA is just another (though better) form of "single access secret for everything".

crossplane-infra-access/README.md
puja108: AFAIK IRSA (as the name suggests is access automation for Service Accounts) and those SAs can then be used separately (in form of secrets for example), they do not need to be part of the deployment, even if that's the usual process.

crossplane-infra-access/README.md
puja108: We should extend our role creation scripts to add additional roles for the things we need access to through Crossplane

crossplane-infra-access/README.md
puja108: like I mentioned above, single use case bounded roles will be best so we can make them defaults and also remove them again

crossplane-infra-access/README.md
```diff
@@ -0,0 +1,49 @@
...
+- Is it possible/realistic to automate creation of these access keys (same set of permissions, but on different infra provider accounts)? What about different providers (AWS vs Azure vs ...)
+
+### 2. "Reasonable" access token by default
+
+To limit the problematic side of creating access tokens with cloud provider's IAM solutions, we make a fire-once effort of creating a set of roles on each MCs' provider. That access token will have a 'reasonable' set of services permitted (for example S3, databases, logs, ...). Token will be saved in management clusters fleet repo and delivered to MCs. RBAC will be configured so that only users in the `giantswarm-admin` Group can access it.
+
+### 3. Minimal default access token
+
+This is exactly like 2., but we start with a limited set of service allowed (right now we know that we need S3, so we include S3 only). When more services are required, we add more permissions to already existing token.
+
```
puja108, line 49: 4. default access token per GS use case

puja108 (commented):
Is the goal of this RFC limited to the GS use case of defaults? If yes, we should have docs and/or recommendations how customers will do this for themselves (slightly dependent on this, as if they can create roles and use them with IRSA in the MC then it looks very different from actually having to bring over tokens using sth like ESO)

piontec:
> Is the goal of this RFC limited to the GS use case of defaults? If yes, we should have docs and/or recommendations how customers will do this for themselves (slightly dependent on this, as if they can create roles and use them with IRSA in the MC then it looks very different from actually having to bring over tokens using sth like ESO)

No, this applies to customers as well. We're providing only a single crossplane deployment on an MC and we want to use corssplane's ability to use different ProviderConfigs to separate our access from customers. So, the solution has to be consistent.

piontec:
I updated the RFC with idea no 4, based on my recent discussion with @calvix

piontec:
Keeping it open until crossplane resolves performance issues and we can actually start recommending solutions to our users.

pipo02mix:
But why crossplane performance affects to how to access provider infra? what if we resolve the discussions and merge it in this state?

piontec:
The solution that has to be implemented is actually described here: https://github.com/giantswarm/crossplane-examples/tree/main/aws-provider/s3-with-web-id

But: because of the big changes crossplane has in progress (related to performance), I thought to keep this open until we can validate that everything works the same after they do provider-separation changes.

Still, if you feel this is fine for now and we can create a new ticket if we have to - feel free to close this one.

piontec:
Done, exaples included in https://github.com/giantswarm/crossplane-examples/ repo

PR #62, opened 2023-02-03
Application ownership models

nprokopic:
When it comes to application ownership, there is no one-size-fits-all, there is no silver bullet for all apps.

This RFC defines different application ownership models, trying to come up with a clear and robust, yet flexible framework that works for all apps and all teams, which can also ease how we structure teams and share ownership of the apps across them.

TL;DR - sections 2.2 and 3 are the main reason for writing this :)

### DIFF
--- a/application-ownership-models/README.md
+++ b/application-ownership-models/README.md
@@ -1,0 +1,116 @@
+# Application ownership models
+
+A lifecycle of an Kubernetes application could be roughly split into the following stages and sub-stages:
+- Development
+- Release engineering, which could be broken down to:
+  - Building
+  - Testing
+  - Packaging
+  - Delivery (deployment, upgrades, removal)
+- Operations, which could be broken down to:
+  - Investigating issues
+  - Improving performance
+  - On-call
+  - Postmortems
+  - Root cause analysis
+
+Additionally, while for the most of applications it's obvious which team can and should take care of its lifecycle, for some apps that is not very clear, as multiple teams have equally important stakes in it.
+
+Based on the above, we can define two different criteria for defining application ownership models:
+- What stages of the application lifecycle are owned by a team?
+- Is there one or more Giant Swarm teams that can be considered owners of the application?
+
+## 1. Application lifecycle ownership
+
+We can define two high-level application ownership models:
+- Full application ownership
+- Partial application ownership, usually including release engineering (full or partial) and operations. This is typical for upstream applications where the development is done outside of Giant Swarm.
+
+### 1.1. Full application lifecycle ownership
+
+(Almost) Full application ownership includes all stages of the application lifecycle.
+
+Development is done by team members.
+
+Release engineering is done by team members who are using RelEng platform provided by Giant Swarm RelEng teams.
+
+Operations are handled by the team members, including business hours as a part of team on-call, and non-business hours as a part of team/area on-call. There are few caveats here:
+- The non-business hours on-call is currently shared with other teams from the same area (hence "almost" full ownership), which is done for the sake of not overwhelming on-callers with frequent night shifts.
+- Ideally, if a team has enough people and we have state-of-the-art monitoring and alerting, meaning that there are next-to-none false positive alerts (for example caused by alerts from other team's components), the team can own 24h on-call for the application.
+
+### 1.2. Partial application lifecycle ownership
+
+Partial application ownership means that application development is done outside of Giant Swarm, therefore this is the case of upstream applications that we are deploying to our clusters.
+
+To name a few examples, we run, but do not develop, Prometheus, Grafana, Cluster API controllers, Kyverno, different CNI, CSI, CPI apps, then core Kubernetes components, such as API server, scheduler, controller manager, etcd, kubelet, etc.
+
+Partial application ownership of these components means that a Giant Swarm team owns:
+- Release engineering, fully or partially;
+- Operations, fully or partially, depending if the team has sole or shared ownership of the application, which is discussed in the next section.
+
+**Release engineering is owned fully** in case of forked projects, where Giant Swarm team builds the application from the forked repository, then tests, packages and finally delivers it.
+
+**Release engineering is owned partially** when Giant Swarm team does not build the app, but, for example, re-tagged container image is used. Here we have two cases:
+- Giant Swarm team is packaging app as a Helm chart in the app repository, by building required Helm templates for deploying re-tagged container image and optionally using existing upstream Kubernetes manifests instead of writing Helm templates from scratch.
+- Giant Swarm is repackaging or importing existing upstream Helm charts into Giant Swarm application repositories.
+
+## 2. Sole or shared application ownership
+
+While different application ownership models based on application lifecycle stages are relatively clear and it's obvious which model is used for every application, for some apps is much less evident which team should own them, and if there can even be just one owning team, or multiple teams cannot be avoided due to shared nature of the app.
+
+Additionally, while application lifecycle stages split the app ownership between external (upstream) and internal (Giant Swarm) teams, sole or shared ownership is about how we handle app ownership internally within Giant Swarm.
+
+### 2.1. Sole application ownership
+
+The is the classical and usually the obvious case. A Giant Swarm team develops the application, in case of full application lifecycle ownership described in the previous section, or the team takes care of release engineering and operations for the app, in case of partial application lifecycle ownership described above.
+
+### 2.2. Shared application ownership
+
+And this is the tricky part, one of the main reasons for which this whole RFC is written.
+
+Shared application ownership applies in some cases of partial application lifecycle ownership, so in a nutshell, the question here is - which Giant Swarm team owns some upstream application? Can it even be one team, or multiple teams must share ownership? If multiple teams share ownership, how does that work in practice?
+
+There is no silver bullet for all upstream applications, so here we need clear and robust, but also flexible ownership model.
+
+Before trying to come up with some definition, let's first check few examples.
+
+#### 2.2.1. Examples of shared applications
+
+While there are more examples, let's take the following two that should paint the picture and describe the issue of shared ownership clearly enough.
+
+**Cluster API provider-independent controllers**
+
+There are three CAPI controllers that are equally used by KaaS teams (Rocket, Hydra, Clippy) - core, kubeadm control plane and kubeadm bootstrap controller. They are all packaged together in the cluster-api-app. All KaaS teams have equal stakes in this app and equal responsibilities.
+
+Which teams is responsible for maintaining the application repository, improving it, automating it, taking care of renovate and other automated pull requests? Which teams takes care of upgrading the app to the latest upstream? Every team for themselves when they have a need? This is not working very efficiently.
+
+**Kubernetes components**
+
+We could say that deploying and using Kubernetes is sort of what we all do for living, right?
+
+Quick reminder that here we are talking about partial application lifecycle ownership, so release engineering and operations, where we can't really say that we do former one, so the latter one is interesting here.
+
+All teams write or package operators that get scheduled to some node, operators that use and maybe sometimes hammer API server, deploy few or few hundreds of CRDs that case clusters to be overloaded, etc.
+
+Let's say that all teams are more or less similarly using all (or most of) Kubernetes components. In that case a single team owning operations for those components would be ~~ludacris~~ ludicrous, and that didn't really work nicely when we tried it.
+
+#### 2.2.2. Implementing shared application ownership in practice
+
+So how do we implement shared application ownership in practice?
+
+Teams own apps with sole application ownership model. And multiple teams take turns, i.e. multiple teams participate in the rotation of sole application ownership.
+
+When a team is assigned to be sole app owner that means that the team takes care of app's release engineering and operations (on-call, postmortems, etc).
+
+How frequently the ownership is rotated depends on the app, and is decided on a case by case basis by the interested teams, but in most cases the app ownership should be passed to the next team every few months to half a year, depending on the release cadence of the app. Less then a month is probably too often, more than half a year could lead to folks falling out of practice (important for operations) or even mean that some team doesn't get to own the app as team structure changes.
+
+This shared application ownership model could be applied to the above examples in the following way (made up examples, not suggestions in the RFC):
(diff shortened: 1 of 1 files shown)


application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Based on the above, we can define two different criteria for defining application ownership models:
+- What stages of the application lifecycle are owned by a team?
+- Is there one or more Giant Swarm teams that can be considered owners of the application?
+
+## 1. Application lifecycle ownership
+
+We can define two high-level application ownership models:
+- Full application ownership
+- Partial application ownership, usually including release engineering (full or partial) and operations. This is typical for upstream applications where the development is done outside of Giant Swarm.
```
pipo02mix, line 27: But GS could collaborate on that part  (as we are doing now in CAPVCD)

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Partial application ownership of these components means that a Giant Swarm team owns:
+- Release engineering, fully or partially;
+- Operations, fully or partially, depending if the team has sole or shared ownership of the application, which is discussed in the next section.
+
+**Release engineering is owned fully** in case of forked projects, where Giant Swarm team builds the application from the forked repository, then tests, packages and finally delivers it.
+
+**Release engineering is owned partially** when Giant Swarm team does not build the app, but, for example, re-tagged container image is used. Here we have two cases:
+- Giant Swarm team is packaging app as a Helm chart in the app repository, by building required Helm templates for deploying re-tagged container image and optionally using existing upstream Kubernetes manifests instead of writing Helm templates from scratch.
+- Giant Swarm is repackaging or importing existing upstream Helm charts into Giant Swarm application repositories.
```
pipo02mix, line 55:
```suggestion
- Giant Swarm team is packaging the app as a Helm chart in the app repository, by building required Helm templates for deploying re-tagged container images and optionally using existing upstream Kubernetes manifests instead of writing Helm templates from scratch.
- Giant Swarm is repackaging or importing existing upstream Helm charts into Giant Swarm application repositories.
```

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The is the classical and usually the obvious case. A Giant Swarm team develops the application, in case of full application lifecycle ownership described in the previous section, or the team takes care of release engineering and operations for the app, in case of partial application lifecycle ownership described above.
+
+### 2.2. Shared application ownership
+
+And this is the tricky part, one of the main reasons for which this whole RFC is written.
+
+Shared application ownership applies in some cases of partial application lifecycle ownership, so in a nutshell, the question here is - which Giant Swarm team owns some upstream application? Can it even be one team, or multiple teams must share ownership? If multiple teams share ownership, how does that work in practice?
+
+There is no silver bullet for all upstream applications, so here we need clear and robust, but also flexible ownership model.
```
pipo02mix, line 73:
```suggestion
There is no silver bullet for all upstream applications, so here we need a clear and robust, but also flexible ownership model.
```

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### 2.2.1. Examples of shared applications
+
+While there are more examples, let's take the following two that should paint the picture and describe the issue of shared ownership clearly enough.
+
+**Cluster API provider-independent controllers**
+
+There are three CAPI controllers that are equally used by KaaS teams (Rocket, Hydra, Clippy) - core, kubeadm control plane and kubeadm bootstrap controller. They are all packaged together in the cluster-api-app. All KaaS teams have equal stakes in this app and equal responsibilities.
+
+Which teams is responsible for maintaining the application repository, improving it, automating it, taking care of renovate and other automated pull requests? Which teams takes care of upgrading the app to the latest upstream? Every team for themselves when they have a need? This is not working very efficiently.
```
pipo02mix, line 85: This was one of the main drivers why we thought about KaaS restructuring, having a single Lifecyle mgmt team for k8s will make ownership clearer. [We have just produced the first proposal](https://github.com/giantswarm/giantswarm/issues/23396#issuecomment-1413962194). WDYT?

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+**Kubernetes components**
+
+We could say that deploying and using Kubernetes is sort of what we all do for living, right?
+
+Quick reminder that here we are talking about partial application lifecycle ownership, so release engineering and operations, where we can't really say that we do former one, so the latter one is interesting here.
+
+All teams write or package operators that get scheduled to some node, operators that use and maybe sometimes hammer API server, deploy few or few hundreds of CRDs that case clusters to be overloaded, etc.
```
pipo02mix, line 93: IMO here we need better visibility. If we cannot identify which of the components is hammering down the API, we would not know which team is responsible

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### 2.2.2. Implementing shared application ownership in practice
+
+So how do we implement shared application ownership in practice?
+
+Teams own apps with sole application ownership model. And multiple teams take turns, i.e. multiple teams participate in the rotation of sole application ownership.
+
+When a team is assigned to be sole app owner that means that the team takes care of app's release engineering and operations (on-call, postmortems, etc).
+
+How frequently the ownership is rotated depends on the app, and is decided on a case by case basis by the interested teams, but in most cases the app ownership should be passed to the next team every few months to half a year, depending on the release cadence of the app. Less then a month is probably too often, more than half a year could lead to folks falling out of practice (important for operations) or even mean that some team doesn't get to own the app as team structure changes.
```
pipo02mix, line 105:
```suggestion
How frequently the ownership is rotated depends on the app, and is decided on a case by case basis by the interested teams, but in most cases the app ownership should be passed to the next team every few months to half a year, depending on the release cadence of the app. Less than a month is probably too often, more than half a year could lead to folks falling out of practice (important for operations) or even mean that some team doesn't get to own the app as team structure changes.
```

pipo02mix: Though with KaaS restructuration we will solve some problems of ownership we may have shared application that is not owned by a single team. You suggestion could work in that case but we will need to go case by case

application-ownership-models/README.md
nprokopic: Yeah, definitely, we can and we should work on those, but it's still an outside/upstream project which is under another organization. Thanks for the comment, I'll clarify it a bit more.

application-ownership-models/README.md
nprokopic: I was following the restructuring proposal doc, left a few comments.

The above here are just few examples, definitely not the only ones. I will add few more.

While this RFC could solve some issues that we are also trying to solve with team restructuring, this RFC and the restructuring serve a different purpose and are not mutually exclusive.

The ownership of many components will change and/or become more clear with team restructuring, but that just one of the effects of the restructuring, not its goal.

If you take a look at our "Ownership of component" spreadsheet, there we have almost 300 components. While assigning all those components to teams, we did our best, but still many of those components got assigned to teams that never worked on them, or to teams that were already overwhelmed and weren't exactly happy to get more responsibility. We have cases of components shared across teams (e.g. Giant Swarm modules used across many projects), our legacy components that we still have to maintain, patch at least to fix CVEs that are coming our way from upstream dependencies, etc. In many of those cases those applications/components are not super nicely maintained because they are not exactly a team priority.

So there are cases where either multiple teams really need something, or where almost none of the teams wants/needs it, and then different manifestations of those problems in different applications/components. All these diverse cases are what this RFC is targeting, by defining different models of application/component ownership (basically adding a rotation model for shared ownership), so we can more easily assign applications/components to teams, by having an option to assign it to multiple teams if that is needed.

puja108 (commented):
I kept my comment high level but added it already in the gdoc around KaaS restructure as that is my current focus: https://docs.google.com/document/d/1IV0ccmIN5U7U8WmYgvEdq2HinIcM-YdWCiu34bTjTk4/edit#

> From a sole app/tech perspective this might work (maybe that's enough for things like core CAPI controllers), but the reasoning behind ownership of something like Kubernetes and OS is that we really need to own them as customer facing "products", which means we need to constantly improve our operations of them (o11y, reliability, security,...) and monitor upstream for upcoming changes and deprecations which we need to work with our customers towards, incl. building tooling to test for deprecations, migrate to new APIs,...There's also a lot of pre-announcing and work with customers involved. Shared/rotated ownership most probably won't cover those, as they won't get any priority from a team that has a different mission and priorities.
> We've hit issues around this before: e.g. being totally surprised by cgroupsv2 landing in Flatcar or only recently realizing the old secrets usage of Service Account is deprecated with K8s 1.25. Our customers are even more surprised ;)
> As long as no team considers these apps/components and more importantly their features and lifecycle as a product and their sole priority towards customers (btw, both internal and external) I feel we won't improve the status quo.

application-ownership-models/README.md
nprokopic: Agreed definitely, we need better visibility there. This is just an example, I believe Ludacris had, and whichever team gets to own "core" k8s components will have similar challenges, which is why I added this as an example, just for the sake of painting a wider picture in the RFC, to give different examples where it could help (not to have those specific examples as proposals in the RFC).

application-ownership-models/README.md
Gacko: I think this should not be part of the RFC. Upstream contributions still needs to be approved by them, so we can only offer changes. I would not call this real development and omit from the RFC for keeping things simple.

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Operations are handled by the team members, including business hours as a part of team on-call, and non-business hours as a part of team/area on-call. There are few caveats here:
+- The non-business hours on-call is currently shared with other teams from the same area (hence "almost" full ownership), which is done for the sake of not overwhelming on-callers with frequent night shifts.
+- Ideally, if a team has enough people and we have state-of-the-art monitoring and alerting, meaning that there are next-to-none false positive alerts (for example caused by alerts from other team's components), the team can own 24h on-call for the application.
+
+### 1.2. Partial application lifecycle ownership
+
+Partial application ownership means that application development is done outside of Giant Swarm, therefore this is the case of upstream applications that we are deploying to our clusters.
+
+To name a few examples, we run, but do not develop, Prometheus, Grafana, Cluster API controllers, Kyverno, different CNI, CSI, CPI apps, then core Kubernetes components, such as API server, scheduler, controller manager, etcd, kubelet, etc.
```
Gacko, line 45:
```suggestion
To name a few examples, we run, but do not develop, Prometheus, Grafana, Cluster API controllers, Kyverno, different CNI, CSI, CPI apps, as well as core Kubernetes components, such as API server, scheduler, controller manager, etcd, kubelet, etc.
```

application-ownership-models/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## 2. Sole or shared application ownership
+
+While different application ownership models based on application lifecycle stages are relatively clear and it's obvious which model is used for every application, for some apps is much less evident which team should own them, and if there can even be just one owning team, or multiple teams cannot be avoided due to shared nature of the app.
+
+Additionally, while application lifecycle stages split the app ownership between external (upstream) and internal (Giant Swarm) teams, sole or shared ownership is about how we handle app ownership internally within Giant Swarm.
+
+### 2.1. Sole application ownership
+
+The is the classical and usually the obvious case. A Giant Swarm team develops the application, in case of full application lifecycle ownership described in the previous section, or the team takes care of release engineering and operations for the app, in case of partial application lifecycle ownership described above.
```
Gacko, line 65:
```suggestion
This is the classical and usually the obvious case. A Giant Swarm team develops the application, in case of full application lifecycle ownership described in the previous section, or the team takes care of release engineering and operations for the app, in case of partial application lifecycle ownership described above.
```

Gacko (approved):

nprokopic:
I was busy working on CAPZ lately, but should have some time next week to update the RFC in order to address the comments and clarify it further (e.g. to make it more clear that it's not targeting same goals as restructuring).

Then I'll re-ping folks, maybe we chat about it in some SIG, so we can hopefully agree here and merge it.

teemow:
TJust to make clear as I don't understand the importance of this RFC. Within product we have clarified why shared ownership doesn't work and we've decided to remove all shared ownerships. This is why we went over the components last summer and removed all shared ownerships. With the restructuring we also take care that we don't have shared ownership again. And it seems this whole RFC was written because of shared ownership. @nprokopic did you know that this topic was already addresses? Do we need this RFC? What does it solve?

See: 
 * https://docs.google.com/spreadsheets/d/1GNXAbYo_HyjjRusd2KTGttuORBkX9iMdgnNp-xTQ9sw/edit#gid=1577959677
 * https://github.com/giantswarm/giantswarm/issues/22964

nprokopic:
With the restructuring this got de-prioritized and now outdated.

> With the restructuring we also take care that we don't have shared ownership again. And it seems this whole RFC was written because of shared ownership. @nprokopic did you know that this topic was already addresses?

@teemow I was aware that the restructuring, in addition to other goals, tries to address the problems of shared ownership. I wasn't aware that the goal was not to have it at all again. I do get that we do not want to have shared ownership at a higher (product) level (some of the examples in the RFC are not good here, especially now after restructuring has been almost done).

> Do we need this RFC? What does it solve?

With the approach that we want to remove shared ownership at all levels we end up assigning components to specific teams which are then not satisfied with those components being assigned to them, as they lack expertise, they do not use them, or they use them and have equal stakes  as some other teams. This again can cause those components to be poorly and inefficiently maintained and not improved, and we're back to square one, because the main problem is not solved.

While we might even succeed at assigning to specific teams all apps that we deploy and support, and those teams being mostly fine with it, we still have a lot of lower level components that are not obvious when looking at the ownership issues from a higher level (e.g. shared go modules used across many projects).

That being said, this RFC did get outdated now with restructuring (plus some examples are not good, and seem to tackle same issues as restructuring did, which was not the goal here; also it should be "Component" instead of "Application" in the title).

I still think that we need something like this, in order to address the issues I mentioned above, but the text definitely needs to be reevaluated, reworded and updated after the restructuring dust settles. Therefore I will close this RFC for now and revisit the topic in 3-4 months, when I'll first try to assess if the issues I mentioned exist in practice.

PR #64, opened 2023-03-16
Simplify basedomain usage

fiunchinho:
The general idea is to stop using the `baseDomain` value that we have in the `config` repo and in the `cluster-values` configmap, in some of our apps, and slowly move away from this global value so that is less risky to work with it, and we make our platform more flexible.

### DIFF
--- a/simplify-basedomain-usage/README.md
+++ b/simplify-basedomain-usage/README.md
@@ -1,0 +1,175 @@
+# Simplify `baseDomain` usage in our applications
+
+The `baseDomain` is the common suffix that we use as a base for the DNS records created for our clusters.
+The meaning of the `baseDomain` is different in `vintage` and CAPI installations. In `vintage`, it contains information about the management cluster (it identifies a management cluster) while in CAPI installations, it doesn't contain any information about the management cluster.
+
+For example in CAPI installation, if the `baseDomain` is `gaws.gigantic.io`, a cluster called `mycluster` will have its kubernetes API endpoint listening at `api.mycluster.gaws.gigantic.io`. It doesn't matter if the cluster is a management cluster or a workload cluster.
+
+In `vintage`, if the `baseDomain` is `goku.germanywestcentral.azure.gigantic.io` (notice that it contains the name of the management cluster), the management cluster kubernetes API endpoint will be listening at `g8s.goku.germanywestcentral.azure.gigantic.io`.
+While a workload cluster called `mycluster` will have its kubernetes API endpoint listening at `api.x6rtd.k8s.goku.germanywestcentral.azure.gigantic.io`.
+
+## Problems
+
+### Global value
+
+We have made the `baseDomain` value available everywhere, and we have lost track of which apps depend on it or how they are using it.
+It's global value that is passed around magically. Changing the `baseDomain` has unknown and potentially catastrophic consequences.
+
+### Complex setup
+
+We configure the `baseDomain` differently depending on how the app that needs the value is deployed:
+
+- For apps that are part of [the app collections deployed to our Management Clusters](https://github.com/giantswarm/capa-app-collection), we pass [the `baseDomain` using the `config` repo](https://github.com/giantswarm/config/blob/main/default/apps/nginx-ingress-controller-app/configmap-values.yaml.template#L2).
+- For [the default apps](https://github.com/giantswarm/default-apps-aws) (apps deployed to both Management Clusters and Workload Clusters), we store the `baseDomain` in a special configmap called `$clustername-cluster-values` created by `cluster-apps-operator` in CAPI installations and by `cluster-operator` in `vintage`. This configmap [is referenced from the `App` CR of the default apps](https://github.com/giantswarm/default-apps-aws/blob/master/helm/default-apps-aws/templates/apps.yaml#L38-L40).
+- For other apps, we also store the `baseDomain` [in several app catalogs configuration](https://github.com/giantswarm/installations/blob/master/gohan/appcatalog/default-appcatalog-values.yaml#L5). That way the `baseDomain` is passed to all applications installed from those catalogs (Honeybadger has already stated that app catalog configuration may go away in the future).
+
+### Differences between `vintage` and CAPI installations
+
+The difference in the value for `baseDomain` between `vintage` and CAPI installations is sometimes confusing while working on applications that will be deployed to both kinds of installations.
+The fact that we reuse the same variable name for values that could potentially contain different things (does it contain info about the management cluster or not?) is confusing and error-prone, adding complexity where this value is used.
+
+### Different clusters in the same installation can't use different `baseDomain` values
+
+We have several controllers reconciling workload clusters, and these controllers receive the `baseDomain` configuration value to create resources for the reconciled clusters.
+Because we pass the `baseDomain` as a parameter to the controllers, all clusters in a single installation must share the same `baseDomain`. There is no way to use different a `baseDomain` for two workload clusters on the same installation.
+
+### Same variable name for different values
+
+In our `installations` repository we save the `baseDomain` in the app catalog, so that this value is automatically passed to all apps from that catalog. For example, [on `goat` we have this value](https://github.com/giantswarm/installations/blob/master/gohan/appcatalog/default-appcatalog-values.yaml#L5)
+```yaml
+baseDomain: gtest.gigantic.io
+```
+
+But most applications expect the `baseDomain` to contain the cluster name as a prefix, like `mycluster.gtest.gigantic.io`.
+
+These differences here and there are confusing.
+
+## Proposal #1: Add `baseDomain` as an annotation to Cluster Custom Resource in CAPI installations
+
+There are some operators that we deploy in management clusters which reconcile workload clusters, and use the `baseDomain` to perform their business logic. As explained above, these operators receive the `baseDomain` through the `config` repository.
+Instead, we could add a new annotation `"giantswarm.io/dns-zone"` to the `Cluster` Custom Resources. That way, when reconciling clusters, our operators could read the annotation from the `Cluster` object instead of requiring to configure the `baseDomain` in the `config` repository.
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  annotations:
+    cluster.giantswarm.io/description: golem MC
+    giantswarm.io/dns-zone: golem.gaws.gigantic.io
+  labels:
+    app: cluster-aws
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/version: 0.26.0
+    application.giantswarm.io/team: hydra
+    cluster-apps-operator.giantswarm.io/watching: ""
+    cluster.x-k8s.io/cluster-name: golem
+    giantswarm.io/cluster: golem
+    giantswarm.io/organization: giantswarm
+    helm.sh/chart: cluster-aws-0.26.0
+  name: golem
+  namespace: org-giantswarm
+```
+
+### Advantages
+
+On top of reducing the usage of the global `baseDomain` value, there are other advantages when using this approach.
+
+#### Custom Resources triggering changes rather than controllers
+
+Moving configuration from operators to custom resources is always preferable. Starting by the obvious, the cluster configuration is right there explicitly in the `Cluster` Custom resource rather than hidden in controllers configuration. 
+Furthermore, if the value is part of the `Cluster` CR, it's the `Cluster` the one driving and triggering changes on cloud resources, allowing clusters to change or update their values independently of each other.
+This is a better scenario than our current approach having the controller receive this configuration as a parameter, because changing the parameter would trigger changes in cloud resources on all clusters.
+
+#### Reduce coupling
+
+Controllers fetching this value from the `cluster-values` configmap depend on the `cluster-apps-operator` or `cluster-operator` deployed and creating the configmap.
+Using the value from the annotation instead of depending on the configmap reduces coupling between our controllers.
+
+#### Potential to use different DNS hosted zones on the same Management Cluster
+
+With our current approach, we configure the `baseDomain` at the installation level through the `config` repository, making the `baseDomain` shared for all clusters in a single installation.
+If we move the value to be part of the `Cluster` CR, different clusters could potentially use different hosted zones, even on the same Management Cluster.
+
+#### Migrate workload clusters to different management cluster
+
+One side effect from this is that by decoupling the `baseDomain` from the Management Cluster would also help moving clusters between installations, if that would be desired.
+
+
+### Downsides
+
+#### Migration
+
+For the migration to this approach, we would need to annotate all `Cluster` Custom Resources to have the annotation.
+We also would need to change our operators so that they read this value instead of the parameter they currently receive using the `config` repository.
+
+##### List of controllers that reconcile clusters and use the `baseDomain`
+
+###### aws-resolver-rules-operator
+
+It needs it to construct the k8s api endpoint of the cluster that's reconciling, so that it can create a Route53 Resolver Rule for it.
+
+###### cluster-apps-operator
+
+It needs it to store it in the `cluster-values` configmap that will be consumed by default apps of the cluster that's reconciling.
+
+###### dex-operator
+
+It needs it to configure dex for the cluster that's reconciling. The `baseDomain` passed to the app in the `config` repository is only used in `vintage` management clusters.
+For CAPI clusters or `vintage` workload clusters, the controller loads the `cluster-values` configmap from the management cluster.
+It should also take the value from the annotation instead.
+
+###### dns-operator-*
+
+It needs it to create the hosted zone and DNS records for the cluster that's reconciling.
+
+###### irsa-operator
+
+For CAPI clusters it loads the `cluster-values` configmap from the management cluster. It should also take the value from the annotation instead.
+
+## Proposal #2: Stop using the `baseDomain` from the `cluster-values` configmap in CAPI installations
+
+The `cluster-apps-operator` in CAPI installations, (and the `cluster-operator` in `vintage` installations) is creating a configmap called `$clustername-cluster-values` that contains values related to the workload cluster called `$clustername`.
+One of these values is the `baseDomain` for that cluster.
+This configmap is then added as [a config source in the `App` CRs created for the workload clusters default apps](https://github.com/giantswarm/default-apps-aws/blob/master/helm/default-apps-aws/templates/apps.yaml#L38-L40).
+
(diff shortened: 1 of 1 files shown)


simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,175 @@
...
+### Different clusters in the same installation can't use different `baseDomain` values
+
+We have several controllers reconciling workload clusters, and these controllers receive the `baseDomain` configuration value to create resources for the reconciled clusters.
+Because we pass the `baseDomain` as a parameter to the controllers, all clusters in a single installation must share the same `baseDomain`. There is no way to use different a `baseDomain` for two workload clusters on the same installation.
+
+### Same variable name for different values
+
+In our `installations` repository we save the `baseDomain` in the app catalog, so that this value is automatically passed to all apps from that catalog. For example, [on `goat` we have this value](https://github.com/giantswarm/installations/blob/master/gohan/appcatalog/default-appcatalog-values.yaml#L5)
+```yaml
+baseDomain: gtest.gigantic.io
```
AndiDog, line 40: Somewhat related, maybe: the `managementCluster` key is also specified there – doesn't that one present similar issues as `baseDomain`? For example, aws-network-topology-operator takes the MC name as argument.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,175 @@
...
+#### Aligned with our GitOps strategy
+
+All the configuration would be in the repository containing the default apps manifests, instead of this configmap being dynamically created by an operator.
+
+#### No need to migrate cluster-apps-operator
+
+If `cluster-aps-operator` wouldn't need the `baseDomain` anymore, we wouldn't need to migrate the operator to take the value from an annotation.
+
+### Downsides
+
```
AndiDog, line 162: Another downside: same as we have now, the value is spread at multiple places, so we cannot change it in one central place, and may have multiple different values by mistake.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,175 @@
...
+#### Verbosity in default apps bundle
+
+We would need to explicitly pass the `baseConfig` to all the different apps that need it in the default apps bundle.
+
+##### List of default apps using the `baseDomain` from the `cluster-values` configmap
+
+###### cilium
+
+It uses the `baseDomain` to talk to the k8s api. It already receives the `baseDomain` from the `default-apps-$provider` bundle, so there is no need to do anything here.
```
AndiDog, line 171: Overall, can we even bring down the usage of `baseDomain` to **either** proposal 1 or 2, removing all other usages? I feel we must keep living with a mixture, or proposal 2 alone.

I don't see how proposal 1 (add a `Cluster` annotation) would allow our app Helm templates to read that value – unless we use Helm's [`lookup` function](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) which IMO raises the level of dependencies and mental complexity.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Differences between `vintage` and CAPI installations
+
+The difference in the value for `baseDomain` between `vintage` and CAPI installations is sometimes confusing while working on applications that will be deployed to both kinds of installations.
+The fact that we reuse the same variable name for values that could potentially contain different things (does it contain info about the management cluster or not?) is confusing and error-prone, adding complexity where this value is used.
+
+### Different clusters in the same installation can't use different `baseDomain` values
+
+We have several controllers reconciling workload clusters, and these controllers receive the `baseDomain` configuration value to create resources for the reconciled clusters.
+Because we pass the `baseDomain` as a parameter to the controllers, all clusters in a single installation must share the same `baseDomain`. There is no way to use different a `baseDomain` for two workload clusters on the same installation.
```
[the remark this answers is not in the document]
fiunchinho: I'd say the issues with this are
- Different workload clusters managed by the same management cluster can't use different domains. If the customer wants a workload cluster with a different `baseDomain` , a new management cluster needs to be created
- Workload clusters can't be moved to a different management cluster, the `baseDomain` creates a dependency between the management cluster and the workload cluster

I'd say "no ability to rename the domain for new WCs" is included in the first issue above. Can you think of any more issues worth mentioning? Do you want me to add these two to the RFC?

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,175 @@
+# Simplify `baseDomain` usage in our applications
+
+The `baseDomain` is the common suffix that we use as a base for the DNS records created for our clusters.
+The meaning of the `baseDomain` is different in `vintage` and CAPI installations. In `vintage`, it contains information about the management cluster (it identifies a management cluster) while in CAPI installations, it doesn't contain any information about the management cluster.
+
+For example in CAPI installation, if the `baseDomain` is `gaws.gigantic.io`, a cluster called `mycluster` will have its kubernetes API endpoint listening at `api.mycluster.gaws.gigantic.io`. It doesn't matter if the cluster is a management cluster or a workload cluster.
+
+In `vintage`, if the `baseDomain` is `goku.germanywestcentral.azure.gigantic.io` (notice that it contains the name of the management cluster), the management cluster kubernetes API endpoint will be listening at `g8s.goku.germanywestcentral.azure.gigantic.io`.
+While a workload cluster called `mycluster` will have its kubernetes API endpoint listening at `api.x6rtd.k8s.goku.germanywestcentral.azure.gigantic.io`.
+
```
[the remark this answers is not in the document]
fiunchinho: Can I resolve this one or do you feel there is something else I could add?

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+###### dex-operator
+
+It needs it to configure dex for the cluster that's reconciling. The `baseDomain` passed to the app in the `config` repository is only used in `vintage` management clusters.
+For CAPI clusters or `vintage` workload clusters, the controller loads the `cluster-values` configmap from the management cluster.
+It should also take the value from the annotation instead.
+
+###### dns-operator-*
+
+It needs it to create the hosted zone and DNS records for the cluster that's reconciling.
```
bavarianbidi, line 134: There are some stumbling blocks we have to be aware of:

E.g. `mycluser.azuretest.gigantic.io` is set as annotation in the Cluster CR, the `dns-operator` need explicit permissions to create the `NS` records for `mycluster` in the DNS-Zone `azuretest`. For `dns-operator-azure` a dedicated app-credential is needed as the `azuretest` DNS-Zone could exist in another subscription.

If we should support `mycluster.customerproject1.example.com` and `myothercluster.customerproject666.example.com` on azure, we need to know where `customerproject1` and `customerproject666` are hosted and we need explicit permissions to create NS records there (if we are on Azure). For sure, with more far-reaching permissions for the NS record creation or a dedicated list of "supported and granted" DNS zones we can achieve this, but we have to be aware of this and we have to implement this.

Not sure how the `dns-operator` for `aws` or `gcp` is handle such things but i assume, the permission topic might hit every controller somehow.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+###### irsa-operator
+
+For CAPI clusters it loads the `cluster-values` configmap from the management cluster. It should also take the value from the annotation instead.
+
+## Proposal #2: Set value in default-apps-* config instead of using `baseDomain` from the `cluster-values` configmap in CAPI installations
```
bavarianbidi, line 140: at the latest with the implementation of the `proxy` feature, the featureset of `cluster-apps-operator` is highly controversial. 

There is
* a [discussion in KaaS-sync](https://docs.google.com/document/d/15fBIUmv5pPvCF1DLwR6_XOaFLNHeVex8YNrVtgm83ME/edit#bookmark=id.j92bbrwm7zno)
* and [slack thread](https://gigantic.slack.com/archives/C01F7T2MNRL/p1669013425769809)
* which lead to a [github issue](https://github.com/giantswarm/giantswarm/issues/24784) to track that

when it came to a decision, we should consider/re-prioritize the `cluster-apps-operator` issue again.

> sorry for adding another complexity level here :hugs:

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+With our current approach, we configure the `baseDomain` at the installation level through the `config` repository, making the `baseDomain` shared for all clusters in a single installation.
+If we move the value to be part of the `Cluster` CR, different clusters could potentially use different hosted zones, even on the same Management Cluster.
+
+#### Migrate workload clusters to different management cluster
+
+One side effect from this is that by decoupling the `baseDomain` from the Management Cluster would also help moving clusters between installations, if that would be desired.
+
+
+### Downsides
```
bavarianbidi, line 109: As much as i like this idea, it's not clear to me how we could/should do the defaulting here.
What value should be used, if no `baseDomain` is given?

bavarianbidi (commented):
Thx Jose for that RFC, i really hope we find a better way to handle this `baseDomain` thing as it's super confusing and already took me hours to understand, why app X has this value set and why app Y has another value for `baseDomain` and where the hell does the value came from :disappointed: 

Maybe as a third (temporary) option, we could also think about renaming `baseDomain` in `config` repo towards something like `installationBaseDomain` as for most (:crossed_fingers: all) components managed via `config` repo, the value defined here is an installation specific one.

And later on, this value got propagated/concatenated together with the cluster-name as the `baseDomain` for all the default-apps.

simplify-basedomain-usage/README.md
fiunchinho: I was thinking it be would a required value that has no default. Would that be a problem?

simplify-basedomain-usage/README.md
fiunchinho: [In this comment](https://github.com/giantswarm/giantswarm/issues/24784#issuecomment-1477348732) you wrote

> with the recent RFC regarding simplifying basedomain usage cluster-apps-operator got named as potential solution by extending the feature set again.

One of us is miss understanding something hehe I'm proposing to stop using the `baseDomain` value added by `cluster-apps-operator`. So it's not extending the feature set, but actually reducing it. I want to get rid of that value from the `cluster-values` configmap. Let me know if I miss understood your comment, or how I can improve the wording on the RFC, as it seems I didn't do a good job explaining it :smile:

bavarianbidi: :see_no_evil: you're absolutely right ... i was to much focused on `cluster-apps-operator` that i totally misread your proposed solution.

kopiczko (commented):
I'm happy with `Proposal #2`, not crazy about `Proposal #1`

puja108 (commented):
I don't have many opinions here, but we are touching the way we are configuring deployment here and in general, and we need to also think about how this would work in a future that might rely more on an upstream interface like HelmReleases. I think @giantswarm/team-honeybadger 's opinion will be very valuable here.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Migration
+
+For the migration to this approach, we would need to annotate all `Cluster` Custom Resources to have the annotation.
+We also would need to change our operators so that they read this value instead of the parameter they currently receive using the `config` repository.
+
+##### List of controllers that reconcile clusters and use the `baseDomain`
```
anvddriesch, line 116: Just a side note, if we want our controllers on the MC to get the baseDomain value from the Cluster CR, we will introduce a dependency on cluster api to all of them. (e.g. dex-operator does not currently have this dependency)

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+In our `installations` repository we save the `baseDomain` in the app catalog, so that this value is automatically passed to all apps from that catalog. For example, [on `goat` we have this value](https://github.com/giantswarm/installations/blob/master/gohan/appcatalog/default-appcatalog-values.yaml#L5)
+```yaml
+baseDomain: gtest.gigantic.io
+```
+
+But most applications expect the `baseDomain` to contain the cluster name as a prefix, like `mycluster.gtest.gigantic.io`.
+
+These differences here and there are confusing.
+
+## Proposal #1: Add `baseDomain` as an annotation to Cluster Custom Resource in CAPI installations
```
anvddriesch, line 52: This proposal seems promising.
However, if we want to reduce the differences between CAPI and vintage, adding this in vintage installations as well would improve the situation.
The only clusters that would still need to rely on the old ways would be the vintage MC (as they don't have a CR) and kvm WC (as they use the kvmconfig CR)

simplify-basedomain-usage/README.md
anvddriesch, line 140: Some other apps which are not default apps also use this configmap to retrieve the `baseDomain` value.
It also would not help a lot in reducing the gap between vintage and capi since it only addresses CAPI default apps.
Therefore we in rainbow favor the first proposal.

fiunchinho: > Some other apps which are not default apps also use this configmap to retrieve the `baseDomain` value. It also would not help a lot in reducing the gap between vintage and capi since it only addresses CAPI default apps. 

Do you know which apps use the configmap?

> Therefore we in rainbow favor the first proposal.

Just for the record, the proposals don't cancel each other, my idea was to implement both.

simplify-basedomain-usage/README.md
fiunchinho: True. I think only dex-operator doesn't depend on CAPI already, so yeah, it would require one new dependency for that controller. I personally think it's not a blocker, but I'll add this as a disadvantage.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+    cluster.x-k8s.io/cluster-name: golem
+    giantswarm.io/cluster: golem
+    giantswarm.io/organization: giantswarm
+    helm.sh/chart: cluster-aws-0.26.0
+  name: golem
+  namespace: org-giantswarm
+```
+
+The `baseDomain` would still be used by the applications that we deploy to management clusters that don't reconcile clusters, because these applications need the DNS zone of the management cluster.
+This variable could potentially be renamed to something like `managementClusterDNSZone` in the future to make clear its purpose, but it's not in the scope of this RFC to discuss this.
```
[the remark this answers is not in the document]
fiunchinho: Currently the `baseDomain` is in three different places. With these changes the `dns-zone` is only in one place, the annotation.

If folks think this change will make the migration harder for some reason we could make this change in vintage as well.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+The difference in the value for `baseDomain` between `vintage` and CAPI installations is sometimes confusing while working on applications that will be deployed to both kinds of installations.
+The fact that we reuse the same variable name for values that could potentially contain different things (does it contain info about the management cluster or not?) is confusing and error-prone, adding complexity where this value is used.
+
+### Different clusters in the same installation can't use different `baseDomain` values
+
+We have several controllers reconciling workload clusters, and these controllers receive the `baseDomain` configuration value to create resources for the reconciled clusters.
+Because we pass the `baseDomain` as a parameter to the controllers, all clusters in a single installation must share the same `baseDomain`. This has some big drawbacks
+
+- Different workload clusters managed by the same management cluster can't use different domains. If the customer wants a workload cluster with a different `baseDomain`, a new management cluster needs to be created.
```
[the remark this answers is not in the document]
fiunchinho: Ok, I'll rephrase it to 

> - Different workload clusters managed by the same management cluster can't use different domains easily. If the customer wants a workload cluster with a different `baseDomain`, they would have to create a CNAME for the new domain and pass special configuration for that specific cluster (they could also create a new management cluster).

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,186 @@
...
+    giantswarm.io/organization: giantswarm
+    helm.sh/chart: cluster-aws-0.26.0
+  name: golem
+  namespace: org-giantswarm
+```
+
+The `baseDomain` would still be used by the applications that we deploy to management clusters that don't reconcile clusters, because these applications need the DNS zone of the management cluster.
+This variable could potentially be renamed to something like `managementClusterDNSZone` in the future to make clear its purpose, but it's not in the scope of this RFC to discuss this.
+
+### Advantages
```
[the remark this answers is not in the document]
fiunchinho: Cool, I'll add it as an advantage.

simplify-basedomain-usage/README.md
anvddriesch: I suppose that's true for most provider-specific stuff.
From a rainbow perspective, as we deal with access, all our operators/apps act the same across all providers, including capi and kvm. Furthermore, dex-operator in particular treats management clusters and workload clusters the same as it just cheks for dex App CRs on the MC. So for workload clusters, it needs to rely on the `cluster-values` configmap to get the right domain.
If we decide to migrate, let's make sure to align and not break SSO access.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### No need to migrate cluster-apps-operator
+
+If `cluster-aps-operator` wouldn't need the `baseDomain` anymore, we wouldn't need to migrate the operator to take the value from an annotation.
+
+### Downsides
+
+#### Verbosity in default apps bundle
+
+We would need to explicitly pass the `baseConfig` to all the different apps that need it in the default apps bundle.
```
nprokopic, line 182: I don't really see it as a downside, I see it like less magic and more explicit, visible and clear passing of configuration to the apps that need it.

simplify-basedomain-usage/README.md
```diff
@@ -0,0 +1,202 @@
...
+In our `installations` repository we save the `baseDomain` in the app catalog, so that this value is automatically passed to all apps from that catalog. For example, [on `goat` we have this value](https://github.com/giantswarm/installations/blob/master/gohan/appcatalog/default-appcatalog-values.yaml#L5)
+```yaml
+baseDomain: gtest.gigantic.io
+```
+
+But most applications expect the `baseDomain` to contain the cluster name as a prefix, like `mycluster.gtest.gigantic.io`.
+
+These differences here and there are confusing.
+
+## Proposal #1: Add `baseDomain` as an annotation to Cluster Custom Resource in CAPI installations
```
nprokopic, line 52: Since the annotation would be set in `cluster-*` app (as that's where the Cluster CR is coming from), does this effectively mean that the source of truth for the base domain value is `cluster-*` app config?

fiunchinho: If here we mean `baseDomain` as the `dns-zone` containing the cluster name like `jose.gaws.gigantic.io`, then I'd say so, yeah.

anvddriesch (approved):

anvddriesch (approved):

nprokopic (approved):

PR #67, opened 2023-04-24
Better control over changes in production environments

fiunchinho:
I've added a small RFC. It contains an alternative way of deploying changes to our production installations.

### DIFF
--- a/better-control-over-production-changes/README.md
+++ b/better-control-over-production-changes/README.md
@@ -1,0 +1,20 @@
+# Better control over changes in production environments
+
+We currently use `flux` to deploy changes to our Management Clusters. `flux` is reconciling the `main` branch on different repositories.
+We have some customers that don't want all changes to be deployed continuously to our production installations.
+Instead, they want us to deploy to a testing installation first, and then deploy to the production installation during a maintenance window.
+
+## Current solution
+
+In order to do that, we keep `flux` suspended on this production installation while it's running on the testing installations. Then, we we want to upgrade the production installation, we manually resume `flux`, wait for the changes to be applied and suspend `flux` again.
+
+## Proposed alternative
+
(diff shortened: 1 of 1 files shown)


fiunchinho:
> Is this a proposal for [giantswarm/giantswarm#26519](https://github.com/giantswarm/giantswarm/issues/26519)?

Nope, as mentioned during stand up, I didn't know about https://github.com/giantswarm/giantswarm/issues/26519. So I'm going to leave my feedback there and close this RFC.

better-control-over-production-changes/README.md
```diff
@@ -0,0 +1,20 @@
...
+We have some customers that don't want all changes to be deployed continuously to our production installations.
+Instead, they want us to deploy to a testing installation first, and then deploy to the production installation during a maintenance window.
+
+## Current solution
+
+In order to do that, we keep `flux` suspended on this production installation while it's running on the testing installations. Then, we we want to upgrade the production installation, we manually resume `flux`, wait for the changes to be applied and suspend `flux` again.
+
+## Proposed alternative
+
+We create a new branch called `production` in the `capa-app-collection` and `config` repositories. And we make `flux` running in our production installation to target and reconcile this `production` branch.
```
puja108, line 13: I guess this way in the future we could also manage more than 1 channel? Like maybe a user-facing RC/staging channel?

puja108: Also cc @piontec as it relates to Honeybadger in two ways:
- Honeybadger is planning to work on MCF improvements that would enable us more flexibility with installations
- I'm not sure app collections are something we want to keep in general /cc @alex-dabija

fiunchinho: Sorry @puja108 , I meant to close this RFC but I forgot. I opened the RFC without knowing [this existed](https://github.com/giantswarm/giantswarm/issues/26519), so I'm going to add my feedback there instead. I believe the way forward is moving apps into MCF instead of using the collections.

PR #68, opened 2023-05-19
Release channels and staged deployments

nprokopic:
WIP, description TBA, early comments are welcome.

### DIFF
--- a/release-channels-and-staged-deployments/README.md
+++ b/release-channels-and-staged-deployments/README.md
@@ -1,0 +1,379 @@
+# Release channels and staged deployments
+
+The document talks about releases, deployments, how we deploy components in different ways, when we deploy them.
+
+It starts with an overview of what we have today, and then continues with how we can structure releases an deployments in the future.
+
+## 1. What and when we release and deploy today?
+
+### 1.1. What are we deploying?
+
+1. cluster-x app
+  - Cluster resources
+  - Most critical apps: CPI, CNI, CSI
+  - Owners: LM, CI, OPI
+
+2. default-apps-x
+  - Workload cluster apps
+  - Owners: LM, CI, OPI
+
+3. x-app-collection 
+  - Management cluster apps
+  - Owners: LM, CI, OPI
+
+This means we can say that we have a have:
+- Workload cluster 
+- Management cluster version
+
+Config is not included above. Applying, changing some config is usage of released APIs, it is not an act of releasing nor deploying. An app that is released exposes available config - setting and changing that config does not affect the release.
+
+Components:
+- Kubernetes
+- Flatcar OS
+- Upstream apps we package
+- In-house apps we build
+- Configuration of all the above (does not really goes into releases story)
+
+### 1.2. When and how often are we deploying?
+
+Today we and our customers have a need and use cases for deploying components in two different ways (when it comes to when and how often):
+1. Automated continuous deployments
+    - We use this approach mostly for MCs.
+    - No need to go through a bigger release process.
+    - Faster feedback loop when working on MC components, which can be especially important in KaaS Area.
+    - Less control over upgrades, bugs can get deployed to many (all) clusters before catching them.
+2. Manually triggered deployments
+    - We use this approach mostly for WCs.
+    - We use this approach for updating vintage WCs to newer releases.
+    - This way we and the customers have more control over the cluster upgrades.
+    - The control over upgrades can be especially important for some customers' production clusters where we also have freeze periods where we must not do updates.
+    - It can be a problem when security and critical bug fixes have to be rolled out as soon as possible.
+
+#### 1.2.1. Automated continuous deployments
+
+Giant Swarm today in vintage:
+- We continuously push some apps to app collections and those are being then deployed automatically to all MCs. (MC)
+
+Giant Swarm today in CAPI:
+- We continuously push some apps to app collections and those are being then deployed automatically to all MCs. (MC)
+- We continuously deploy some of cluster-x apps to some management clusters with renovate automatically merging PRs. (WC, MC)
+- We continuously deploy some of default-apps-x apps to some management clusters with renovate automatically merging PRs. (WC, MC)
+
+#### 1.2.2. Manual deployments
+
+Today in vintage:
+- We and our customers manually update vintage WCs to newer releases.
+
+Giant Swarm today in CAPI:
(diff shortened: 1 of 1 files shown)


fiunchinho:
Thanks for putting this together. I think this is a great overview on how things could look like in the future. I believe it could be more useful if would be a little more specific. Right now it describes a generic set of components, but we do package things in a specific way, and we use specific tooling.
For example, we package both the kubernetes version and the OS version as part of our `cluster-$provider` apps. Keeping that in mind, I think the RFC would be more actionable if we would describe how we would release and promote a new version of a `cluster-$provider` app.

In a similar way, I'd be more specific and describe what channels would look like. Each MC uses a different version of apps depending on which channel they are on? Do we need different repos for different channels? etc.

nprokopic:
Thanks @fiunchinho for the feedback here, many comments are spot on, I was also thinking about some of those.

I will continue iterating on the RFC this week, and will try to address your comments.

In general, I did not want to include specific technologies into the RFC, and left than intentionally out of scope here, but I see that it would be more clear if it went more into the specifics.

I would include specific examples that answer those questions, e.g. examples that illustrate how the RFC implementation would look like in `cluster-$provider` apps, how it would look like in app collections, how it would look like for Kubernetes and OS versions, etc. These would be just examples that try to illustrate the idea, and they would not be concrete proposals that would be the part of the RFC.

My idea is to write and clarify as a part of the RFC/proposal **what** should be implemented and where at the point, but **not how** that should be implemented, since I would like to **define the "contract"/"API"** here for release channels and staged deployments, and not specific implementations. Does that sound good?

nprokopic:
FYI I will be on the vacation for the next 2 weeks, so I'll continue with this as soon as I come back, as this is getting back up in priority list, as other work (CAPZ-related) got done.

nprokopic:
Quick thoughts before vacation (will try to structure these into more meaningful and more structured test in the RFC after doing more research so I don't write total nonsense :smile:):

>  Each MC uses a different version of apps depending on which channel they are on?

Yes. To put it a bit differently (sounds better in my head), all provider X MCs from some release channel would have same versions of all apps. MCs that are on different release channels can have versions of some apps different (for those apps where the latest version of the app still didn't reach production release channels).

In practice, MCs would use `Testing`, `Staging` and `Production Edge` release channels, where all app versions are deployed continuously. This would give us similar deployment experience like today, just more structured.

IMO `Production Current` and (some day if/when we have it) `Production LTS` release channels would not make much sense, as we would loose ability to continuously deploy new features to MCs.

> Do we need different repos for different channels?

I would really like to avoid this and keep the same repos that we have now.

> if we would describe how we would release and promote a new version of a cluster-$provider app

Without going into depth, every change would probably go through all release channels, meaning:
- We first update clusters in `Testing`
- Then we update clusters in `Staging`
- Then update clusters in `Production Edge`, and so on.

> Right now it describes a generic set of components, but we do package things in a specific way, and we use specific tooling.

Yeah, details here are missing. Some thoughts here:
- We keep same repos if possible (or at least with minimum changes to what repos we have)
- We implement multi-release-channel support in places where we specify apps/components and their versions. For example:
  - In app collection, instead of specifying `app-x` and version `1.1.0`, we specify `app-x`, version `1.1.0` for `Testing`, version `1.1.0` for `Staging`, version `1.0.1` for `Production Edge`.
  - In default apps / cluster apps, where we specify default apps, similarly in Helm values we can specify a version for every release channel, and then in the template we apply one of those versions, based on the release channel that the cluster is using.

nprokopic:
> In app collection, instead of specifying app-x and version 1.1.0, we specify app-x, version 1.1.0 for Testing, version 1.1.0 for Staging, version 1.0.1 for Production Edge

So here in the app collection you get to see which apps you have in that app collection, and for every app which version is in which release channel. The act of promoting an app version to next release channels means updating app version for some release channel.

In the above example:
> app-x, version 1.1.0 for Testing, version 1.1.0 for Staging, version 1.0.1 for Production Edge

We have app-x 1.1.0 in Testing and Staging. We promote 1.1.0 to Production Edge my updating app collection to have:
> app-x, version 1.1.0 for Testing, version 1.1.0 for Staging, version 1.1.0 for Production Edge

nprokopic:
In the end, as somebody who is is working on app-x, you just have to release new app version like today, and automation deploys it to all release channels at the right time per process that we define here in the RFC (starting from Testing release channel, then Staging, then Production Edge, then Production Current).

> starting from Testing release channel

Per current text this means that you have to create a new beta release (with `-beta` suffix), or RC (with `-rc` suffix), or GA release (without suffixes), and then that release will get deployed to appropriate release channel at appropriate moments (per defined release process for every release channel).

(suffixes are the same like for Kubernetes and Cluster API, where we would define similar quality expectations for beta/RC/GA)

nprokopic:
> and automation deploys it to all release channels at the right time per process that we define here in the RFC

Of course, we would have ability to stop the release promotions to next release channels, in case we discover issues that we don't want to deploy to more clusters.

release-channels-and-staged-deployments/README.md
```diff
@@ -0,0 +1,379 @@
...
+It starts with an overview of what we have today, and then continues with how we can structure releases an deployments in the future.
+
+## 1. What and when we release and deploy today?
+
+### 1.1. What are we deploying?
+
+1. cluster-x app
+  - Cluster resources
+  - Most critical apps: CPI, CNI, CSI
+  - Owners: LM, CI, OPI
```
pipo02mix, line 14: I am not familiar with these acronyms, can you link to Glosary or explain in the first appearance?

release-channels-and-staged-deployments/README.md
```diff
@@ -0,0 +1,379 @@
...
+
+For release channels we define 3 stability levels, each of which has one or more release channels:
+- Testing
+- Staging
+- Production
+
+For production stability level we define 3 release channels, which have different release cycles, but all are considered to be production:
+- Production Edge - fast rolling releases
+- Production Current - normal point releases
+- Production LTS - slow point releases
```
pipo02mix, line 95: I dont like the naming here since it can be very confusing with customer env naming. In the end, we will not force customers to set a real `dev` env cluster as `Testing` release channel. So example a customer can set `Testing` channel in their `Staging` cluster if they consider to that channel as appropriate to the staging cluster. I would advocate for something along the lines of `alpha`, `beta` and `stable` instead of.

PR #70, opened 2023-07-13
Adds decision process RFC

JosephSalisbury:
This PR adds an RFC describing a potential RFC based decision making process.


JosephSalisbury:
Asking for review from sig arch because we'd most likely own this process, team team because they're doing decision making stuff recently, and team horizon because they should know what i'm up to :D

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+```
+---
+type: general
+---
+```
+
+## Decision Making Process
+
+The two main aims of the decision making process are to ensure that the relevant groups have seen and approve the decision, and that the implementation of the decision is being tracked.
```
MarcelMue, line 37: I would like to see some more in depth proposal here - there are multiple issues which keep re-appearing with RFCs:
- How long should an RFC be open until a decision is reached?
- How do you deal with people who do not respond in a reasonable timeframe (what is considered reasonable?)?
- What to do (roughly) if a decision can not be reached? (probably schedule a meeting)

I am fine if you don't want to include this here directly but to me, answers to these questions will be necessary to have a smoother RFC process going forward.

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+This RFC follows on from the RFC repository README (see [here](https://github.com/giantswarm/rfc/blob/b3f4d2af6ff8965e542baae7d0e8fadb9163f356/README.md)).
+
+This RFC aims to expand on the current RFC process by adding a lightweight decision making process.
+The aim of this decision making process is to help enable, improve, and streamline technical decision making at Giant Swarm.
+
+## RFC Types
+
+To allow for a decision making process without removing the current free-form nature of RFCs, RFC types are proposed. The following changes will be introduced:
+
+- All RFCs will have a YAML header.
```
lyind, line 21:
```suggestion
- All RFCs will have a YAML header (front matter)
```

decision-process/README.md
lyind, line 37:
```suggestion
The two main aims of the decision making process are to ensure that the relevant groups have seen and approved the decision, and that the implementation of the decision is being tracked.
```

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+---
+```
+
+As another example, this RFC itself also implements the `decision` type RFC YAML headers.
+
+## Automation
+
+YAML headers will be checked and ensured by CI, specifically:
+
+- That the `type` field is present, and is either `general`, or `decision`.
```
lyind, line 74:
```suggestion
- That the `type` field is present and has the value `general` or `decision`.
```

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+- That if the `type` field is `decision`, that the `stakeholders` and `issue` fields are present.
```
lyind, line 75:
```suggestion
- That the `stakeholders` and `issue` fields are present, if the `type` field has the value `decision`.
```

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+- That if the `stakeholders` field is present, there is an approval from one member of each of the listed groups.
```
lyind, line 76:
```suggestion
- That there is an approval from one member of each of the listed group, if the `stakeholders` field is present.
```

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+- That if the `issue` field is present, that it is a valid GitHub URL.
```
lyind, line 77:
```suggestion
- That the value of the `issue` field is a valid GitHub Issue or PR URL, if set.
```

lyind (approved):
LGTM (only minor improvements)

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
+---
+type: decision
+stakeholders:
+    - sig-architecture
+    - team-horizon
+    - team-team
+issue:
+---
+
+# RFC Decision Making Process
```
piontec, line 10: It would be nice to add how RFC (comments, discussions) are related to ADR (short notes about taken decisions)

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+---
+type: general
+---
+```
+
+## Decision Making Process
+
+The two main aims of the decision making process are to ensure that the relevant groups have seen and approve the decision, and that the implementation of the decision is being tracked.
+
+To that end, the `type` field can also be `decision`. It is understood that when a `decision` RFC has been merged, that the decision has been taken and agreed on.
```
piontec, line 39: I see a contradiction here. RFC is "request for comments", nor for making decision. When, after the discussion, a decision is made, it should be recorded in ADR (if it was about architecture), PDR (if about product and its features) or as a part of our docs (standards, guides).

decision-process/README.md
JosephSalisbury: generally yes, but i think i'm only going to accept _issues_, i'm not sure how an RFC would be tracked by a PR

decision-process/README.md
JosephSalisbury: yeah, i tried to get some language down for these, but i'm not happy enough with it to make it "official"

generally:
- i don't want to give a specific timeframe / deadline for how long an RFC should be open - i could believe it's very dependent on complexity of the topic / number of participants etc. 
- i think if people do not respond / if decision can't be reached, we need to schedule a meeting, sure. i could believe there's some case-by-case work here until we figure / smooth it out

my general thinking is to use sig arch sync to review / push hanging RFCs - i.e: if we have open decision RFCs, let's bring them up in sig arch and see what we need to do to unblock, and then formalise that going further

sounds good? i know this is non-concrete, but i think we need to live it a bit

decision-process/README.md
JosephSalisbury: i'm not sure yet - i have a feeling that RFCs should replace ADRs (because they include more of the backstory), but i could also believe that they're both useful tools for us to have. i kind of want to live with decision rfcs for a bit, and see if they're somehow more useful

decision-process/README.md
JosephSalisbury: see https://github.com/giantswarm/rfc/pull/70#discussion_r1266904630 - i think RFCs could/should replace ADRs in the long-term. PDRs / docs are somehow out of scope for this, i feel

pipo02mix:
what about offering a template that helps with frontmatter? I have thought also suggesting some structure in the content can help to describe the case

As an idea, it could have (taken from KEPs):
```
    # Motivation
    ... 
    # Proposal
    ... 
    # User stories
     ...
    # Risks and Mitigations
    ... 
    # Implementation ideas
    ... 
```

JosephSalisbury:
@pipo02mix i reckon we'll eventually end up with something like this, i just don't want to prescribe it too early

piontec:
> @MarcelMue / @piontec any other thoughts? blockers?

TBH, I still would prefer better discussion about RFCs vs ADRs and their future. For me, they are not the same things, but if you want to merge them, let's make sure that the main objective of an ADR is still there (short and informative note about the decision taken, wihtout going into details).

MarcelMue:
> @MarcelMue / @piontec any other thoughts? blockers?

I am generally ok with it - just feels like it doesn't address some of the known issues of our RFCs in the current form. If you want to live it more first, then fine with me.

JosephSalisbury:
@piontec i think it's entirely reasonable to have decision RFCs that look a lot like our current ADRs (i.e: short, information, just the decision). we can also go into further explanation if needed

@MarcelMue agree with the direction, i just want to take smaller steps

either of you wants to drop an approval?

MarcelMue (approved):

nprokopic (approved):
> i just want to take smaller steps

In the name of this :point_up: :smiley:

TheoBrigitte:
> 
> As an idea, it could have (taken from KEPs):
> 
> ```
>     # Motivation
>     ... 
>     # Proposal
>     ... 
>     # User stories
>      ...
>     # Risks and Mitigations
>     ... 
>     # Implementation ideas
>     ... 
> ```

I like the whole thing, but I +1 on fer about the idea of proposing a more structured way to write RFCs, this would help us going through RFCs in the future.

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+This RFC aims to expand on the current RFC process by adding a lightweight decision making process.
+The aim of this decision making process is to help enable, improve, and streamline technical decision making at Giant Swarm.
+
+## RFC Types
+
+To allow for a decision making process without removing the current free-form nature of RFCs, RFC types are proposed. The following changes will be introduced:
+
+- All RFCs will have a YAML header (front matter)
+- The YAML header will contain the following mandatory fields:
+    - `type`
```
AndiDog, line 23: Date should be required, and since it should be easy to change/fix the date, it should go into the YAML header and not into the directory name. And there could be multiple dates such as `created: 2023-08-30`, `approved: 2023-09-24`, `deprecated: 2024-05-10`. Some companies including us even tried `must_be_reviewed_until`, but that attempt will fail until we have an actual process in place that reminds the reviewer group automatically, and even then time pressure is very dependent on the case and may still be a failure, so I wouldn't recommend this initially.

Some CI job should ensure correctness, and if a new field is introduced later on, could align the existing files automatically.

There could also be fields that relate RFCs to the others, such as `followup_for: [ ... ]`, `related: [ ... ]` or `replaces: [...]` (newer version of a deprecated decision). This would make it easier to render the information in a structured way, rather than people writing "Further reading" bullet-point lists in free form somewhere in the "article", impossible to parse.

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Decision Making Process
+
+The two main aims of the decision making process are to ensure that the relevant groups have seen and approved the decision, and that the implementation of the decision is being tracked.
+
+To that end, the `type` field can also be `decision`. It is understood that when a `decision` RFC has been merged, that the decision has been taken and agreed on.
+
+If the `type` is `decision`, the following fields would also be mandatory:
+
+- `stakeholders`
```
AndiDog, line 43: "stakeholders", similar to "owners", is a very vague word. Since the purpose of this field is described below, a clear name might be `required_approvers`. 

What's very important for RFCs are visibility and searchability. If I want to list all RFCs that relate to KaaS, that needs to be possible in two clicks. So for example, this could be `areas: [...]` or `topics: [...]` where a fixed set of topics could be chosen from (KaaS, product, sales, ...).

decision-process/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `issue`
```
AndiDog, line 44: All such fields should be plural. There may be multiple issues, even multiple epics relating to the RFC.

decision-process/README.md
AndiDog: True, the terms RFC/ADR/PDR have a strong connotation in the IT worl, and we already have those terms in use.

\<strong-opinion\>
Neither of the terms is great, and in Giant Swarm's usual way, we should make the best use out of the solution, no matter how it's called. I don't mind finding a new word, but it doesn't matter as long as we have a great and speedy decision process. For comparison, we don't use ["OKRs"](https://rework.withgoogle.com/guides/set-goals-with-okrs/steps/introduction/) and I'm happy that TeamTeam never even mentioned that horribly-working concept to us, but we track quarterly goals in simple "3 stones", and if we ever strongly believe that a little more tracking or complexity is needed, we will adapt. We didn't just take the off-the-shelf – and I have to say it again: _totally not working_ – concept and throw it at employees. From what I saw since I joined, many smart people had great ideas to take the good parts of something and apply it to the swarm, from which we have benefited a lot compared to companies that struggle to apply some allegedly-standard concept like RFCs.

One of the reasons why others have failed is to split RFCs from ADRs. Why is it so hard to put everything necessary for a decision into one easily-findable Markdown file? Why do companies write pages-long articles that could be summarized in some bullet points? Why do tech companies write down implementation details in an ADR when either 1) it could be a few bullet points or 2) the implementation will anyway happen on the spot as with regular work tasks, is therefore not plannable, and should simply go into issues that refer to the one central RFC?
\</strong-opinion\>

Suggestion: we decide for one term, which could even be a funny, GS-specific thing like RFS (Request for Swarm), and use it only in the way defined in _this RFC_, without trying to use anything from how other companies use RFCs, or how we had previously utilized the RFC/ADR/PDR model/intranet/repo. Let's start from scratch. One CI-checked YAML+Markdown format, clear state diagram (which transitions can happen, which fields are necessary in each), one file (_this one_, moved into a more discoverable place, replacing the README) documenting the process.

With that in mind, I unfortunately don't see a decision making process in this document (at time of reading i.e. 42daf33bc6c87a86064d65779fd9befb674596f0), despite the heading "Decision Making Process". Right now, we're defining a file format for the decisions, but we need to write down the whole process. Most probably it will be quite concise if we only describe the different states (e.g. `open` = in review, `approved`, `deprecated`) and what is required in the file, and required to be done, in each of those. I'm offering my help to write and implement that, but based on experience, we first need to have concensus that this file-based model will work and be used by the whole company, and that we'll move existing RFCs into the new format.

decision-process/README.md
pipo02mix: But IMO we need to start from some point and then refine it. Like @AndiDog I missing here a bit of how should I proceed if I want to take a decision. Andreas also offer to collaborate maybe can give you a hand

PR #71, opened 2023-07-31
Add base for CRD management RFC

uvegla:


manage-essential-crds-via-mcb/README.md
```diff
@@ -0,0 +1,194 @@
...
+essential:
+
+- https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+- https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml
+
+As with `kustomize` we can patch delete some of those CRDs from the rendered Helm chart, but that is an exclude list
+instead of an allow list, therefor needs more maintenance. However, I think pre-installing all the CRDs in this Chart
+in order to avoid this issue is fine too.
+
+Or for `cilium` we do not have a Chart, but could use the remote resources approach like:
```
puja108, line 87: I'd prefer we control them, and have some chart / local mirror

manage-essential-crds-via-mcb/README.md
```diff
@@ -0,0 +1,194 @@
...
+### Update build scripts
+
+Beyond adding the new targets to build the CRDs as shown above, the `ensure-versions` target in MCB needs to be updated
+to point to the new local cache folder of Flux CRDs. (Tho with the proxy in better state in CN cluster we could try to
+go back to remote resources as well, could also be done separately.)
+
+### Considerations for Giant Swarm CRDs
+
+None, I think we should definitely do this and as an extra win we can remove `opsctl ensure crds` for good and our
+CRDs will be rolled out much faster too from now on.
```
puja108, line 180: <3 burn opsctl ensure

fiunchinho:
I think this is a great idea to manage CRDs that we need on MCs. Using this approach to install CRDs instead of `opsctl ensure` will simplify things a lot.

But I would use a different approach for CRDs that we need on both MCs and WCs. For this scenario, I suggest we install required CRDs before anything else, using dedicated charts. We already have those charts for [VPA CRDs](https://github.com/giantswarm/vertical-pod-autoscaler-crd) and [Prometheus CRDs](https://github.com/giantswarm/prometheus-operator-crd).

For cilium, I think it's simpler to rely on upstream operator installing the CRDs. We use `HelmRelease` / flux dependencies to make sure cilium is installed before any other apps that may use the cilium CRDs. So I wouldn't include the cilium CRDs on your proposal, and I would manage them the same way for both MCs and WCs.

uvegla:
@fiunchinho We can utilise the CRD only charts here too by using kustomize helm chart rendering.

```yaml
# base/crds/vpa/kustomize.yaml
helmCharts:
  - name: vertical-pod-autoscaler-crd
    includeCRDs: false
    namespace: giantswarm
    repo: https://giantswarm.github.io/control-plane-catalog/
    releaseName: vertical-pod-autoscaler-crd
    version: v2.0.1
```

I am thinking they do not necessarily need to be part of the `all` flux kustomization tho (VPA probably yes, also Prometheus because of the service monitor migration, but maybe not cilium), cos they are misleading to some of our charts maybe if the CRD is present but we do not actually install the controllers for them (those nasty has api version checks).

With the upstream operator installing the CRD my "problem" is the above fact of us using the api version checks in our Helm charts. Because Helm is not constantly reconciled by nature you might end up with a release that does not have e.g. Cilium NP depending on when you installed the app.

nce (approved):

PR #73, opened 2023-08-14
Add base for GS CRD management via Flux extension RFC

uvegla:


giantswarm-crd-managmenet-via-flux-extension/README.md
```diff
@@ -0,0 +1,87 @@
...
+- https://github.com/giantswarm/giantswarm/issues/27597#issuecomment-1632630001
+
+#### Pros
+
+- Sort of drop-in and implicit for customers (tho this can be interpreted as a con too in some use cases)
+- We probably need the MCB source soon anyway
+
+#### Cons
+
+- We take away control and the extension point from customers
```
piontec, line 70: Is it a good thing or bad thing? Do we want customers to mess up with the way we deploy our CRDs?

giantswarm-crd-managmenet-via-flux-extension/README.md
```diff
@@ -0,0 +1,87 @@
...
+- Complex and leads to some duplication in MCB repository
```
piontec, line 71: I think shared parts can be also abstracted away to separate bases and reduce duplication - but at a cost of some complexity

giantswarm-crd-managmenet-via-flux-extension/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Sidestep - Handle Vintage provider specific CRDs from new `crds-vintage` kustomization
```
piontec, line 73: I think I like this one the best, but it's really hard to tell

giantswarm-crd-managmenet-via-flux-extension/README.md
uvegla: They can use the same mechanism to deploy their stuff instead of reinventing the wheel. If we want to give them control over the CMC eventually - which we do - I feel hiding stuff from them contradicts that approach. Hiding means we reconcile stuff directly from MCB.

giantswarm-crd-managmenet-via-flux-extension/README.md
uvegla: I am leaning towards the first one as that way is:
- similar to what we already do with other stuff, like catalogs, crossplane, flux-extras
- gives control and visibility to customers
- same mechanism for all clusters, not making vintage special

but his one is the 2nd I like the most beyond that.

PR #74, opened 2023-08-16
Policy orchestration subsystem for security platform

stone-z:
This RFC describes Shield's intent to create a Giant Swarm abstraction layer on top of the various types of security (and soon, non-security) policies currently distributed to secure our clusters.

The goal of this review is to gather feedback on the proposed implementation, voice any blockers, and minimize duplication of effort.

Shield intends to start working in this general direction immediately to meet urgent customer needs, but the RFC will remain open for review until the end of September to collect input. The RFC will be closed on or after October 2nd when no "Request changes" reviews remain.

A second RFC describing the scope for our immediate customer need will be opened shortly.


puja108 (commented):
There's a few topics that might merit discussion here:
1. Reasoning behind this new product
2. Architecture of the solution
3. Specific details of the implementation, i.e. CRD design

I'm going to concentrate this comment on the 1. and I think this RFC might also need a bit of focus or a way to coordinate the different discussions, so we don't get too much confusion in.

On the 1. I am quite convinced that this is basically a fully GS product we aim to build here. And from what I know there's nothing OSS on the market for this. From my limited discovery in this space I can also imagine there being a market for this, i.e. a real pain this solves. The strongest reason to have this for me is that we will have tools overlapping in enforcment of policies, be that k8s core, kyverno, falco/tetragon, Cilium, Linkerd,... and on top of that these tools might change over time (e.g. enforcement of certain things might move from kyverno to intree k8s, or from falco to tetragon).
That said, this would need to be properly discovered as a product with talking to customers, the market, experimenting and finding proof it will be worth it.
Additionally, we might need to work out somehow if this work can be upstreamed or at least somehow better open sourced to make it less GS specific, which is often a risky thing.

In general, with this RFC the whole concept of another layer of abstraction makes more sense, so do not take my rather negative comment from the other RFC as a downside to this one.

puja108:
btw, something that gets close to this is https://open-cluster-management.io/getting-started/integration/policy-framework/ It seems to be mainly implementing Gatekeeper resources, but looks to also be extensible. It integrates tightly into the central OCM management though, but that might be also sth we should check out in general.

stone-z:
I'd maybe explicitly say that while this orchestration very heavily prioritizes customer experience, it is only partially intended _for_ their direct use. The orchestration is also to make it easier for Giant Swarm (/Shield) to manage the policies that we've agreed to manage, given that as Puja mentioned there is overlap between control points and we expect to need to swap out the actual tools over time. The goal isn't to create a generic way of representing policies or exceptions, it's a way for our platform users to use _our_ policies.

We've only recently become aware of open cluster management and will take a deeper look. It is clear from the current APIs though that at least right now it doesn't handle all the cases we're most concerned with, specifically exceptions. If that project ends up doing what we need, it could be the replacement for this abstraction.

puja108:
Even if just limited to our policies, there's still a bit of a risk (to be confirmed with actual customers) that customers perceive this as too proprietary and a lock-in, but as I said, I can also see the huge benefits of it in the future, and it might be a good way to sell it to customers mentioning those benefits.

stone-z:
> customers perceive this as too proprietary and a lock-in

Sure, it is very Giant Swarm specific. But if they want, they can skip the whole thing and directly manage Kyverno policies and exceptions. They would just need to manage them in that case (i.e. they would have to migrate themselves to whatever is next)

policy-orchestration/README.md
```diff
@@ -0,0 +1,296 @@
...
+# Policy Orchestration System
+
+> [!NOTE]
+> :bulb: This RFC describes a subsystem of the security platform which is an intended long-term approach to improving policy management.
+
+## :beginner: Current state / context
+
+Giant Swarm manages several third-party applications which provide security features to our customers.
+
+This management includes a set of default policies for each tool, which we recommend, deploy, and manage based on the perceived value for most / all customers.
```
pipo02mix, line 18: Maybe an example, reference to a existing policy, can help to understand what do you mean

pipo02mix:
Additionally I have 2 concerns, 

1) I read we don't want to implement new policy language but does it mean we need to create a CRD with all possible use-cases as part of the spec? (registries, pss, volumes, ...)

2) Have we considered how many CRs and how them will affect big cluster API? maybe checking how many policies we already have and how many resources we would have hypothetically when we implement a solution like that. 

And I have open question, should be this managed in a central point (MC) or per WC?

stone-z (commented):
> Additionally I have 2 concerns,
> 
> 1. I read we don't want to implement new policy language but does it mean we need to create a CRD with all possible use-cases as part of the spec? (registries, pss, volumes, ...)

1. The GS Policy spec is mostly just a name. We map that name to multiple specific implementations with metadata and within the internal orchestration logic. For example, say a customer enables the GS Policy `require-trusted-registries` and create a GS PolicyConfig for that policy containing only the registry `docker.io`. The `policy-meta-operator` contains our internal logic for finding which specific policies implement the `require-trusted-registries` control. Because of differences in tooling (which is the whole point here), that logic can look slightly different. For Kyverno, we can use metadata on the Kyverno Policy to identify it. For Falco, we can't do that, so we will need custom logic.

>  2. Have we considered how many CRs and how them will affect big cluster API? maybe checking how many policies we already have and how many resources we would have hypothetically when we implement a solution like that.
2. Great point. I actually expect to be able to reduce the overall storage size and number of distinct CRs with this approach in the long term. for starters, the spec of the GS types is much slimmer than e.g. the Kyverno native types. By having the native types managed automatically, we can also combine them. For example, instead of requiring one Kyverno PolicyException per workload, we can generate a single PolicyException containing many workloads. There is still storage needed for the GS types, but the syntax is smaller and it allows us to create more compact downstream resources and also automatically clean them up and deploy only what is necessary.

> And I have open question, should be this managed in a central point (MC) or per WC?

The components run per-WC, since WCs for a given MC have different policy needs (e.g. one MC likely has both dev and prod clusters) and to support self-managing / air gapped environments. However, there are lots of options for extensibility; we expect customers will manage resources with gitops right now, but it would be possible to manage them through other means, like by creating them directly from happa / a dev portal or from customer-managed systems. For example, some customers already use a request system for allowing firewall / network policy exceptions. These portals could directly create GS PolicyExceptions.

pipo02mix:
thanks for the clarification! BTW should @giantswarm/sig-architecture be involved?

stone-z:
> thanks for the clarification! BTW should @giantswarm/sig-architecture be involved?

We did an architecture review session today :slightly_smiling_face:

pipo02mix:
happy to read the outcome :)

Strigix (approved):

PR #76, opened 2023-09-21
Add RFC on crossplane capi import

mproffitt:
Adds a page discussing the architecture of importing EKS/AKS/GKE clusters to CAPI using crossplane


mproffitt:
There's still a couple of points to add to this before it's ready for review

- differentiating cluster types
  e.g. Capi vs capi eks vs adopted 

- cloud specific Crossplane types
  e.g. The types used to read from each cluster 

These should be added in the morning, after which I'll open it for review

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+For this to be effective, there are then three primary components to the import
+process.
+
+1. Security of the cloud account.
+
+   We work on the basis that we only have read only access to see only components
+   necessary for the import of the cluster.
+
+1. The capability of reading existing customer resources
```
JosephSalisbury, line 34: nit: can we have 1, 2, 3 here? :D

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+For the purposes of this RFC, point 1, Security of the cloud account is
+considered out of scope and will be managed on a cloud by customer basis as part
+of the implementation.
+
+The caveat to this is that we are under the assumption that only read access
+will be granted and that the customer may place strict limitations on our
+visibility of components inside their accounts.
+
+Whilst there is some functionality to read resources directly from CAPI, it
```
JosephSalisbury, line 46: i don't follow here - what does "read resources directly from CAPI" mean?

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,257 @@
...
+![eks-architecture](./crossplane-capi-relationships.png)
+
+### Composition and Definition
+
+The crossplane composition for this importer follows the definition which allows
+for a minimum of the following to be defined
+
+- The region the cluster is built in
+- The cluster name
+- The nodegroup name
```
JosephSalisbury, line 117: what about with clusters with multiple nodegroups?

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,257 @@
...
+During the generation process, there are certain details that may (excluding the
+use of composition functions) require the editing of specific blocks of the
+composition.
+
+The reason for this is that there is no looping inside the composition and
+details returned from the cloud do not directly correlate to information required
+by CAPI.
+
+The example of this inside EKS is the subnet groups required by the
+`AWSManagedControlPlane` resource which then require hard coded specification
```
JosephSalisbury, line 139: dumb question - this means even with the XRD, a human has to go in and edit some stuff?

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,257 @@
...
+compatible with our own deployments via `App-Platform` we may need to enhance
+our `App` CR to accept a secret data key, or include this location in the list
+of key locations checked for a `kubeconfig` value.
+
+### Discovery
+
+With the introduction of EKS/AKS/GKE clusters to CAPI management clusters, we
+need to be able to differentiate between the different types and usage.
+
+This can be achieved through the use of the type of cluster defined as
```
JosephSalisbury, line 200: why do we need the cluster type to determine whether a cluster was imported? we only need `k8s.io/managed-by: crossplane`, right?

JosephSalisbury (commented):
first pass random notes / thoughts, but looks legit

crossplane-capi-import/README.md
mproffitt: Using AWS/EKS clusters as the example, the default check to determine if a cluster is standard CAPA or CAPA EKS would be to use the `infrastructureRef` as the type is different between these, however the `infrastrucureRef` kind would be the same between a fully managed CAPA EKS cluster and an imported EKS cluster.

Simply checking for the presence or absence of the annotation only tells you if the cluster is adopted or not but doesn't allow for separation between fully managed EKS and a standard CAPI cluster.

The inclusion of the table below was to specifically highlight how a CAPI cluster would appear in each of the given scenarios and how you can differentiate between what's fully managed and what's imported.

crossplane-capi-import/README.md
mproffitt: At present yes, on our current version of crossplane it is not possible to automatically generate this.

In the next iteration I'm looking at how we can solve this using Composition Functions to remove the remaining requirement on human interaction and make this more automated.

JosephSalisbury: cool - with us wanting to productionise this, we'd need to make sure that it's hands off from an operator point of view

crossplane-capi-import/README.md
JosephSalisbury: gotcha, nice

crossplane-capi-import/README.md
mproffitt: I've added a section on multiple nodegroups.

In essence multiple nodegroup clusters do not require full reflection for the cluster to be imported or for us to interact with, and deploy to the cluster

Trying to track and generate machine pools for each nodepool is IMO more volatile than simply targetting the primary nodegroup for an imported cluster and with crossplane in its current version, we're not yet in a position whereby we can dynamically react to nodepools being added to, or removed from the remote cluster.

JosephSalisbury: had to think about this a little bit

somehow we have more static and more dynamic resources (i.e: the cluster cr is fairly static, node pool crs are more dynamic). i think importing the more dynamic stuff could mean things start getting a bit weird - e.g: if node pools get changed over time, we'll have a different view of them from out side

i also don't think some "importing things repeatedly over time" approach makes a lot of sense

do we want / need some representation of the node pools / node for this to work (i.e: to be able to use the app platform on top)? could we get status of node pools / nodes from the capi controllers (to avoid the importing problem entirely)? 

does that make sense?

mproffitt: I think this is one of the more product based questions of "How much reflection do customers require" and I don't really have an answer on this.

A couple of points here though. 

> i also don't think some "importing things repeatedly over time" approach makes a lot of sense

On this I am inclined to disagree. I think it makes a lot of sense, the MC should contain a reflection through reconciliation of how the WC infrastructure appears at any given time but here I think the question is less "should we do it" (to which the answer IMO is always yes) but more "Do we need it" and here the answer is No. We can, in theory skip this entire section and still have a representative cluster that can be interacted with through the app platform.

> could we get status of node pools / nodes from the capi controllers

Again, this is a no. Trying to use CAPI to answer this problem is a direct violation of the `clusters.x-k8s.io/managed-by` contract which states:

> The provider InfraCluster controller must:
>
> - Skip any reconciliation of the resource.
> - Not update the resource or its status in any way

My opinion here is that I would like to see this problem solved. To the question of whether or not we need to solve this now, then no. This can be tomorrows problem.

JosephSalisbury: > This can be tomorrows problem.

i like this approach, we don't need to import nodepools now, so let's not. bam!

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,301 @@
...
+The purpose of this is to provide a mechanism whereby customers may come to
+Giant Swarm on a Bring Your Own Cluster (BYOC) basis such that we can support
+and manage apps on the customers cluster without compromising the integrity of
+their existing infrastructure deployment mechanisms. That is to say, customers
+operate and manage their own infrastructure, whilst we provide support and
+deployment strategies for our own applications to run inside the customers
+cluster.
+
+Throughout this RFC, where the words __import__ and/or __adopt__ are used, they
+are to be read as "unmanaged".
```
puja108, line 17: I'd rather we agree to only use "import" for unmanaged, as we tried to define "adopt" as "we adopt and manage lifecycle"

crossplane-capi-import/README.md
puja108: AFAIK currently we do not rely on nodepool information, however, especially with Observability we've been thinking if some information about cluster size might help us with right-sizing our components better.

That said, future problem as long as no current GS platform functionality depends on it.

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,301 @@
...
+To work around potential limitations here, crossplane leverages the
+`cluster-auth` capabilities and stores this secret as `<cluster_name>-kubeconfig-cluster-auth`
+
+This secret contains a slightly different structure from CAPI and places the
+kubeconfig at the secret data key `data.kubeconfig`.
+
+This cannot be controlled via crossplane and in order to make this secret
+compatible with our own deployments via `App-Platform` we may need to enhance
+our `App` CR to accept a secret data key, or include this location in the list
+of key locations checked for a `kubeconfig` value.
```
puja108, line 237: The issue here is that we might depend on the kubeconfig format and naming conventions in other places in our product (e.g. operators like PMO needing to talk to the cluster).

We need to check if that is the case, and might need a plan to "reformat" the secret somehow (I've seen some use cases where people did that via Kyverno or other webhook)

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+The `kubeconfig` secret is normally generated by CAPI controllers and stored
+in the cluster namespace as `<cluster_name>-kubeconfig`
+
+As CAPI should be running in a non-reconcilling mode for imported clusters,
+it is not known if this kubeconfig will be generated automatically by CAPI or
+whether it needs to be imported separately.
+
+To work around potential limitations here, crossplane leverages the
+`cluster-auth` capabilities and stores this secret as `<cluster_name>-kubeconfig-cluster-auth`
```
puja108, line 229: This is assuming we have access to that, we need to make sure access is given within the IAM of the cloud provider.

In general we need a concrete IAM ARN or sth like that, that lists exactly what we need to access. Ideally even a pre-made Terraform or other script that creates that when the customer runs it.

puja108: Another assumption here is that customer will be fine with giving us cluster-admin access within WCs. We'll see if that assumption holds.

crossplane-capi-import/README.md
mproffitt: The functionality to manage tracking subnets from a crossplane composition function now exists: https://github.com/giantswarm/crossplane-fn-generate-subnets/

puja108 (commented):
RFC generally LGTM, my comments go more towards two topis that we might need to follow-up on with the team:
1. What do we need to do to make this a full product feature that is documented and works with no (or few) manual interventions?
2. Does the result of this work for all platform teams?

crossplane-capi-import/README.md
mproffitt: Agreed. I added a clarification here that `import` refers to unmanaged clusters whilst adopt refers to full management.

crossplane-capi-import/README.md
mproffitt: I understand the gist of this but have concerns with trying to mutate a secret controlled by crossplane.

As an alternate approach, it would be possible to use either crossplane directly, or ESO to copy the `kubeconfig` value out of the generated secret and map this directly into a CAPI formatted secret.

In this scenario, my preference would be to use crossplane and try and keep the whole process self contained to a single product.

crossplane-capi-import/README.md
mproffitt: For the first comment, I have IAM policies for AWS (and largely GKE) that restrict to just the required resource types. 

These could be used as a base template for the customer but may need additional filter adding for targetting specific named or regional resources. For example in scenarios where customers may want us to only be able to target clusters in the `eu-west-1` region, or only clusters starting with a given name prefix.

mproffitt: For the second comment, for the purposes of the RFC, I have avoided to a large degree making assumptions about the internals of the cluster. 

This can be controlled by the customer when assigning the `kubeconfig` to roles inside the cluster and I can add a section on how this can be managed, but in general the RFC was specifically about the process of importing infrastructure.

crossplane-capi-import/README.md
mproffitt: Added a section on how the composition function solves this problem, carrying the link to the PoC and the upstream issue that blocks current further enhancement.

crossplane-capi-import/README.md
mproffitt: So there is one place we do rely on nodepool information, that being within Happa and towards this, @gusevda has a PR bringing this to light here: https://github.com/giantswarm/happa/pull/4525.

For an imported cluster do we care about this information and without it, what impression would that present to customers who use Happa as a product if this was to vary between the different cluster types? (CAPA, CAPA-EKS, Imported-EKS)

My overall feeling here is yes. We should be able to reflect that information and where possible, all cluster variants should reflect / be presented in the same way. Technically however there are limitations to being able to capture this information via crossplane without resolution on https://github.com/crossplane/crossplane/issues/4141

JosephSalisbury: Somehow I think this depends on UX (ping @marians). Having to support "dynamic" information after importing clusters introduces some problems, but I'm not 100% certain how this should look like from a product point of view

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+To achieve this lookup, we use crossplane resources for the cloud being searched
+
+| provider-aws-eks                 | provider-azure-containerservice                               | provider-gcp-[container,compute]   |
+|----------------------------------|---------------------------------------------------------------|------------------------------------|
+| Cluster (eks.aws.upbound.io)     | KubernetesCluster (containerservice.azure.upbound.io)         | Cluster (container.gcp.upbound.io) |
+| NodeGroup (eks.aws.upbound.io)   | KubernetesClusterNodePool (containerservice.azure.upbound.io) | NodeGroup (compute.gcp.upbound.io) |
+| ClusterAuth (eks.aws.upbound.io) |                                                               |                                    |
+
+During the generation process, there are certain details that may (excluding the
```
piontec, line 136: Is this a one-off sync or is crossplane constantly reconciling this info from the cloud provider?

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,315 @@
...
+Crossplane requires a direct one to one mapping between Custom Resource and
+Cloud Infrastructure types. This makes it more difficult to directly reconcile
+multiple nodepools without building custom compositions for each variant of
+customer infrastructure we may encounter.
+
+There are two potential solutions to this problem:
+
+1. We develop a library of common structures that we encounter. This will be
+   built as part of discovery with a custom using boilerplate code to  enhance
+   the compositions for each customer.
```
piontec, line 193: But will a logic needed to map these structures to our library entries be manageable? Won't it become a pile of huge `if`s?

crossplane-capi-import/README.md
```diff
@@ -0,0 +1,315 @@
...
+| AWSManagedMachinePool  | AzureManagedMachinePool  | GCPManagedMachinePool  |
+
+### `kubeconfig`
+
+The `kubeconfig` secret is normally generated by CAPI controllers and stored
+in the cluster namespace as `<cluster_name>-kubeconfig`
+
+As CAPI should be running in a non-reconcilling mode for imported clusters,
+it is not known if this kubeconfig will be generated automatically by CAPI or
+whether it needs to be imported separately.
```
piontec, line 240: What does it depend on from the CAPI side?

crossplane-capi-import/README.md
mproffitt: Crossplane reconciles this information roughly every 10 minutes after initial creation. This may actually work in our favour as it allows for changes to remote infrastructure to be reflected back to the management cluster on a continuous basis

crossplane-capi-import/README.md
mproffitt: In this instance no. The idea of a library here would have been that we develop an XRC for each cluster configuration we found. E.g. one with 3 subnets defiend, one with 6, etc.

In my mind this was never really a viable option as it requires constant maintenance.

crossplane-capi-import/README.md
mproffitt: It requires a change to CAPI to ensure that the `cluster.x-k8s.io/managed-by` annotation is honoured where CAPI operators are installed.

Although not preferable, this should operate fine on clusters where the operators are not installed but the CRDs are.

crossplane-capi-import/README.md
puja108: In the WG kickoff we said we do not want consider this part of the MVP and will rather do sth like not show nodepools (or their size) and display a notice. We will still get that info in grafana dashboards and we can think of other ways we would expose it going towards more of a fleet management UI and use cases for that. With that let's have an issue for this for maybe revisiting, but close it and any related work towards support for it at least until we reach our next WG goals.

ljakimczuk (approved):

PR #77, opened 2023-10-03
New registries architecture

piontec:


single-registry-architecture/2023-10-registries.drawio.svg
pipo02mix, line 1: Is rettager triggering the `build pipelines`? maybe we could do a reference

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Specific problems and goals to solve
+
+More specifically, our aim is to:
+
+- Stop using Docker Hub by the end of 2023.
+- Simplify registries architecture.
+- Switch entirely to OCI based registries for both container images and helm charts.
+  - Flux developers have already declared that no new feature will come to Flux for users of HTTP based repos, only for
+  - OCI. Currently, we use a mix of HTTP and OCI repos, but charts discovery uses HTTP based ones only.
```
pipo02mix, line 52:
```suggestion
- Switch entirely to OCI based registries for both container images and helm charts.
  - Flux developers have already declared that no new feature will come to Flux for users of HTTP-based repos, only for OCI. 
  - Currently, we use a mix of HTTP and OCI repos, but chart discovery uses HTTP-based ones only.
```

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+- If possible, switch to single OCI provider.
+- Increase availability and lower the network transfer cost by providing a pull-through cache service.
+- Automatically clean old dev release artifacts.
+- If possible, use the registry provider to replicate registries between Europe and China.
+- Use a solution, that works natively with `containerd` and `helm` and doesn't need tricks like mutating webhooks,
```
pipo02mix, line 57: @silenteh has worked in pull-through cache lately and he can provide good insights about this

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+- <https://goharbor.io>
+  - Seems very complex to run, just [the list of needed components](https://goharbor.io/docs/2.2.0/install-config/) scares
+    me off. We want a simple cache, not all the bells and whistles with a ton of components we have to keep running.
+- <https://github.com/enix/kube-image-keeper>
+  - seems to support container images only, works by a webhook rewriting pod's image - doesn't meet assumptions.
+- <https://gitlab.cronce.io/foss/oci-registry>
+  - OCI compliant, works with `containerd` by registering as a mirror, no webhooks, optional S3 storage - seems
+    like exactly what w need.
+  - Tested, I unfortunately [couldn't make it work](https://github.com/mcronce/oci-registry/issues/14) even for a simple
+  - test case, doesn't seem to be well supported.
```
pipo02mix, line 102:
```suggestion
  - Tested it, but I unfortunately [couldn't make it work](https://github.com/mcronce/oci-registry/issues/14) even for a simple test case, doesn't seem to be well supported.
```

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+## Open questions
+
+Before we start on the implementation, we have to figure out answers to questions below:
+
+1. Should we use a single cache instance per installation, or rather per cloud provider + region?
+
+   The disk requirements for a cache might be significant and as a result cost of running an instance per MC can be quite
+   high in total. Additionally, we can expect many of the images to be the same for all clusters (MCs and WCs) that run
+   for the same provider. So, it seems running a single instance of the cache per cloud provider can be both much cheaper
+   and also more efficient and performant.
```
pipo02mix, line 183: If we only run for GS workloads I will vote for the region, it will save costs significantly.

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+   the best way to do it?
+
+1. Customers will stop using our Docker Hub subscription.
+
+   Most customers aren't probably even aware of this, but right now all the images the nodes of any cluster pull from
+   Docker Hub, they do so using our Docker Hub subscription. After implementing this change, we wil stop doing this.
+   Moreover, we will ensure unlimited downloads only for images that we manually picked and uploaded (using `retagger`)
+   to our new ACR registry. We have to explain and announce this change to our customers. Also, we probably have to
+   accept that some customers might want us to configure `containerd` to use their Docker Hub credentials to avoid
+   rate limiting. How do we do that?
```
pipo02mix, line 199: Big customers use their own registries and have policies to stick to that. We can ask AEs to check other customers to get an estimation of workloads affected

single-registry-architecture/2023-10-registries.drawio.svg
piontec: No, they are 2 separate sources of artifacts pushed to a registry.

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+Check the discussion [below](#open-questions).
+
+### Choosing OCI repository provider
+
+We checked for the single best registry and chose ACR. We did some evaluation before and you can read results here:
+
+- <https://github.com/giantswarm/roadmap/issues/2382>
+- <https://hackmd.io/gqF6PT53RcG_i4OhrmtWIw#Note-on-compatibility>
+
+Additionally, ACR offers th following:
```
uvegla, line 76:
```suggestion
Additionally, ACR offers the following:
```

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+There are at least a few solutions possible, below is a list of solutions evaluated:
+
+- <https://goharbor.io>
+  - Seems very complex to run, just [the list of needed components](https://goharbor.io/docs/2.2.0/install-config/) scares
+    me off. We want a simple cache, not all the bells and whistles with a ton of components we have to keep running.
+- <https://github.com/enix/kube-image-keeper>
+  - seems to support container images only, works by a webhook rewriting pod's image - doesn't meet assumptions.
+- <https://gitlab.cronce.io/foss/oci-registry>
+  - OCI compliant, works with `containerd` by registering as a mirror, no webhooks, optional S3 storage - seems
+    like exactly what w need.
```
uvegla, line 100:
```suggestion
    like exactly what we need.
```

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+- switch our CI/CD pipelines to not tag dev build artifacts
+  - use a tool (`skopeo`, `retagger`?) to replicate all (do we really need all?) the tagged versions of images we build from the old `giantswarm` repo to the new `gsoci`
+  - configure CI/CD so it uploads to both the old `giantswarm` and the new `gsoci`
+- deploy a new instance of `retagger` that will replicate all the configured public images to the new `gsoci` registry
+- switch container runtimes to use the cache as a source of images and upstream `gsoci` as a fallback
+- cleanup (after some time)
+  - delete our registries in docker hub and quay
+  - delete the old `giantswarm` registry in ACR
+- for this part, we keep China registry unchanged
+  - we reconfigure the `crsync` to replicate images from `gsoci` to `aliyun` in China
```
uvegla, line 143: My top of the head thought on this is the delay we introduce using crsync for this instead of uploading to a "primary" registry from the build.

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+High level migration plan:
+
+- prepare and deploy `distribution` cache instance, configure it for the upstream ACR registry
+  - ensure monitoring and alerting
+- switch our CI/CD pipelines to not tag dev build artifacts
+  - we want to use retention policies for helm charts as well
+- implement changes in the `app-operator`
+  - use OCI discovery instead of `index.yaml` from HTTP registries
+  - (optional) solve the problem of presenting only X (by default 5) most recent versions in the ACE CR
+  - introduce a concept of primary and fall-back registry
```
uvegla, line 168: We have these on the catalogs already, no?

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+so nothing is blocking implementation and a switch.
+
+High level migration plan:
+
+- prepare and deploy `distribution` cache instance, configure it for the upstream ACR registry
+  - ensure monitoring and alerting
+- switch our CI/CD pipelines to not tag dev build artifacts
+  - we want to use retention policies for helm charts as well
+- implement changes in the `app-operator`
+  - use OCI discovery instead of `index.yaml` from HTTP registries
```
uvegla, line 166: How does this -if so - affect creating Flux HelmRepositories instead of Catalogs?

single-registry-architecture/README.md
uvegla: Where should have host the per region one? Do we loose on network / transfer costs if we do so?

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+   high in total. Additionally, we can expect many of the images to be the same for all clusters (MCs and WCs) that run
+   for the same provider. So, it seems running a single instance of the cache per cloud provider can be both much cheaper
+   and also more efficient and performant.
+
+1. How to not tag dev builds?
+
+   ACR has a neat new feature that allows for automatic cleanup of untagged artifacts after a certain period. Currently,
+   we upload a ton of dev build images. This feature allows us to clean them up automatically, with zero effort on
+   the cleaning process on our side. Still, we have to switch our CI/CD configs to not tag dev builds in any way. What's
+   the best way to do it?
```
uvegla, line 190: I guess we can reference with hash, but I am not sure app platform supports it, e.g. AppCR.spec.version.

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+1. Customers will stop using our Docker Hub subscription.
+
+   Most customers aren't probably even aware of this, but right now all the images the nodes of any cluster pull from
+   Docker Hub, they do so using our Docker Hub subscription. After implementing this change, we wil stop doing this.
+   Moreover, we will ensure unlimited downloads only for images that we manually picked and uploaded (using `retagger`)
+   to our new ACR registry. We have to explain and announce this change to our customers. Also, we probably have to
+   accept that some customers might want us to configure `containerd` to use their Docker Hub credentials to avoid
+   rate limiting. How do we do that?
+
+1. How too reconfigure existing and future nodes to use the cache as a basic registry and upstream ACR as a backup
```
uvegla, line 201:
```suggestion
1. How to reconfigure existing and future nodes to use the cache as a basic registry and upstream ACR as a backup
```

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+  going through a formal process, though).
+- Automatic retention
+  - When enabled, manifests that don't have any associated tags (untagged manifests) and are not locked, will be
+    automatically deleted after the number of retention days specified. [Docs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-retention-policy).
+- Best-in-class support for the OCI standard (it's MS that is pushing it forward).
+- Reasonable pricing
+  - price per region €1.540/day (500 GB included), then €0.003/GB/day + traffic fee
+  - [Azure cost calculator](https://azure.microsoft.com/en-us/pricing/calculator/?service=container-registry)
+
+### Choosing caching solution
```
mcharriere, line 89: Hopefully it's not too late :)
have you considered Zot registry? It seems it fits our needs (https://zotregistry.io/v1.4.3/articles/mirroring/)

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+We checked for the single best registry and chose ACR. We did some evaluation before and you can read results here:
+
+- <https://github.com/giantswarm/roadmap/issues/2382>
+- <https://hackmd.io/gqF6PT53RcG_i4OhrmtWIw#Note-on-compatibility>
+
+Additionally, ACR offers th following:
+
+- High availability by definition.
+- Automatic geo-replication, including geo-replication between Europe and China (confirmed by Azure support, requires
```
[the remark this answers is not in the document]
piontec: To some extent, the feature is available, but you still pay for storage and traffic

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+        - there's no mention of how to deploy outside of AKS edge cluster, seems Azure IoT edge thing only
+- [docker's distribution/distribution](https://github.com/distribution/distribution)
+  - one instance can proxy only for a single upstream registry (but that's OK for us)
+  - tested, works with `containerd` for single upstream repo, works as well with `helm` charts
+  - no extra dependencies, can work with just local filesystem storage
+  - exposes reasonable prometheus metrics (transfer times, cache hit ratio)
+  - <https://github.com/distribution/distribution/blob/main/docs/configuration.md#proxy>
+  - <https://github.com/docker/docs/blob/main/content/registry/recipes/mirror.md>
+
+As a result, it seems we can use the `distribution` project from docker.
```
MarcelMue, line 113: I think there is more to consider here - `distribution` has no user management / SSO, no storage possibilities apart from local filesystem.

Additionally AFAIK deployment of `distribution` is essentially "figure it out yourself", there is no official guidance apart from "here are some scripts to run it on your local machine". There are some guides online, but it would be something that we need to take care of.

single-registry-architecture/README.md
silenteh: I think we have 2 case scenarios here:

- moving to a single registry architecture, which is the single source of truth for all GiantSwarm container images, and which is the part you have started the conversation about

- having a pull-through cache for:
    * GiantSwarm container images
    * Customer private registries

Currently I am focussing on the customer private registries, which will require a mutating admission controller.

The mutating admission controller should have the possibility to define the hostname of the container registry that needs to be replaced (actually multiple of them, so a `map[string]string`), so that it acts only on the customer private registries, to start with.

A potential second phase would be to add support for multiple upstream registries in the pull-through cache, so that we can benefit from it, also for the GiantSwarm images.

The idea would be to deploy this app on-demand on the workload clusters (again to start with)

I agree that of all the pull-through caches listed, the docker disribution one is the best, with the caveat that there will be the need to deploy multiple instances, one for each upstream registry

single-registry-architecture/README.md
silenteh: I don't disagree, but ideally there is proper ownership of the component then.  
Whichever team owns it, need to make sure it's running day and night.  

Because of the above, my vote goes for a managed solution.

single-registry-architecture/README.md
silenteh: `distribution` in this case is for caching purposes only.  
The user management should be done in the upstream registry and not at the cache level. So personally I see this as a plus.

Regarding the storing capabilities, considering this is a cache, I personally think it's fine. The only consequence is that only one replicas can run at a given point in time. Also from time to time, the mounted volume should be cleaned up, because I doubt the cache checks with upstream whether all the blobs are still present. So this operation is as simple as deleting the volume and creating a new one.

What I would be more interested in finding out is how it manages changes in the upstream manifest and whether it always checks the upstream manifest before serving the cached content.
This is important, for example, in case customers use the `latest` tag for development purposes and always override the `latest` image. The cache needs to know about it, otherwise the users might get unexpected results

single-registry-architecture/README.md
mcharriere: I mean zot for the caching/mirror solution.

single-registry-architecture/README.md
marians: Can we think of another cleanup strategy that's easy to implement with the ACR built-in features? One where dev builds can have tags, too. I'm asking for the case that we find we want tags on dev artifacts.

single-registry-architecture/README.md
piontec: In general, I agree, but I think we need the GS images caching solution from day 1, otherwise we risk high cost. Also, I really want to avoid webhooks based approach, as it's a potential failure point, and an ugly one :|

single-registry-architecture/README.md
piontec: AFAIR, for example in AWS, transfer within region is free (or very cheap). Inter-region and internet is where gets more expensive.

The cache solution I've tested is a read-only cache, you can't push to it (I tried to depict it in the diagram). As for HA, it is possible, but might require multiple storage copies (for filesystem option). I guess we can avoid that by using an S3 storage, that makes the Pods almost stateless, but will be probably slower and might cost more (although it doesn't need any storage scaling, that's a pro).

single-registry-architecture/README.md
piontec: Sure, that's another approach. But, remember all those failed builds or builds that take 30 mins instead of 5 mins, where the only reason for that is that upload to China is slow or fails? We can only pick our poison here...

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+empty repo and populate it using a separate instance of `retagger`. We won't start uploading build artifacts there
+before we figure out how to use the retention policy.
+
+High level migration plan:
+
+- bootstrap a new ACR registry for images, let's call it tentatively `gsoci`
+  - configure access, geo-replication and a retention policy for the repo
+- prepare and deploy `distribution` cache instance, configure it for the new repo
+  - ensure monitoring and alerting
+- switch our CI/CD pipelines to not tag dev build artifacts
```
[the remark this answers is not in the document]
piontec: I was thinking about it, separate registry for that has at least a few shortcomings:
- Big one: we have to configure all the runtimes to use this registry as well. Moreover, they need to know when to pull from the "release" registry and when from the "dev" one. So this probably means different registry domains. And CI/CD that needs to "know" which one to upload to. A lot of complication, IMHO. I would prefer just not to tag them and be done. Or - if we have to tag them - have a tagging schema that we can use to detect "dev" builds and write a piece of software, that runs and "untags" dev builds. But that's again a service we have to write to maintain and I want to use as few custom pieces here as possible
- Slightly bigger cost: there's another repo we have to pay the base fee for
- One more thing to keep in mind, monitor, manage... (yeah, mostly fire&forget, but still)

single-registry-architecture/README.md
piontec: I wasn't sure. I think we have round-robin, but if it always starts with the primary one, then we're good and don't need anything else

single-registry-architecture/README.md
marians: Traffic within an AWS region is 0.01 USD per GB.

single-registry-architecture/README.md
piontec: I think it doesn't - [`HelmRepository` supports OCI just as a type config option](https://fluxcd.io/flux/components/source/helmrepositories/#type) and we should already have this available, as we already support installing stuff from OCI registries. So, I expect changes done by Łukasz cover this already, but can you confirm @ljakimczuk ?

single-registry-architecture/README.md
```diff
@@ -0,0 +1,217 @@
...
+  - we want to use retention policies for helm charts as well
+- implement changes in the `app-operator`
+  - use OCI discovery instead of `index.yaml` from HTTP registries
+  - (optional) solve the problem of presenting only X (by default 5) most recent versions in the ACE CR
+  - introduce a concept of primary and fall-back registry
+- stop uploading charts into HTTP registries
+- reconfigure `app-operator` to use the OCI cache as the primary registry and upstream ACR as fallback
+- cleanup HTTP registries
```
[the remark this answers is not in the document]
piontec: Yes, they do. So, as part of the migration, we'll have to update registry URLs/references for apps we support.

single-registry-architecture/README.md
piontec: We can have a tagging schema that we can use to detect "dev" builds and then write a piece of software, that runs and "untags" dev builds. But that's again a service we have to write to maintain, and I want to use as few custom pieces we have to support as possible.

single-registry-architecture/README.md
piontec: No, I missed this one somehow. I'll catch up and try to check if and how we can use it for caching.

single-registry-architecture/README.md
piontec: OK, commenting one-by-one about `dsitribution`

1. Running is pretty easy, I already tested it in a local-storage mode, just works (TM)
2. Storage has options, it can be either local or S3 ([reference](https://github.com/docker/docs/blob/main/content/registry/deploying.md#customize-the-storage-back-end), not tested. S3 obviously makes it easy, but we have to be cross-provider, including on-prem, and here's where the dragons start. As such, I'm leaning towards filesystem storage, as it's easily supported everywhere - but has its own drawbacks (scaling size). Please note, that it's a read-only cache. And that it supports TTL for items, so it's not just super dumb save-and-forget solution.
3. It can run in HA mode, but then it either needs S3 storage or each Pod has its own copy of the cached data (so this increases the cost and miss-rate). [ref](https://github.com/docker/docs/blob/main/content/registry/deploying.md#run-the-registry-as-a-service)
4. `latest` tag is very good point, I'm not 100% sure, but I can easily test that. But I'm 90% certain, that when running in the debug mode, it was always doing HEAD upstream to check for SHA. A question is 'so what's when upstream is down and doens't respond'.

piontec: One more comment: please not, that even if we run the registry as a single Pod, it's not a SPoF, as we want to configure all the client services with explicit fail over to upstream registry. This obviously means higher cost and probably higher latency, but at least is not critical.

single-registry-architecture/README.md
uvegla: I feel my main pain point here was that this causes interconnection / dependency between MCs, making one more critical, even if the platform can fall back to upstream when cache is down.

The where do we host it issue is:
- Do we have / want to have a GS MC every region and provider our customers have?
- Would feel bad hosting in Customer A lets's say `us-west-2` and somehow expose it to Customer B `us-west-2`.

single-registry-architecture/README.md
uvegla: If we need such a software, maybe we can utilise OCI here to store that something is a dev build.

single-registry-architecture/README.md
uvegla: You are right! This is not fully implemented in this sense.

single-registry-architecture/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+More specifically, our aim is to:
+
+- Stop using Docker Hub by the end of 2023.
+- Simplify registries architecture.
+- Switch entirely to OCI based registries for both container images and helm charts.
+  - Flux developers have already declared that no new feature will come to Flux for users of HTTP-based repos, only for OCI. 
+  - Currently, we use a mix of HTTP and OCI repos, but chart discovery uses HTTP-based ones only.
+- If possible, switch to single OCI provider.
+- Increase availability and lower the network transfer cost by providing a pull-through cache service.
```
JosephSalisbury, line 54: i'd really challenge whether this needs to be included here - considering the timeline, i'd argue it's better to tackle the migration first, and then separately add the pull-through cache

single-registry-architecture/README.md
```diff
@@ -0,0 +1,216 @@
...
+- Automatically clean old dev release artifacts.
+- If possible, use the registry provider to replicate registries between Europe and China.
```
JosephSalisbury, line 56: And maybe we can also remove this for scope - aim is to get off DockerHub, we can tackle Aliyun later if needed

JosephSalisbury (commented):
I feel like the scope can be reduced to just migrating off Dockerhub, and then we can save on a lot of discussion in this RFC

single-registry-architecture/README.md
piontec: I added a note that we might do one or the other. To get started, we won't change anything for China anyway.

single-registry-architecture/README.md
piontec: I think we have (at least currently) 1 MC per each provider and region our customers use. We can use that to host the cache. Since the cache caches only "our" images, it should be us, who pays for that. It will still be well worth it compared to our savings on the upstream registry. As for the interconnection: I agree, that's why I want a soft-failure model: even if the cache fails, we can still use the highly performant upstream source, we just have to pay for it.

single-registry-architecture/README.md
piontec: Do you mean some meta-data associated with an image? Sure, that's a valid idea.

single-registry-architecture/README.md
piontec: Yes, I refined the scope, changes will be pushed in 1 min.

JosephSalisbury (approved):
yeah, i think this is pretty much good to go - we move from dockerhub to acr for images now, and can focus on specifics for charts / caches / china later

bam!

JosephSalisbury:
what are next steps? merge here, start working on an implementation plan for the dockerhub migration?

puja108 (approved):
LGTM overall, minor things might need adjustment over time, but we'll see when we get to them. We need to get phase 1 going soon, and that one sounds about right.

piontec:
> what are next steps? merge here, start working on an implementation plan for the dockerhub migration?

Yeap, turn that into tickets, start checking what needs to be reconfigured where.

single-registry-architecture/README.md
pipo02mix: >  who pays for that

If we running the cache in the MC account, customer will pay for it AFAIK

PR #78, opened 2023-10-10
Add decision process (alternative proposal)

AndiDog:
This turns out to be pretty much a rewrite of https://github.com/giantswarm/rfc/pull/70, taking out things that we might not need and trying to write down the process as intuitive and minimalistic one. Therefore, please consider this an alternatively-worded replacement of #70.

@JosephSalisbury I decided against merging into the other PR because of the vast changes in text. Think of _this_ PR as alternative suggestion, replacing #70.

Towards https://github.com/giantswarm/giantswarm/issues/24661


lyind (commented):
I like the simplified version better.

However, I can't tell if it works for all stakeholders or if there are any missing rules or details.

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
+---
+creation_date: 2023-10-10
+issues:
+  - https://github.com/giantswarm/giantswarm/issues/24661
+owners:
+  - https://github.com/orgs/giantswarm/teams/sig-architecture
```
JosephSalisbury, line 6: Why did you go with URL over just the team name?

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+To allow for a decision making process without removing the current free-form nature of RFCs, some structured information is proposed.
+
+The following changes will be introduced:
+
+- All RFCs will have a YAML header with structured information that can be parsed and used e.g. for rendering the RFCs in our [handbook](https://handbook.giantswarm.io/)
+- The YAML header must contain the following fields in this order:
+
+  - `creation_date: YYYY-MM-DD`
+  - `issues` containing links to GitHub issues. Can be null or an empty list. The issues can be the origin of the RFC or where the work for implementing the decision takes place.
```
JosephSalisbury, line 39: When would this be acceptable to be null?

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+- **Review phase begins.** The author posts a link to the PR to appropriate `#news-*` or other Slack channels, and adds stakeholders as reviewers on the PR. The reviewers use GitHub comments on the PR. They should not push directly to the PR branch unless agreed with the author(s).
+- **Any employee can review. The relevant SIG takes care that the RFC progresses and does not become stale.**
+
+  - For example, the SIG's regular sync meetings can be used to discuss open RFCs, but discussions should end up in GitHub PR comments.
+  - For engineering RFCs, typically the members of SIG Architecture would take responsibility.
+  - If discussions do not progress and the PR is open for quite a while, the author has the option to merge the PR containing `state: review`. This way, a new review can be done by proposing changes and `state: approved` in a _second PR_. Hopefully, we won't need this last resort hack. Let's see if this makes sense. If not, we can remove the state `review` and immediately go for `state: approved` ❓.
```
JosephSalisbury, line 102: i'd drop this, it feels somewhat like an exception?

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+To allow for a decision making process without removing the current free-form nature of RFCs, some structured information is proposed.
+
+The following changes will be introduced:
+
+- All RFCs will have a YAML header with structured information that can be parsed and used e.g. for rendering the RFCs in our [handbook](https://handbook.giantswarm.io/)
+- The YAML header must contain the following fields in this order:
+
+  - `creation_date: YYYY-MM-DD`
+  - `issues` containing links to GitHub issues. Can be null or an empty list. The issues can be the origin of the RFC or where the work for implementing the decision takes place.
+  - `owners` containing GitHub team URLs. This list must not be empty. Please enter only the team and/or SIG names of the author and owners, not the teams who should comment on the RFC (the _stakeholders_, for which we don't have a structured field). The `owners` field contains teams who have the expertise to check the RFC after a few months. It may later be used to automatically create GitHub issues "Please review if this RFC is still relevant" (see [Future ideas](#future-ideas)).
```
JosephSalisbury, line 40: i feel like stakeholders is somehow necessary? how can we make sure that relevant parties outside of owners have seen / accepted an RFC?

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Therefore, we will automatically render the merged RFCs into our [handbook](https://handbook.giantswarm.io/docs/) and provide at least a table as overview.
+
+## Decision making process
+
+- **Author opens pull request**
+
+  - Clone the [`rfc` repo](https://github.com/giantswarm/rfc)
+  - Create a new directory such as `nginx-ingress-app-migration`. It should be named like the others, i.e. without spaces. Mind that this name (so-called _slug_) is later used in the handbook URL of the rendered RFC and should not change.
+  - In that directory, create a `README.md` with the following template:
```
pipo02mix, line 79: We might create this as template file?

decision-process/README.md
pipo02mix, line 102: I would mention to elevate in the SIG or last case to Horizon

decision-process/README.md
AndiDog: Same with the `docs` repo (where the field is called `owner` – I changed that to plural _here_). It's also easier to verify whether it points to an actual GitHub group, as we desire here for future automation.

decision-process/README.md
AndiDog: I assume you are asking about getting a minimum number of approvals from relevant teams without forgetting important people – that must be represented as "reviewers" on the PR, since there's no reasonable way to sync `stakeholders: [ ... ]` and PR reviewers (too complex). The chance of forgetting a stakeholder in a YAML field is likely the same as forgetting to set a reviewer on the PR – and adding a team on the PR is just a few clicks and therefore much easier. If you later want to see who reviewed/approved/commented, we just look at the original PR. If there are complaints that a certain team wasn't involved, that's not really what a YAML field can heal, in my opinion, but it could be a mindset or laziness issue, for instance, missing knowledge of the author, or a plain human mistake. That's why I think we should have some very central place where RFCs get announced, and/or regularly discussed. The relevant SIG should take care of adding missing reviewers on the PR. ◀️ Should I add this clarification to the RFC? I agree about the concern and that we should solve it as part of this written process.

If you had concrete ideas about the YAML field, let me know. Maybe you had a good solution in mind?

decision-process/README.md
AndiDog: For personal stuff, I have something like a `create-rfc.sh` script that asks certain information (date, slug, title) and creates the file accordingly. I had a similar idea, but was wondering whether we can make this whole procedure as un-technical as possible so that it's open to anyone. Is there a way to solve this by clicking around GitHub's UI, not requiring a local checkout of the repo, and using the WYSIWYG Markdown editor of GitHub? A template file sounds good if GitHub allows copying that, or has some way to get started writing an RFC without cloning the repo. But maybe I'm mistaken and git is not a hurdle – WDYT?

decision-process/README.md
AndiDog: Ok, will do

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+review --> declined
+
+approved --> obsolete
+declined --> obsolete
+```
+
+## Future ideas
+
+These ideas could be made mandatory and added in _this process_ later (by means of amending _this RFC_):
+
```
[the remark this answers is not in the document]
AndiDog: Okay, I will add this to the written process

decision-process/README.md
JosephSalisbury: works for me, sure

decision-process/README.md
JosephSalisbury: i think i'm okay with this - my thinking is that i want all (important) rfcs to have some tracking information, but maybe that's too heavyweight

let's go with this for now

decision-process/README.md
JosephSalisbury: my thinking was having a field similar to owners, and then do some CI magic to ensure that at least one person out of the listed teams has approved

my hunch is that having it in the text means that people will consider it more, than if it's in github, but i'm not too sure on that

happy to go either way, i think

JosephSalisbury:
i'll approve once we've got the last comments sorted out, this is a good direction

decision-process/README.md
AndiDog: I'm leaving the template here in the "RFC of RFCs" for now. As long as people use a correct header and h1 heading, it doesn't matter if the file got copy-pasted, came from a template or whatever. People will shout if they think the process is complicated or needs a simple script to create an RFC. We can improve on that when we see it necessary.

decision-process/README.md
AndiDog: Replace the existing bullet point with your suggestion (@JosephSalisbury please comment, since you recommended removing it)

decision-process/README.md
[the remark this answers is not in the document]
AndiDog: See "Author merges the PR" bullet point where this was already mentioned. I added `#news-*` in the sentence.

decision-process/README.md
AndiDog: I'd recommend we leave this out for now, and if it doesn't work, we can improve the flow by adding the required field

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+      - https://github.com/orgs/giantswarm/teams/team-FILL_IN_YOUR_TEAM # or https://github.com/orgs/giantswarm/teams/sig-FILL_IN_YOUR_SIG
+    state: review
+    ---
+
+    # This is the RFC title
+    ```
+
+  - Replace the placeholders
+  - Push your branch and open a pull request
+- **Author asks their own team/SIG for a short review.** They should check if the content and grammar make sense. There should be no discussion yet about the decision since not all stakeholders are invited at this time.
```
AndiDog, line 96: @anvddriesch commented that "no discussion" may not be feasible. Maybe there's a better wording. And I think this item is optional.

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+- **Review phase begins.** The author posts a link to the PR to appropriate `#news-*` or other Slack channels, and adds stakeholders as reviewers on the PR. The reviewers use GitHub comments on the PR. They should not push directly to the PR branch unless agreed with the author(s).
+- **Any employee can review. The relevant SIG takes care that the RFC progresses and does not become stale.**
```
AndiDog, line 98: @stone-z commented that the RFC author should primarily drive the effort, in their own interest.

@stone-z / @puja108 reported previously using a "review until" date. We'll have to see how this works out culturally. Please comment if you want such a concept in here for the "hard" cases. We discussed in SIG Architecture that RFCs which are important strategically or for the roadmap will usually progress without problems.

decision-process/README.md
```diff
@@ -0,0 +1,134 @@
...
+    owners:
+      - https://github.com/orgs/giantswarm/teams/team-FILL_IN_YOUR_TEAM # or https://github.com/orgs/giantswarm/teams/sig-FILL_IN_YOUR_SIG
+    state: review
+    ---
+
+    # This is the RFC title
+    ```
+
+  - Replace the placeholders
+  - Push your branch and open a pull request
```
AndiDog, line 95: @nprokopic commented that the PR should be added to the GitHub team/SIG board

AndiDog (commented):
Added commentary from SIG Architecture meeting

decision-process/README.md
nprokopic: My first gut feeling was like Joe's, that we should also have stakeholders, but then I thought that I prefer this RFC being as is right now, simpler.

Then after we we see how this simpler version is working in practice, we can have another RFC which makes stakeholders mandatory, so we can also have e.g. like Joe said CI checks if there is an approval from every stakeholder.

decision-process/README.md
AndiDog: Added as action item

decision-process/README.md
AndiDog: Reworded

decision-process/README.md
AndiDog: Mentioned author as primary driver

JosephSalisbury (approved):
yeah, let's go - @AndiDog can we get some ci magic to check headers? what are next steps for you?

AndiDog:
> yeah, let's go - @AndiDog can we get some ci magic to check headers? what are next steps for you?

I had already prepared rendering of RFCs into a table within the handbook. That helps getting an overview.

And changing the `rfc` repo to check YAML headers shouldn't be that hard.

But let's first get this decision done. There are only 2 approvals by now. When should we go ahead?

JosephSalisbury:
@AndiDog i think we can go ahead now

nprokopic (approved):
Great improvements to the RFC process 🎉 Thanks @AndiDog and @JosephSalisbury!

AndiDog:
Made the last call in SIG Architecture. Merging this week and going ahead to implementation if there are no more objections or ideas.

fiunchinho (approved):

MarcelMue (approved):

mcharriere (approved):

PR #80, opened 2023-11-15
Adapt existing RFCs to new structure, add script and CI pipeline to check structure

AndiDog:
As required by our new [decision process](https://github.com/giantswarm/rfc/tree/main/decision-process).

On top of the new check script, which also outputs the RFCs in a structured way, I'll add RFC rendering to the `handbook` repo.

I filled the `creation_date` field of each RFC from its oldest commit, and marked all RFCs as `state: approved`. We can clean up later and introduce missing fields for old RFCs. Only new RFCs require all fields while existing RFCs can remain with the minimal set `creation_date`/`state` for now.


pipo02mix (approved):
LGTM solid work!

PR #81, opened 2023-12-07
Trigger RFC rendering workflow in `handbook` repo, fix dead links

AndiDog:
Previously tested on a dev branch.

Since the handbook is now auto-released, I changed the RFC process description to reflect that the rendered list of RFCs should be available soon after merging something in _this_ repo.


pipo02mix (approved):
👌

PR #85, opened 2024-02-08
New RFC: multi-tenancy

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/29750

### DIFF
--- a/multi-tenancy/README.md
+++ b/multi-tenancy/README.md
@@ -1,0 +1,27 @@
+---
+creation_date: 2024-02-08
+state: review
+issues:
+- https://github.com/giantswarm/roadmap/issues/2771
+summary: How GiantSwarm manages multi-tenancy to isolate data by tenants (i.e. metrics, logs, traces).
+---
+
+# Multi-tenancy proposal
+
+## Introduction
+
+Since we have a logging infrastructure, we are thinking about the multi-tenancy.  
+The observability stack with metrics, logs, traces is treating a lot of data.  
+According the role of people, some data must be accessible or not.  
+We would like to propose our customers the option of defining their tenants so that they can isolate the data as they want.
+
+Our product currently allows anyone with access to Grafana to request all the data: metrics and logs.
+Those datas are written with a tenant id that corresponds to the cluster id.
+
+Our idea of multi-tenancy is to be able to isolate data by tenant. A tenant can be anything: a namespace, a cluster id, a group of people, a feature, etc.
+
+We need to discuss how we handle multi-tenancy when accessing data (read path) and when sending it to object storage (write path).
+
+## Read path
+
+## Write path


multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+According the role of people, some data must be accessible or not.
+We would like to propose our customers the option of defining their tenants so that they can isolate the data as they want.
+
+Our product currently allows anyone with access to Grafana to request all the data: metrics and logs.
+Those datas are written with a tenant id that corresponds to the cluster id.
+
+Our idea of multi-tenancy is to be able to isolate data by tenant. A tenant can be anything: a namespace, a cluster id, a group of people, a feature, etc.
+
+We need to discuss how we handle multi-tenancy when accessing data (read path) and when sending it to object storage (write path).
+
```
hervenicol, line 26: We may add something along the lines of:
```
In this document we will talk mostly about the logs (with `Loki`), but we expect to share the same logic for all components.
```

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+    "giantswarm-ad:Developers"
+  ],
+  "name": "People People"
+}
+```
+
+The component `loki-multi-tenant-proxy` should be renamed `grafana-multi-tenant-proxy` to be more generic because it will handle multi-tenancy for logs, metrics, traces, etc.
+That component should have a configuration in which the mapping between groups of people and tenants is defined.
+
+Below a proposal such of configuration:
```
hervenicol, line 67:
```suggestion
Below a proposal of such a configuration:
```

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+  - wcdev-name
+- name: Ops
+  tenant_ids:
+  - wcdev-name
+  - wcprod-name
+...
+```
+
+That configuration should be dynamically created and updated regardless of the data source.
+We are thinking of having an operator to manage that mapping. The operator will be able to reconcile any kind of data source and generate the configuration expected by `grafana-multi-tenant-proxy`.
```
hervenicol, line 87:
```suggestion
We are thinking of having an operator to manage that mapping. The operator would be able to reconcile any kind of data source and generate the configuration expected by `grafana-multi-tenant-proxy`.
```

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+[Teleport JWT issue](https://github.com/giantswarm/giantswarm/issues/29719)
+
+We are authentified on Grafana through a JWT token provided by teleport.
+But that token is not forwarded to the datasource.
+So we are not able to authenticate the user into the multi-tenant proxy.
+We are considering a number of options that are not ideal at the moment:
+
+- having double datasource: one for customer with OAuth authentication and another dedicated to GiantSwarm people authenticated throw teleport.
+- having double grafana (same idea than above but for the all grafana instance, not just the datasource).
+- other ideas are welcome!
```
hervenicol, line 107: If the cookie idea is still an option:
```suggestion
- share an authentication secret via cookies
- other ideas are welcome!
```

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+#### Operator
+
+`logging-operator` is an existing component responsible to create a secret for the multi-tenant-proxy. That secret contains a mapping between credentials and tenant (now, it's equivalent to the cluster id).
```
hervenicol, line 111:
```suggestion
`logging-operator` is an existing component responsible for creating a secret for the multi-tenant-proxy. That secret contains a mapping between credentials and tenant (it's currently equivalent to the cluster id).
```

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+Do we add a feature in that operator to handle the multi-tenant-proxy configuration we described above ?
+In that case, we should rename it: any suggestion ?
+If we don't reuse `logging-operator`, we will need to implement a new one: any suggestion for its name ?
```
hervenicol, line 114: This list of question in a single paragraph is a bit dense.
Maybe this should be a list?

multi-tenancy/README.md
```diff
@@ -0,0 +1,134 @@
...
+creation_date: 2024-02-08
+issues:
+- https://github.com/giantswarm/roadmap/issues/2771
+owners:
+- https://github.com/orgs/giantswarm/teams/team-atlas
+state: review
+summary: How GiantSwarm manages multi-tenancy to isolate data by tenants (i.e. metrics, logs, traces).
+---
+
+# Multi-tenancy proposal
```
mcharriere, line 11: In our public docs we already use `multi-tenancy` to name the features and structure around Organizations. Since this is not about that, could you consider changing the title to reflect that they are not the same topic? 

```suggestion
# Multi-tenancy on monitoring
```

same goes for the folder name and other generic references to "multi-tenancy". wdyt?

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
+---
+creation_date: 2024-02-08
+issues:
+- https://github.com/giantswarm/roadmap/issues/2771
+owners:
+- https://github.com/orgs/giantswarm/teams/team-atlas
+state: review
+summary: How GiantSwarm manages multi-tenancy to isolate observability data by tenants (i.e. metrics, logs, traces).
```
QuentinBisson, line 8:
```suggestion
summary: How Giant Swarm manages multi-tenancy to isolate observability data by tenants (i.e. metrics, logs, traces).
```

QuentinBisson, line 8: nitpick: I think it's better to use per tenant than by tenant

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+---
+
+# Multi-tenancy proposal on observability
```
QuentinBisson, line 11:
```suggestion
# Multi-tenancy for observability
```

I would remove the proposal part

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Introduction
+
+Since we have a logging infrastructure, we are thinking about the multi-tenancy.
```
QuentinBisson, line 15: I'm not sure we need this sentence tbh

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+The observability stack with metrics, logs, traces is managing a lot of data.
```
QuentinBisson, line 16:
```suggestion
Giant Swarm managed components produce a lot of observability data (metrics, logs, traces and so on) For instance, we have reached our ingestion limits in Prometheus more than a year ago.

On top of this, as part of the developer platform, we want to be able to ingest all sorts of customer observability data.

To that end, we now need to move towards multi-tenancy in our observability stack.

Thankfully, our logging solution Loki supports multi-tenancy based on a http header (`X-Org-ID`) and we already had to implement things .. @marieroque let's talk about the read user here

To be able to support multi-tenancy in our monitoring stack, we are currently working on moving to mimir on CAPI as mimir supports the same multi-tenancy mechanism as Loki (@marieroque maybe link the epic)
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+According to the role of people, some data must be accessible or not.
+We would like to propose our customers the option of defining their tenants so that they can isolate the data as they want.
+
+Our product currently allows anyone with access to Grafana to request all the data: metrics and logs.
```
QuentinBisson, line 20:
```suggestion
Presently, our product allows anyone with access to Grafana to request all the data (metrics and logs) for all tenants. This is part of the reason why only a subset of our customers (only the platform teams) have access to our managed grafana and it also why our shared installation grafana is not accessible to customers.
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+Those datas are written with a tenant id that corresponds to the cluster id.
+
+Our idea of multi-tenancy is to be able to isolate data by tenant. A tenant can be anything: a namespace, a cluster id, a group of people, a feature, etc.
+
+In this document we will talk mostly about the logs (with `Loki`), but we expect to share the same logic for all components.
```
QuentinBisson, line 25:
```suggestion
In this document we will talk mostly about the logs (with `Loki`), but we expect to share the same logic for all our observability stack components.
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+We need to discuss how we handle multi-tenancy when accessing data (read path) and when sending it to object storage (write path).
```
QuentinBisson, line 27: Because the read and write path for the multi-tenancy is technically different, we decided to split it into different sections.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+- `write path`: path used by components to store observability data
+
+## Introduction
+
+Giant Swarm managed components produce a lot of observability data (metrics, logs, traces and so on).
+For instance, we have reached our ingestion limits in Prometheus more than a year ago.
+On top of this, as part of the developer platform, we want to be able to ingest all sorts of customer observability data and provide data isolation between tenants.
+
+To that end, we now need to move towards multi-tenancy in our observability stack.
+Thankfully, our logging solution Loki supports multi-tenancy based on a http header (`X-Org-ID`) and we already had to implement things.
```
JosephSalisbury, line 27:
```suggestion
Thankfully, our logging solution Loki supports multi-tenancy based on a http header (`X-Org-ID`) and we already had to implement things.
```

what does 'implement things' here mean?

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+To be able to support multi-tenancy in our monitoring stack, we are currently working on moving to mimir on CAPI as mimir supports the same multi-tenancy mechanism as Loki, see our current status on the [epic](https://github.com/giantswarm/roadmap/issues/3039).
+
+According to the role of people, some data must be accessible or not.
+We would like to propose our customers the option of defining their tenants so that they can isolate the data as they want.
+
+Presently, our product allows anyone with access to Grafana to request all the data (metrics and logs) for all tenants. This is part of the reason why only a subset of our customers (only the platform teams) have access to our managed grafana and it also why our shared installation grafana is not accessible to customers.
```
JosephSalisbury, line 34:
```suggestion
Presently, our product allows anyone with access to Grafana to request all the data (metrics and logs) for all tenants. This is part of the reason why only a subset of our customers (only the platform teams) have access to our managed grafana and is also why our shared installation grafana is not accessible to customers.
```

grammar nitpick

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+<img src="./assets/future-multi-tenancy.png" width="300" alt="Multi-tenancy proposal on the read path">
+
+### Opened questions
+
+#### Teleport issue
+
+We are currently facing an issue with `teleport` authentication:
+[Teleport JWT issue](https://github.com/giantswarm/giantswarm/issues/29719)
+
+We are authentified on Grafana through a JWT token provided by teleport.
```
JosephSalisbury, line 118:
```suggestion
We are authenticated on Grafana through a JWT token provided by teleport.
```
grammar nitpick

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,153 @@
...
+## Introduction
+
+Giant Swarm managed components produce a lot of observability data (metrics, logs, traces and so on).
+For instance, we have reached our ingestion limits in Prometheus more than a year ago.
+On top of this, as part of the developer platform, we want to be able to ingest all sorts of customer observability data and provide data isolation between tenants.
+
+To that end, we now need to move towards multi-tenancy in our observability stack.
+Thankfully, our logging solution Loki supports multi-tenancy based on a http header (`X-Org-ID`) and we already had to implement things.
+The graph below shows our current implementation of the multi-tenancy on logs:
+<img src="./assets/scope-orgid.png" width="500" alt="Loki multi-tenancy and X-Scope-OrgID header">
```
marians, line 29: For better readability, to make the image a separate paragraph

```suggestion
The graph below shows our current implementation of the multi-tenancy on logs:

<img src="./assets/scope-orgid.png" width="500" alt="Loki multi-tenancy and X-Scope-OrgID header">
```

puja108:
I feel we have a bit of confusion of concepts in here, or at least reading the RFC I'm unsure if the concepts are cleanly separated and will enable us to serve current known and future use cases. We need to have a clear data model and then see that our intended UX model together with that data model can enable all intended use cases for end users.

For the product level, I don't know if you have an epic listing use cases, yet, but if not we should have one (cc @Rotfuks @QuentinBisson). Some abstracted requirements, summarized to a technical data level to be more useful in this discussion:
1. There's users that should have access to all o11y data
2. There's users that should have access to certain WCs' data
3. There's users that should have access to certain NS' data
4. There's users that should have access to certain workloads' data (I'm unsure about this vs 3, as it might be enough to limit to ns level, but in a developer platform, developers don't care about namespaces but their workloads, so this has at minimum UX implications)
5. All of the above can be present at the same time

From what I can gather on the architectural level we need to solve the following:
- There should be three levels: 1. OIDC user/group identiy; 2. mapping of OIDC to a tenant 3. mapping of data to tenants
- There seems to be a conflation of 2 and 3 in what I'm seeing proposed in this RFC
- Specifically, I feel we are mixing data and tenant concepts, a WC or MC or NS should not be a tenant, they are descriptions of data that should be mapped to tenants
- Tenant to OIDC mapping (2) needs to change independent of changing data to tenant mapping (3)
- For 3 I am unclear how that works and if data can get mapped to several tenants? There might be technical limitations in the way the write path looks that I'm not aware of
- Additionally we have existing concepts like the MC orgs and MC as well as WC RBAC, but how those can be integrated can be solved separately as it could be additional automated mapping through the same interfaces that we need to define for manual mapping.

I suggest we have a live design session with the team and some people from Horizon to clarify this and then refine the RFC (or do an architecture preview like we plan with backstage) to get input from the other groups. WDYT?

QuentinBisson:
@puja108 I sadly did not have time to read the RFC yet but our original intention is to have the 2 levels done separately in what we call read and write path

Level 2 (mapping of OIDC to a tenant) would be done by the grafana-multi-tenant-proxy and some configurations from customers (could be a project CR, a tenant CR, a configmap, we do not know and it can change later for sure).

Level 3 (mapping of data to tenants) would be the write path per se which is not written yet.
Our idea is that customers would provide the tenant (could be via some CRs or via a label in promtail config and that would be the tenant). For now this field will be quite open and maybe it will always stay that way, we will have to see so teams could decide to set a tenant per application, per namespace, per cluster and that would be up to them. We might have to veto some things if encounter some issues but that's the gist of our idea.

So if we take concrete examples, teams could send logs from namespace `my-team-namespace` with tenant my-team-namespace and we would also send kube-system logs with tenant `giantswarm`. On the read path we can give them access to read the 2 of them or only 1.

If they want to they can also send logs from some pods in kube-system with their own tenants as well. This is technically feasible, the only downside is that the data will be stored twice.

We can have a call with horizon for sure but hopefully this should clear things up a bit

puja108:
Ok, maybe I need to see the write path to understand the whole thing, and maybe a data model diagram would also help.

Also, am I understanding right, tenants in this case have separated data stores, so having the same data for several tenants would mean we duplicate data?

In that case we actually have a 1:1 relationship of data to tenant (at least if we don't want too much duplication of data) and could only actually configure things on the OIDC to tenant access side. If that is the case we should consider some type of auto-mapping of data to grafana tenant, and limit customer condigurability to the access side. As long as our mapping is fine-grained enough to fullfil customer use cases we would not need to increase the complexity of the UX by exposing 2 configuration interfaces that in the end would need tight control to not get misconfiguration into the stack.

QuentinBisson:
Yes having the same logs for multiple tenants would duplicate data. This is not that much of an issue as we use object storage but it should not be the default case :)

I agree tightening it makes sense but we cannot really work on the auto-mapping unless we have some tooling on the MC side to take care of that and understanding each customer's needs will take a long time. Now having this flexible write path will allow us to start ingesting customer logs and we can figure out a proper ux later on. The main issue we have is that we cannot really use CRs to achieve that goal yet because they do not exist upstream yet

puja108:
I meant rather deciding on a mapped write path and fixing data/tenant. Like we could just say it is fully namespace based for now so that in the you get the data a la `cluster_name-namespace_name` (maybe we need to build or define sth for that, but it can be non-user facing) and then we only worry about the interface where the admin gives access to certain data to certain users and model use cases into that. Like when a customer needs to give access to all namespaces of a cluster or all namespaces that belong to a certain team we can do the mapping with an operator maybe. Does that make sense?

And if we see that along the way we need more fine-grained than namespace-based, we should be able to change it later and map programmatically existing mappings.

One question in this scenario would be, if there's any other limitations to the tenant model, like can there be dashboards that show data from several tenants (as long as the user has that access)?

QuentinBisson:
From some of our initial customer meetings, we cannot map 1 tenant = 1 namespace as we know some customers want to move logs from outside our clusters to loki already so we will need so flexibility from the start. Also we know that some customers want to start ingesting logs step by step to start with (as in moving teams one by one) so we think it is easier to onboard customers by being a bit less restrictive in the beginning and we can enforce patterns later when we know a bit more about their usage but we can discuss that decision in a call with @giantswarm/team-horizon 

There is no reason why we cannot access data from multiple tenants in one dashboard as the data is limited based on all the tenants a user has.

QuentinBisson:
@puja108 this is bit of a rewrite of the existing content as I can understand some of the confusions:
https://github.com/giantswarm/rfc/pull/91

I will start to write about the write path next week

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
+---
+creation_date: 2024-02-08
+issues:
+- https://github.com/giantswarm/roadmap/issues/2771
+owners:
+- https://github.com/orgs/giantswarm/teams/team-atlas
+state: review
+summary: How Giant Swarm manages multi-tenancy to isolate observability data per tenants (i.e. metrics, logs, traces).
```
AndiDog, line 8: I stumbled over _this PR_ from `#noise-rfc` and because of the summary. Since I'm currently preparing some best practices for technical writing, I have some nits regarding technical writing quality 😉. I know the RFC is still being drafted, but I wanted to leave these comments already.

---

This is not a summary, but the title of the RFC. Please put here what the RFC proposes, i.e. actions and changes.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+- `tenant`: a group of users who share a common access with specific privileges to observability data
+- `observability data`: all data related to metrics, logs, traces, profiles
+- `read path`: path used by users to access observability data
+- `write path`: path used by components to store observability data
+
+## Introduction
+
+Giant Swarm managed components produce a lot of observability data (metrics, logs, traces and so on).
+For instance, we have reached our ingestion limits in Prometheus more than a year ago.
```
AndiDog, line 23: Please use absolute terms ("a year ago" is undefined if somebody reads this later)

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+- https://github.com/orgs/giantswarm/teams/team-atlas
+state: review
+summary: How Giant Swarm manages multi-tenancy to isolate observability data per tenants (i.e. metrics, logs, traces).
+---
+
+# Multi-tenancy for observability
+
+## Glossary
+
+- `tenant`: a group of users who share a common access with specific privileges to observability data
```
AndiDog, line 15: Which users? Of the customer, of Giant Swarm? Maybe provide examples.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,154 @@
...
+- `observability data`: all data related to metrics, logs, traces, profiles
```
AndiDog, line 16: "data related to metrics [...]", or rather "metrics [...]" directly?

puja108:
> From some of our initial customer meetings, we cannot map 1 tenant = 1 namespace as we know some customers want to move logs from outside our clusters to loki already so we will need so flexibility from the start. Also we know that some customers want to start ingesting logs step by step to start with (as in moving teams one by one) so we think it is easier to onboard customers by being a bit less restrictive in the beginning and we can enforce patterns later when we know a bit more about their usage.

Wouldn't logs from outside anyway be handled outside the cluster in terms of tenant tagging?

Also, even with namespace based tenats we could enable namespace by namespace.

I do agree though that we need to work with customers a bit more to understand if there's needs we do not anticipate. I just fear that if we leave everything open to the customer and it's not straight forward to understand and configure the customer won't do it and we'll not see the usage we'd like to. We need to make sure this is easy to set up, and well explained with examples and maybe the 1 tenant + 1 namespace could be an example that we show and that we can then also use for ourselves.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+This is the approach we are currently using for technical reasons. We do not think this would match with customers workload as they mostly have a lot of customers sharing clusters.
+
+##### Namespace == Tenant
+
+This approach makes it really easy for us and customers to configure tenants and it would definitely be easy to onboard teams (e.g. adding a label on the namespace would make the agents collect logs from the containers in the namespace).
+
+But this approach brings a few questions:
+1. What tenancy should common internal apps use (e.g. prometheus-operator, api-server, cilium logs are accessed by multiple internal teams)? Do we need to ensure customers are putting those apps into a namespace accessible by all?
+2. What about teams that share namespaces?
```
puja108, line 200: From what I know from customers, we rarely have that outside of kube-system

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+#### Approach 2: Extract tenant information from a label/PodLog
+
+Existing logging agents support setting the tenant to send to loki using a custom log label.
+We can configure our logging agents to get the logs of specific namespaces/pods if they have a label like `giantswarm.io/tenant: my-tenant` or using a PodLog CR (if applicable).
+
+Pros:
+- Quite easy to configure for logs
+
+Cons:
+- Tenant configuration will need to be configured in multiple places (MC for the read path and WC for the write path)
```
puja108, line 241: Might not be a big problem as read path is access management and MC is a good place for that and write path workload related so deploying together with workload makes sense. The two lifecycles are separated.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+#### Approach 4: We enforce 1 namespace == 1 tenant with flexibility by app and possibility of opting out
+
+It is quite easy to configure 1 tenant per namespace in our current setup and we could use a giantswarm.io/logging: false on specific namespace or containers to not get the logs.
+
+Pros:
+- Customers would only have to fill out the read path configurations (unless they want to opt-out a lot)
+- With this approach, we could still have a `giantswarm.io/tenant: my-tenant` label on some apps that should have a specific tenant attached to them.
+
+Cons:
+- Our existing customers want to start slow (team by team) and opting-out for all namespaces from the start will be a hassle. An idea to avoid this would be to be able to opt-out of the multi-tenancy features with a cluster label (`giantswarm.io/tenancy: false`).
```
puja108, line 263: We can find that out with customers. AFAIK a lot of customers have managed namespaces that are controlled by the Platform team and where they should be able to add a default opt-out label on creation as they also add other things.

puja108 (approved):
Thanks, I think I understand much better now (also thanks to your explanations on other RFCs and previous comments). As I kinda suggested it, I'm in favour of approach 4 and I like that there's flexibility on top. Only need to find out a bit more about opting out and how feasible that is.

Main next step would be the read path mapping config we'd want to expose to customers. My main concern there is that we most probably would have a 1:1 there between out o11y permissions to namespaces and WC RBAC permissions to namespaces. Having a config interface (like a CR) makes sense, but we might be able to somehow generate it to reflect WC RBAC. We could sync with Big Mac here to see what we currently offer here in terms of WC permission management. IIRC there's an app we offer, but most customers might be using their own config.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+On the write path, we currently use promtail to collect container and machine logs and grafana agent to collect kubernetes events.
+
+We currently need those 2 tools because:
+1. Promtail cannot get kubernetes events
+2. Grafana Agent is going through quite a big rewrite in it's configuration and all features we needed were not supported when we deployed it
+
+Here is what the flow looks like:
+1. The logging-operator creates a `password` for the cluster's logging agents to be able to send logs to Loki and configures them (actual configuration of what to scrape and so on) via an extra-config.
+2. The logging-operator adds the new cluster_id/password pair to the proxy configuration
+3. The logging agents collect and send logs to the loki-multi-tenant-proxy which validates the basic authentication sent in the http header and transform it into a tenant information (X-Scope_OrgID)
```
pipo02mix, line 181: What in a future case where customer can define multiple tenants for write path, is that possible? does it require multiple instances of promtail and/or grafana agent?

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+- having double datasource: one for customer with OAuth authentication and another dedicated to GiantSwarm people authenticated throw teleport.
+- having double grafana (same idea than above but for the all grafana instance, not just the datasource).
+- share an authentication secret via cookies.
+- other ideas are welcome!
+
+#### Mapping user's groups to tenants
+
+What interface could we offer our customers to enable them to define the mapping between their groups and their tenants:
+
+- Custom resources like TenantCR, GroupCR
```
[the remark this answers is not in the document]
puja108: It is related, but I can see us pulling the concept apart so we can "translate" a project CR to specific custom CRs like a o11ytenant CR or sth. Because Project CR should be depicting project/namespace/team and other info, but here for example we only map OIDC group to o11ytenant, so this is more like RBAC binding.

multi-tenancy-observability/README.md
puja108: From what we discussed before IIRC we can have multiple tenants , but it would mean some data duplication. But @QuentinBisson can explain it better

multi-tenancy-observability/README.md
QuentinBisson: It's also the case for the monitoring namespace and it would probably the same tenants that can access the kube-system namespace

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+##### Cluster == Tenant
+
+This is the approach we are currently using for technical reasons. We do not think this would match with customers workload as they mostly have a lot of customers sharing clusters.
+
+##### Namespace == Tenant
+
+This approach makes it really easy for us and customers to configure tenants and it would definitely be easy to onboard teams (e.g. adding a label on the namespace would make the agents collect logs from the containers in the namespace).
+
+But this approach brings a few questions:
+1. What tenancy should common internal apps use (e.g. prometheus-operator, api-server, cilium logs are accessed by multiple internal teams)? Do we need to ensure customers are putting those apps into a namespace accessible by all?
```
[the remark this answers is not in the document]
QuentinBisson: For my discussion with customers, prometheus-operator was moved to another namespace because teams did not have access to kube-system logs :D

multi-tenancy-observability/README.md
QuentinBisson: Yes exactly, so the tenants can be set in promtail as a label via what they called a pipeline stage so if they want they can set multiple tenant, but that means the data will be stored and queries twice so we would advise against it :)

This is why the idea of setting a tenant per namespace makes sense :)

multi-tenancy-observability/README.md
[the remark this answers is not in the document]
QuentinBisson: So, my main questions in this case is more for our managed apps because customers can install them in namespaces they want and it could also be running with their own workload like it is the case with prometheus-operator. Now we technically have access to all the logs because we manage the clusters but in loki, we could theoretically reduce our access to only kube-system, giantswarm namespaces and managed apps logs only.

Do we want to go in that direction or should Giant Swarm have access to all logs by default?

multi-tenancy-observability/README.md
QuentinBisson:
```suggestion
- `observability data`: metrics, logs, traces, profiles and so on
```

multi-tenancy-observability/README.md
QuentinBisson:
```suggestion
- `tenant`: a group of individuals (from our customers or Giant Swarm alike) who share a common access with specific privileges to observability data
```

multi-tenancy-observability/README.md
QuentinBisson: Could you maybe give me examples so I could improve?

multi-tenancy-observability/README.md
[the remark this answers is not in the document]
QuentinBisson: In the beggining, we can have this maintained by us via giant swarm config until we think about the UX side to not block this RFC, but I'm not sure where (should we create a WG around multi-tenancy)?

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+2. What about teams that share namespaces?
+
+##### App == Tenant
+
+This approach is especially flexible for customers but we do not think this approach makes sense because the mapping configuration would be impossible to maintain and we would probably face technical limits pretty fast
```
stone-z, line 204: IMO this is the natural end state.

Two valid requests:
- "The team that owns namespace X can view logs from any application in that namespace"
- "The monitoring team can view prometheus logs from any namespace in any cluster where they run prometheus"

Roughly app-level granularity is inevitable in my mind. It also works well with other potential features (described in the general comment).

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+We can configure our logging agents to get the logs of specific namespaces/pods if they have a label like `giantswarm.io/tenant: my-tenant` or using a PodLog CR (if applicable).
+
+Pros:
+- Quite easy to configure for logs
+
+Cons:
+- Tenant configuration will need to be configured in multiple places (MC for the read path and WC for the write path)
+- Label approach would most likely not work for metrics OOTB (this would require customer to add new labels like metrics port, path and so on).
+
+#### Approach 3: Configure promtail based on the MC tenancy CR
```
stone-z, line 244: Unless the community provides a better option, some Giant Swarm abstraction is the likely end state for this use case. BUT, whether it is WC or MC level depends on architectural decisions about the Project CR which haven't been made yet.

Assuming a GitOps setup, consider the possible relationships between a GS Project CR and Flux. Does the Project CR live inside an application repository? If so, it would be a WC level resource. If the Project CR is managed separately from the application, it could be an MC-level resource. Those have tradeoffs and to my knowledge they haven't been discussed yet. 

So this mechanism could be WC-level (or could have complementary abstractions, like a label-based version of the same Project CR fields).

stone-z (commented):
The structure presented here overall makes sense to me and I agree with Atlas's current intent.

I would add:
1. App-level granularity seems inevitable to me. See my comment below.
2. To achieve app-level granularity, the theoretically _best_ approach IMO would be to use workload identities in some form. Logging as a platform feature will ship long before identity as a platform feature, but if logging is designed with that future in mind, it would be very powerful.
3. Shield has very similar use cases for security data. We are also building toward an identity-based model, but are currently working with (essentially) app-level granularity.
4. For the initial implementation, +1 on a single operator. Individual controllers can be spun out on their own if future needs demand it (Shield is doing the same for Policy API controllers).

Question/opinion - this seems like an enterprise feature of the underlying tools -- do you know of or anticipate any artificial limits from that? e.g. is there a limit to the number of tenant that would actually make per-app tenancy impossible?

multi-tenancy-observability/README.md
QuentinBisson: So the issue I have is that in my mind, a project CR would need need to be in the MC because a project can be cross clusters especially if customers have a cluster per stage as this is often the case.

multi-tenancy-observability/README.md
QuentinBisson: It could be but then I do not think it works for all things.

Usually, customers run applications for one team in one namespace with a clear boundary. Sure, prometheus might be a bit different but that's really dependant on the customers organizations.

What I do not like in this is that if you want all the logs of all apps in one namespace, then loki will actually run at least number of apps queries. Sure they're paralellized and so on but that does look a bit too much to me for most use cases.

QuentinBisson:
@stone-z 
How do you see workload identity having any relation to logging? I would be interested to know why you think they would be related? To me, tenancy is about who has access to the logs so maybe the identity would then be a label in the log but it should not defined a user tenancy model.


> Question/opinion - this seems like an enterprise feature of the underlying tools -- do you know of or anticipate any artificial limits from that? e.g. is there a limit to the number of tenant that would actually make per-app tenancy impossible?

I think there is no limits on a technical level (max header size can be configured in most ingress controllers) but I think this would definitely add a lot more load to the read path as loki does at least 1 query per tenant even if we know we query only 1 actual tenant from what I understood of the code and that is still dependant on the duration of logs you want to retrieve as it would most likely retrieve multiple chunks per tenant and the number of read queries would be too much in that case.


Other that that, after some self monologue, I'm starting to envision a bit some of the technicalities of this.

What I would think make sense would be to allow customers to configure the tenancy they want at the cluster level (cluster label for instance) with the allowed tenancy values of:
- cluster - all logs apart from gs logs are mapped to 1 tenant. I would think this make sense for small clusters where there is no real use for cross namespace tenancy.
- namespace - all logs that are a part of a namespace are mapped to 1 tenant for all clusters (e.g. prod namespace on all clusters would be in the same tenant).
- cluster_namespace - all logs that are in 1 namespace in 1 cluster are in a tenant .
- project - this would then be dependant on the project crs so it's still unclear

logs from kube-system, giantswarm and I guess kyverno namespaces would anyway be sent to the giantswarm tenant on our side and platform teams would have access the giantswarm tenant configured

We can definitely allow exceptions in here by creating a label like "application.giantswarm.io/tenant" that could be set on a namespace or a deployment that would override the tenant for specific things or maybe add the PodLogs CR that comes with the grafana agent.

Now I'm a bit more concerned about the following things:
- What do we do with managed apps that are not deployed in their own namespace or kube-system. For instance, a lot of our customers install prometheus-operator in the monitoring namespace as the kube-system namespace is inaccessible to teams and they can deploy their own prometheus. But then we can also go even deeper in that questions. In this specific case, I'm not sure what the tenant should be or if we then need to store logs twice.

@giantswarm/team-atlas I would like your thoughs on this

multi-tenancy-observability/README.md
stone-z: > loki will actually run at least number of apps queries

The actual query implementation is beyond my knowledge :sweat_smile: Sure, it might be prohibitively expensive or that this is intentionally nerfed in OSS and reserved for enterprise versions.

BUT to the greatest extent possible, I would double down on the likelihood of needing app-level granularity support. Some other examples:
- any type of less-trusted collaborators (externals, students, etc.) who should have access to only certain components inside a namespace
- selective access to "system" resources - for example, there are multiple types of users who will need access to coredns or nginx, but they shouldn't have access to everything in the namespace, especially kube-system stuff. 
- automation - similar to above, but for machine accounts. 

If it is possible to change the tenant behavior later, maybe logging-operator could default to tenant-per-ns until the first app-level rule is applied, or create app tenants only for those targeted by allow rules. This probably means the capability would only apply to future logs, which is a tradeoff, but could be a workable middle ground.

multi-tenancy-observability/README.md
stone-z: Sure, maybe the Project CR ends up being MC-scoped, it just hasn't been decided. It definitely needs more wider discussion. My comment here was mostly to consider if there are aspects of logging that need to be under dev control, and if so, what that interface would look like.

multi-tenancy-observability/README.md
stone-z: You could probably roll out slowly (and under shared control) with Policy API. We can discuss

stone-z (commented):
> How do you see workload identity having any relation to logging? I would be interested to know why you think they would be related? To me, tenancy is about who has access to the logs so maybe the identity would then be a label in the log but it should not defined a user tenancy model.

Having workload identities would make it easier to manage the access rights for human (or machine) users to application logs. Take GS or a customer monitoring team as a potential group of humans. That group of humans should have access to prometheus logs across the fleet. But there are multiple types of prometheus apps and instances hanging out in the fleet. How do we tell them apart in a clear enough way that an operator can configure access rules reliably? It isn't a requirement now though, we just have multiple places to configure things.

Regarding the rest of your larger comment:
I think we're starting to get more on the same page. As long as there is no technical reason it isn't possible, we could let customers choose a tenancy model so that they/we can balance loki performance with the actual RBAC needs. I still think at least some will need app<-->tenant (see my other comment), and that ensures the widest coverage later on, but it could be turned down for customers that don't need it.

I also echoed some of your ideas about shared namespaces in my other comment. It might be useful to discuss all this (identity/tenancy needs, GS Project vs dev absractions) in the next SIG Architecture as well. I have lots of ideas about these at a platform level, but they haven't been refined outside Shield and Atlas is now the first with a real need

QuentinBisson:
Sure let's discuss it on wednesday but we might need to make this an hour long meeting :D

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Our Logging stack
+
+Our logging solution Loki supports multi-tenancy based on a http header (`X-Org-ID`) and we already had to implement some basic form of multi-tenancy to be able to support the logs coming from our clusters.
+
+The graph below shows our current implementation of the multi-tenancy on logs:
+
+<img src="./assets/scope-orgid.png" width="500" alt="Loki multi-tenancy and X-Scope-OrgID header">
+
+In summary, we use a Reverse Proxy in front of all Loki components. This Proxy sets the `X-Org-ID` header based on the authentication information it receives. The proxy configuration (tenant and user/password keypairs) are managed by an operator named `logging-operator`. This operator also configure the tenant and user/password keypairs information for the logging agents that they then use to send logs to loki via the Reverse Proxy.
```
QuantumEnigmaa, line 45:
```suggestion
In summary, we use a Reverse Proxy in front of all Loki components. This Proxy sets the `X-Org-ID` header based on the authentication information it receives. The proxy configuration (tenant and user/password keypairs) are managed by an operator named `logging-operator`. This operator also configures the tenant and user/password keypairs information for the logging agents that they then use to send logs to loki via the Reverse Proxy.
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+But we also want to provide flexibility to customers to allow them to isolate their data as they want.
+For instance, some customers may want to map:
+- 1 tenant to 1 namespace
+- 1 tenant to 1 cluster
+- multiple tenants to 1 application (e.g. ingress controllers)
+We would like to propose our customers the option of defining their tenants so that they can isolate the data as they want.
+
+__Disclaimer:__ We might change this flexibility later on based on product (e.g. project CR) or operational decisions ( e.g. too many tenants).
+
+In this document we will look at the multi-tenancy concept on the example of the logging implementation (with `Loki`) as this is where we already collected experience with it. But we expect to share the same logic across all of our observability stack components.
```
QuantumEnigmaa, line 64:
```suggestion
In this document we will look at the multi-tenancy concept with the example of the logging implementation (with `Loki`) as this is where we already collected experience with it. But we expect to share the same logic across all of our observability stack components.
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+    "giantswarm-ad:Giant Swarm Global",
+    "giantswarm-ad:GiantSwarm",
+    "giantswarm-ad:Giant Swarm EU",
+    "giantswarm-ad:Developers"
+  ],
+  "name": "People People"
+}
+```
+
+The component `loki-multi-tenant-proxy` which has now been renamed to `grafana-multi-tenant-proxy` should be configured with the mapping between groups of people and tenants
```
QuantumEnigmaa, line 105: Maybe it could be worth to explain in a short sentence why the name has changed.

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Write path
+
+### Current architecture for the write path
+
+On the write path, we currently use promtail to collect container and machine logs and grafana agent to collect kubernetes events.
+
+We currently need those 2 tools because:
+1. Promtail cannot get kubernetes events
+2. Grafana Agent is going through quite a big rewrite in it's configuration and all features we needed were not supported when we deployed it
```
QuantumEnigmaa, line 176:
```suggestion
2. Grafana Agent is going through quite a big rewrite in its configuration and all features we needed were not supported when we deployed it
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+This approach is especially flexible for customers but we do not think this approach makes sense because the mapping configuration would be impossible to maintain and we would probably face technical limits pretty fast
+
+##### Hybrid mode
+
+This approach would allow customers to be able to set the tenancy as they want (per namespace or per app depending on the use case).
+
+#### Implementation details
+
+The implementation details are at this stage are still an open topic, because no single tool fulfills all of our requirements. Until this RFC we use two tools, grafana-agent and promtail, which are bot not yet feature complete.
```
QuantumEnigmaa, line 212:
```suggestion
The implementation details are at this stage still an open topic, because no single tool fulfills all of our requirements. As for the moment we wrote this RFC we used two tools, grafana-agent and promtail, which are both not yet feature complete.
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+In this approach, based on the tenant information that the customer provided to us, we would generate a user/password pair that the customer would then use in their favorite log shipper to send their logs to Loki.
+
+Pros:
+- This is quite easy for us to implement as we already have the password generation logic
+- We allow customers to use the tool they know and love.
+- We will not receive pages if the tool is not working properly
+
+Cons:
+- This is far from the idea of a platform
+- There is a high risk that configuration will be messy and hard for use to debug when we need to
```
QuantumEnigmaa, line 228:
```suggestion
- There is a high risk that configuration will be messy and hard to use to debug when we need to
```

multi-tenancy-observability/README.md
```diff
@@ -0,0 +1,270 @@
...
+Pros:
+- Quite easy to configure for logs
+
+Cons:
+- Tenant configuration will need to be configured in multiple places (MC for the read path and WC for the write path)
+- Label approach would most likely not work for metrics OOTB (this would require customer to add new labels like metrics port, path and so on).
+
+#### Approach 3: Configure promtail based on the MC tenancy CR
+
+Altough the customer UX for the read path is still unclear, we can envision a CR like a Tenant CR that would map user's roles to namespaces/deployment under a given tenant name. The logging operator would then take care of applying all that logic
```
QuantumEnigmaa, line 246:
```suggestion
Altough the customer UX for the read path is still unclear, we can envision a CR like a Tenant CR that would map users' roles to namespaces/deployments under a given tenant name. The logging operator would then take care of applying all that logic
```

multi-tenancy-observability/README.md
puja108: Yes, from what I see, we can add levels of granularity on-demand, and we could even do that opt-in for apps, so it's not by default tenanted by app, but there's additional tenants for apps where we see demand for selective access control (e.g. we might at some point pull nginx, or coredns tenants into our default mapping). WDYT?

multi-tenancy-observability/README.md
puja108: We might in any case need a leveled approach where we have project CR on MC level, and some other CR on WC levels (a project needs to be deployed to x WCs). But even that WC level in some cases lives on the MC, when you look at HelmRelease or App CR for example. Other parts of WC level would live within WC and that's most probably unavoidable. Thus, to me it means we can handle these independently. It depends a lot on the architecture of implemenation, which the con below shows. Handling some delegation from an MC level resource to a WC level one is not a big issue and we do that regularly with Apps anyway.

multi-tenancy-observability/README.md
puja108: Yes, from what I understood a new tenant is a new and separate datastore, which makes duplication of data a topic in that context, but also the issues you mentioned, you basically won't be able to access historical data previous to tenant creation. Not sure about how that looks with queries, but that I would assume should not change as tenants should not influence data, but you never know, maybe @QuentinBisson knows more.

QuentinBisson: Yes you can think of it as a datastore and a tenant is only a http header so there is no reason why customers that want to migrate to 1 tenant per app would lose old logs if they keep the old and new tenants together :)

marieroque:
I would say it's part of an epic: https://github.com/giantswarm/roadmap/issues/3558

QuentinBisson:
Closing, we went a différent way

PR #86, opened 2024-02-12
Provider naming standardisation

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/29790

This RFC looks to align the naming scheme across all providers.

### DIFF
--- a/provider-naming-standardisation/README.md
+++ b/provider-naming-standardisation/README.md
@@ -1,0 +1,53 @@
+---
+creation_date: 2024-02-12
+issues:
+- https://github.com/giantswarm/giantswarm/issues/29790
+owners:
+- https://github.com/orgs/giantswarm/teams/team-horizon
+- https://github.com/orgs/giantswarm/teams/team-phoenix
+- https://github.com/orgs/giantswarm/teams/team-rocket
+state: review
+summary: Giant Swarm cluster providers should use a consistent naming scheme across all platforms.
+---
+
+# Standardisation of provider naming scheme
+
+## Current situation
+
+Depending on the provider, we use a different naming scheme. Our current provider names are:
+
+- CAPA
+- CAPG
+- CAPZ
+- cloud-director
+- openstack
+- vsphere
+
+For CAPA/CAPZ, this is currently unavoidable as the aws/azure names refer to our vintage providers. However, this blocker is removed once all vintage clusters have been migrated to CAPI.
+
+Having this split in naming conventions causes some confusion, and whilst this is relatively minor it makes sense to pick a specific scheme (especially before adding any more providers). There appears to be no documented reason for choosing either scheme (aside from the aforementioned collision with our vintage providers) - this RFC aims to address that.
+
+## Moving forward
+
+There are two options moving forward. One is to use the CAPx acronym, and the other is to use the platform name.
+
+### CAPx
+
+Pros:
+
+- Caters for multiple providers per platform (e.g. CAPA and EKS clusters on AWS).
+- More uniform than platform names.
+
+Cons:
+
+- Not as descriptive and harder to grok at a glance.
+
+### Platform name
+
+Pros:
+
+- Descriptive and easy to understand.
+
+Cons:
+
+- Creates a 1:1 mapping between provider and plaform (see CAPA vs EKS above).


erkanerol:
I recall we discussed this topic and decided to use the provider names instead of CAPX format. We said it is better to use what the customer cares about (the provider) instead of what we use under the hood (CAPI). 

Because of vintage, we couldn't use AWS and Azure but we used cloud-director, vsphere and OpenStack. My understanding was that we will use Azure and AWS after the migration. 

Now, we have more scenarios and it is not easy to refer to a case (e.g. EKS) by mentioning only the provider. I am not sure. It seems like a product topic to me. @puja108 @alex-dabija

glitchcrab:
Note that this PR is still WIP, I forgot to set the draft flag when I created it 🤦🏼‍♂️

vxav:
Going through old issue to dig up why we did that.

https://github.com/giantswarm/giantswarm/issues/22671

https://gigantic.slack.com/archives/C01BYMF6RN0/p1671697045575519

https://github.com/giantswarm/roadmap/issues/1543

QuentinBisson:
No it does not clash. We still need a label for managed/unmanaged CPs to be defined by a team but so far so good :) We will have the eks provider in PMO for now

Gacko (approved):

puja108:
> Because of vintage, we couldn't use AWS and Azure but we used cloud-director, vsphere and OpenStack. My understanding was that we will use Azure and AWS after the migration.
> 
> Now, we have more scenarios and it is not easy to refer to a case (e.g. EKS) by mentioning only the provider.

Yes, we wanted to have the customer-facing naming be descriptive and not about the implementation as the issues linked above by Xavier reflect. I do remember @marians gathering this and discussions taking place around this, too, so I'd rather not want to undo the plan that we had.
Unless we see actual issues with that, then we can gladly re-discuss. cc @alex-dabija 

That said, that does not mean that for technical reasons we might use other naming under the hood, e.g. we should not just rename the CAPI controllers or sth like that. If there's specific technical/architectural reasons we need a finer differentiation in monitoring or other config, that should be handled separately.

glitchcrab:
> > Because of vintage, we couldn't use AWS and Azure but we used cloud-director, vsphere and OpenStack. My understanding was that we will use Azure and AWS after the migration.
> > Now, we have more scenarios and it is not easy to refer to a case (e.g. EKS) by mentioning only the provider.
> 
> Yes, we wanted to have the customer-facing naming be descriptive and not about the implementation as the issues linked above by Xavier reflect. I do remember @marians gathering this and discussions taking place around this, too, so I'd rather not want to undo the plan that we had. Unless we see actual issues with that, then we can gladly re-discuss. cc @alex-dabija
> 
> That said, that does not mean that for technical reasons we might use other naming under the hood, e.g. we should not just rename the CAPI controllers or sth like that. If there's specific technical/architectural reasons we need a finer differentiation in monitoring or other config, that should be handled separately.

@puja108 so you're saying the opposite is true - the idea was always that we should rename CAPA/CAPZ once vintage is dead?

puja108:
Yes, at least that's what the issues and discussions that Xavier posted above reflect. In this case we seem to have actually documented sth even if it's in issues and not ADR/PDR.

That said, if people are convinced that the decision was bad and we need to have another discussion, we might need to open the lid on that again. I do not want to block that, just thinking past GS most probably already had that discussion and we might not need to.

glitchcrab:
The main issue with that method is highlighted in the [roadmap issue](https://github.com/giantswarm/roadmap/issues/1543):

> The commonly used cloud provider name should be used as Giant Swarm's provider name.

This doesn't work when multiple provisioning methods can be used on a single provider (e.g. CAPA and EKS on AWS)

Personally I'm not bothered which providers get renamed - it would just be better to have a consistent naming scheme.

puja108:
I'm a bit unclear on the usage and where it would have impacts I'm not aware of, but in general product-wise we totally might have EKS and GS-flavored CAPA clusters next to each other on an aws installation, and IIRC we had some discussions on how we would differentiate the two flavors of CPs, e.g. for monitoring of and scheduling to CPs. Would such differentiation also help here in that case?

JosephSalisbury:
@puja108 see https://github.com/giantswarm/rfc/pull/86#issuecomment-1961407255

ssyno (approved):

JosephSalisbury:
okay, there's a ton of confusion, and i even approved it and let's just clear this up now :D

we have https://github.com/giantswarm/roadmap/issues/1543 from a while back. That aimed to align us along "aws", "azure", etc. lines. As far as I understand from puja and alex, that approach got approved then, we just don't have an rfc for it, ez

the new thing is here the whole "capa and eks are both on aws" (if i understand correctly), which i think is better solved by differentiating between managed and unmanaged control planes (@QuentinBisson do we have an issue for this? we clearly have agreement on it). we obviously also need to do all the alignment work, but that'll be over time (and can happen when we have an agreed direction)

so, saying that - i think the better thing to do is stick with the "original" plan, and align on "aws", "azure", etc., and ensure we can differentiate between managed and unmanaged control planes.

@glitchcrab does that sound reasonable? i'd actually end up closing this RFC (or update the discussion and point it to https://github.com/giantswarm/roadmap/issues/1543?) and let's communicate and align on that

QuentinBisson:
@JosephSalisbury I do not think we do have an issue. This was discussed in kaas-sync https://docs.google.com/document/d/15fBIUmv5pPvCF1DLwR6_XOaFLNHeVex8YNrVtgm83ME/edit and it was left at @alex-dabija find a team to own this

gawertm:
I am fine with what you suggested @JosephSalisbury.
In a standup in Team Rocket, we were just wondering how realistic it is to rename CAPA to AWS after the Vintage Migration. We will for quite some time still have Vintage Repos I believe. they may conflict with naming or or at least confuse. But if you think it's still good to go that way (and especially if Phoenix @T-Kukawka is fine), that's obviously good for us in Rocket :)
Just for background, why we re-opened that discussion

JosephSalisbury:
yeah, i think we'll have the naming confusion for a while whatever we do, it's not like we should stop the world to align all this naming, we'll have to do it over time

my reckoning is as long as we agree (and i think it's easiest for us to agree on https://github.com/giantswarm/roadmap/issues/1543), then we'll get there eventually

marians:
I get the impression that we need a provider taxonomy with two levels.

The higher level would be AWS, Azure, Google etc., and maybe something like "on-prem". This differentiation is relevant when talking about many features, e. g. in docs. Example: "This function is only available on AWS."

The lower level would differentiate further and would allow to tell apart CAPA from EKS. Maybe this level does not exist for all of the above higher level providers, or maybe there is only one valid value for some higher level provider.

provider-naming-standardisation/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Cons:
+
+- Creates a 1:1 mapping between provider and plaform (see CAPA vs EKS above).
+
+## Outcome
+
+I would suggest that given the possibility of having multiple providers for a single platform, we should rename gcp/vcd/vsphere/openstack in order to align with the CAPx naming scheme.
+
+### Changes
```
AndiDog, line 63: If we decide for this, I think we should at the same time decide how to distinguish sub-providers (CAPA EC2 vs. CAPA EKS, CAPZ with VMs vs. later on maybe CAPZ with AKS, ...)

vxav: > The higher level would be AWS, Azure, Google etc., and maybe something like "on-prem". 

I would stay away from cloud vs on-prem. At the moment the on-prem providers are vcd and vsphere but in the future we might have bare-metal, nutanix or who knows what, which will be very different (like capa vs capv different).

- AWS
  - CAPI (or CAPA)
  - Vintage
  - EKS
- Azure
  - CAPI (or CAPZ)
  - AKS
- vSphere (only 1 flavour)
- Cloud Director (only 1 flavour)
- ...

But that would maybe require quite a significant refactor in the operators to support 2 properties instead of one.

@AndiDog The latest direction is towards the original plan as per https://github.com/giantswarm/rfc/pull/86#issuecomment-1966295463

nprokopic: My 2c here :) 

Big disclaimer: assuming it's future and we just have CAPI, because I don't think we should do any changes to try to make it nicer for vintage+CAPI mix 🙈 

IMO we are trying to put multiple properties in to a provider name here.

To take a few steps back, we have these providers in CAPI product ("providers" as actual names of cloud or on-prem providers like AWS and Azure, not names of projects like CAPA and CAPZ):
- AWS
- Azure
- vSphere
- Cloud Director

Then some of these above have 2 "flavours" (not sure if there is a better single word), where one "flavour" is the one where control plane is deployed by kubeadm and CAPA controllers, and the other one is where the control plane is fully managed by the provider itself. So above expanded would be:
- AWS
  - Self-managed or CAPI-managed control plane
  - Provider-managed control plane (EKS)
- Azure
  - Self-managed or CAPI-managed control plane
  - Provider-managed control plane (AKS)
- vSphere
- Cloud Director

To elaborate on the difference:
  - Self-managed or CAPI-managed control plane (kubeadm is doing the provisioning, CAPI+CAPx controllers manage the nodes)
  - Provider-managed control plane - cloud provider fully manages the control plane nodes, maybe you get to configure something but that's it, you don't deploy them and maybe you don't even see them.

So IMHO we should not hide the above difference behind a provider name, but we should introduce managed vs self-managed concept here, because that is essentially the difference.

So if needed in manifests somewhere, or if you have to specify the provider name (e.g. kubectl-gs and other tooling), we could have something like this and you specify the other info as well:
```
provider: aws
managed: true/false
---
provider: azure
managed: true/false
---
provider: vsphere
---
provider: cloud-director
```

Then we might even have to expand the above with `adopted: true/false` or `imported: true/false` (for the clusters that we have not deployed initially, but we have adopted/imported them with CrossPlane I think).

With the above in mind, we could think how to name components, so we could maybe have:
- `cluster-aws` + `cluster-eks` (or `cluster-eks` could be `cluster-aws-managed` but I don't think it matters)
- `aws-app-collection` and `eks-app-collection`/`aws-managed-app-collection`/`managed-aws-app-collection` (or just a single `aws-app-collection` if it works for bot self-managed and managed)

nprokopic: Even in upstream CAPI providers the names like CAPA and CAPZ are not provider names, and also them have AWS and Azure as provider names, so it's:
- cluster-api-provider-aws, 
- cluster-api-provider-azure,
- etc

This means that cluster-api-provider-aws (provider being AWS) covers both self-managed clusters (where kubeadm does the provisioning) and clusters managed by the cloud provider (where EKS manages the control plane).

IMO we should try to align our terminology with this, so the name of the provider is the same, but we have a new "parameter" that indicates if is it "self-managed" (kubeadm) or managed by the cloud-provider.

PR #90, opened 2024-02-26
Update Container Registry Configuration RFC for new structure

erkanerol:
Towards https://github.com/giantswarm/giantswarm/issues/29973

### DIFF
--- a/container-registry-configuration/README.md
+++ b/container-registry-configuration/README.md
@@ -1,6 +1,12 @@
 ---
 creation_date: 2023-03-01
+issues:
+- https://github.com/giantswarm/giantswarm/issues/24237
+last_review_date: 2024-02-26
+owners:
+- https://github.com/orgs/giantswarm/teams/team-turtles
 state: approved
+summary: This RFC is to specify how we configure Kubernetes clusters for container registries including the API in our charts and account management for installations.
 ---
 
 # Container Registry Configuration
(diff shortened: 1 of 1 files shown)


container-registry-configuration/README.md
```diff
@@ -1,6 +1,12 @@
 ---
 creation_date: 2023-03-01
+issues:
+- https://github.com/giantswarm/giantswarm/issues/24237
+last_review_date: 2024-02-26
+owners:
+- https://github.com/orgs/giantswarm/teams/team-turtles
```
erkanerol, line 7: I am not sure about this. I created this RFC and Rocket and Phoenix implemented this in cluster-<provider> charts. Now, @giantswarm/team-turtles is responsible for this kind of provider-independent thing. 

I wonder if I should write my name. @AndiDog

container-registry-configuration/README.md
```diff
@@ -1,6 +1,12 @@
...
 state: approved
+summary: This RFC is to specify how we configure Kubernetes clusters for container registries including the API in our charts and account management for installations.
```
AndiDog, line 9: The summary must not be an introductory sentence, but only describe the outcome of the RFC (at best without "fill words"). Here's a suggestion:

```suggestion
summary: Since Docker Hub has an image download rate limit which can lead to unhealthy clusters, configure containerd such that it uses other registries as mirrors (fallback).
```

container-registry-configuration/README.md
AndiDog: Fine to have only Team Turtles here, since they'll need to make the containerd config consistent across providers.

Gacko (approved):

AndiDog (approved):

PR #98, opened 2024-05-31
Add RFC: CLI command for creating a new cluster

marians:
Towards https://github.com/giantswarm/roadmap/issues/2115

Feedback is welcome.

### DIFF
--- a/cluster-creation-cli/README.md
+++ b/cluster-creation-cli/README.md
@@ -1,0 +1,163 @@
+---
+creation_date: 2024-05-28
+issues:
+  - https://github.com/giantswarm/roadmap/issues/2115
+owners:
+  - https://github.com/orgs/giantswarm/teams/team-honeybadger
+state: review
+summary: Specification of a kubectl-gs command for creating a workload cluster in the Cluster API generation of the Giant Swarm platform
+---
+
+# CLI command for creating a new cluster
+
+This RFC proposes the spec of a new command line interface (CLI) command to create a workload cluster in the Giant Swarm platform based on the Kubernetes Cluster API.
+
+## Terminology
+
+- **CLI** or **command line interface** refers to a text-based user interface to interact with a computer program.
+
+- **Configuration** in the context of this RFC refers to the configuration of a workload cluster. This configuration is a set of properties that define the desired state of the cluster.
+
+- **Cluster app** refers to a Helm chart that defines the resources to be created in the workload cluster. The chart is maintained by Giant Swarm and is versioned.
+
+## Background
+
+So far, Giant Swarm users created workload clusters mainly using these two methods:
+
+- Via the `kubectl gs template cluster` CLI command
+- Via the web user interface (happa)
+
+As the command name `template cluster` suggests, the existing CLI method has been introduced with the goal to simplify the creation of a manifest, mainly by filling in some placeholders of a template. During the evolution that followed, flags have been added to allow configuring more and more details. But the command never kept up with the development of the underlying APIs. There were always some options that flags were missing for. Often, single flags were added and new releases were published, only to enable one more feature. As a result, users were required to stay up-to-date with their CLI, only to have all configuration options for clusters available. In addition, the multitude of flags made it hard for users to discover options. In addition, differences between providers sometimes required provider specific flags, while some flags could be used across (some) providers. Also the command never accounted for the differences between (vintage) releases, where a flag would only make sense in certain releases.
+
+In short, the existing CLI method is extremely hard to use. For Cluster API (CAPI), we therefore decided to use a different approach.
+In CAPI, clusters are deployed as apps (helm charts) via the app platform. Each app has a default configuration and users can override defaults according to their needs. The configuration schema is provided with the app and thoroughly maintained, including property titles, descriptions, example values, and value constraints like, for example, string validation patterns.
+
+Our web user interface (happa) leverages these app's schema to dynamically generate a form which can be used to create a cluster configuration. Annotations from the schema, like title, descriptions, and examples, help the user decide which default values to override. Constraints help prevent mistakes like entering unexpected characters or numeric values outside certain limits. As we move towards Backstage plugins as our web UI, we plan to continue using this approach.
+
+The new CLI command drafted here is supposed to make use of the above concepts, too. We want the command to work hand in hand with our web UI. More about this in the next section (TODO: which one?).
+
+## Goals
+
+- The command should provide an easy-to-use method for vcreating clusters, optionally assisted by a web UI to discover and configure options.
+- The command should work the same way across all providers we support with Cluster API.
+- The command code should require little maintenance. It should stay compatible with a large range of cluster app versions.
+
+## Target use case
+
+The most important use case we target with our CLI command is the creation of a short-lived cluster for testing purposes, as this is the likely the most frequent reason to create a cluster.
+
+Production clusters, in contrast, are created much more rarely, often based on the experience gathered through testing with various clusters. Production clusters are also more likely to be provisioned through GitOps, while test clusters are often created and deleted ad-hoc.
+
+## Requirements
+
+- Users should be able to **re-use a configuration** they have used for cluster creation previously, optionally overriding only a few details.
+
+- A **dry run option** should allow "previewing" the result without actually writing any resources to the management cluster.
+
+- A **connection with the management cluster** should only be required to write resources. For a dry run, no connection should be necessary. (However, an internet connection will be required to access public information from GitHub repositories.)
+
+- Configuration provided by the user should be **validated** against the apps's schema, and in case of validation errors, users should receive detailed feedback.
+
+- Users should decide whether to actually create the cluster, or print a manifest only.
+
+- A **cluster name must be provided by the user**. We won't generate a cluster name for the user.
+  - At Giant Swarm we decided around November 2023 that we want users to set workload cluster names and avoid auto-generated names ([roadmap#2999](https://github.com/giantswarm/roadmap/issues/2999)). In theory, we could still offer an option (a flag) to auto-generate the name. However, this would require knowledge of the cluster-PROVIDER app schema in the command logic. For example, as of cluster-aws v0.73.0, the cluster name is set in the property .global.metadata.name. Although we are working towards alignment across providers, this may be different in other versions or other provider apps.
+
+## Basic syntax
+
+The command should be named `create cluster`, hence the entire command, without any flags, would be:
+
+    kubectl gs create cluster
+
+The most important flags would be:
+
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
+  - TODO: How do we restrict and validate the provider list per installation?
+- `--config`: Path to a config file (optional). Can be used multiple times. See "Configuration layering" below. As an alternative, or in addition, configuration can be passed via standard input.
+- `--app-version`: the cluster-PROVIDER app version to use. If not specified, the latest version is used.
+- `--dry-run`: only validate the configuration and print the resulting manifest.
+- `--output`: file path to write the result to. If not specified, resources are written to the management cluster. The option `STDOUT` can be used to print the result to the console.
+- `--format`: This flag can be used to specify the format of the output.
+  - Default is `manifest` for a Kubernetes manifest that includes an App resources and one or several ConfigMap resources.
+  - `config` for merged configuration YAML.
+  - `command` for a self-contained representation of the command just executed, as a one-liner.
+- `--set`: Override a configuration value. This flag can be used multiple times. The format is `path.to.property=value`.
+  - Note: `--set` can only be used with scalar values (string, number, boolean, integer, etc.). It cannot be used to override an array or object.
+
+TODO:
+
+- How should the resource **namespace** be specified? It could be derived from the organization name included in the cluster config, however that would require logical understanding of the config, which may differ over time and between providers.
+
+### Examples
+
+Ex. 1: Simple cluster creation based on a file
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml
+
+Ex. 2: Simple cluster creation based on standard input
+
+    kubectl gs create cluster --provider capa <<EOF
+    global:
+      metadata:
+        servicePriority: lowest
+        name: test01
+        organization: testorg
+        description: Just a test cluster
+    EOF
+
+While ex. 1 appears simpler and allows for easier re-use of a file-based configuration for a single user, ex. 2 has the benefit of being completely self-contained. It's a pure one-liner without the need for a separate file. This way it's easy to share the command, or copy the command from a website or web UI and paste it into a terminal.
+
+Ex. 3: Overwrite a configuration value
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --set global.metadata.name=test02
+
+Ex. 4: Dry run
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --dry-run
+
+Ex. 5: Write the result to a file
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --output manifest.yaml
+
+## Configuration layering
+
+Configuration can be passed to the command in several ways:
+
+1. Any number of YAML files specified via the `--config` flag.
+2. YAML passed to standard input (STDIN).
+3. Any number of single values defined via the `--set` flag.
+
+The command will create one ConfigMap per source, and one [App](https://docs.giantswarm.io/vintage/use-the-api/management-api/crd/apps.application.giantswarm.io/) resource for the cluster-app. In the App resource, the ConfigMaps will be referenced under the `.spec.extraConfigs` property, which is an array. This way, the number of ConfigMaps is not limited.
+
+The individual ConfigMaps' priority will be configured explicitly in the following order, where later sources overwrite earlier sources:
+
+1. Each file specified via the `--config` flag, in the order they are specified.
+2. Standard input (STDIN).
+3. Each single value specified via the `--set` flag, in the order they are specified.
+
+The ConfigMap representing STDIN always comes last and takes highest priority. This allows to override configuration form one file with a another file specified later in the chain, or with a value passed via STDIN.
+
+In other words, we leverage the App platform's configuration layering principle to provide a flexible templating mechanism, where the underlying layer is always provided by the app's default values.
+As a usage example, this layering could be used to create a cluster using a template configuration file, and overriding some specific values via STDIN:
+
+    kubectl gs create cluster --provider capa --config template.yaml <<EOF
+    global:
+      metadata:
+        name: test02
+    EOF
+
+Alternative syntax for above example, using the `--set` flag:
+
+    kubectl gs create cluster --provider capa --config template.yaml \
+      --set global.metadata.name=test02
+
+This enables the target scenario as described earlier, by re-using most configuration values from one file, and providing the required difference (here: the cluster name) via standard input. The same could of course be achieved by passing a second YAML file via another `--config` flag instead.
+
+In input validation, each source is evaluated against the app's values schema independently. We avoid replicating any values merging logic (as known from helm or the app platform) into the kubectl-gs CLI code base.
+
+## Progress display during creation
+
+If the user chooses to create the cluster immediately (by not specifying an output destination or format), we want to provide detailed progress on the creation. The user should be encouraged to quit watching the progress at any time, without any effect on the provisioning progress.
+
+We may even introduce another subcommand to continue watching the creation progress at any time.
+


cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+Our web user interface (happa) leverages these app's schema to dynamically generate a form which can be used to create a cluster configuration. Annotations from the schema, like title, descriptions, and examples, help the user decide which default values to override. Constraints help prevent mistakes like entering unexpected characters or numeric values outside certain limits. As we move towards Backstage plugins as our web UI, we plan to continue using this approach.
+
+The new CLI command drafted here is supposed to make use of the above concepts, too. We want the command to work hand in hand with our web UI. More about this in the next section (TODO: which one?).
+
+## Goals
+
+- The command should provide an easy-to-use method for vcreating clusters, optionally assisted by a web UI to discover and configure options.
+- The command should work the same way across all providers we support with Cluster API.
+- The command code should require little maintenance. It should stay compatible with a large range of cluster app versions.
```
pipo02mix, line 43:
```suggestion
- The command should provide an easy-to-use method for creating clusters, optionally assisted by a web UI to discover and configure options.
- The command should work the same way across all providers we support with Cluster API.
- The command code should require little maintenance. It should stay compatible with a large range of cluster app versions.
```

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
+  - TODO: How do we restrict and validate the provider list per installation?
+- `--config`: Path to a config file (optional). Can be used multiple times. See "Configuration layering" below. As an alternative, or in addition, configuration can be passed via standard input.
+- `--app-version`: the cluster-PROVIDER app version to use. If not specified, the latest version is used.
+- `--dry-run`: only validate the configuration and print the resulting manifest.
+- `--output`: file path to write the result to. If not specified, resources are written to the management cluster. The option `STDOUT` can be used to print the result to the console.
+- `--format`: This flag can be used to specify the format of the output.
+  - Default is `manifest` for a Kubernetes manifest that includes an App resources and one or several ConfigMap resources.
+  - `config` for merged configuration YAML.
+  - `command` for a self-contained representation of the command just executed, as a one-liner.
```
pipo02mix, line 83:
```suggestion
  - `manifest` is the default for a Kubernetes manifest that includes App resources and one or several ConfigMap resources.
  - `config` for merged configuration YAML.
  - `command` for a self-contained representation of the command just executed, as a one-liner.
```
I don't get what it means, I don't see examples below either. Can we elaborate on what it really does?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+- `--set`: Override a configuration value. This flag can be used multiple times. The format is `path.to.property=value`.
+  - Note: `--set` can only be used with scalar values (string, number, boolean, integer, etc.). It cannot be used to override an array or object.
+
+TODO:
+
+- How should the resource **namespace** be specified? It could be derived from the organization name included in the cluster config, however that would require logical understanding of the config, which may differ over time and between providers.
```
pipo02mix, line 89: But all providers run cluster resources on the org namespace, could we just use that?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+This enables the target scenario as described earlier, by re-using most configuration values from one file, and providing the required difference (here: the cluster name) via standard input. The same could of course be achieved by passing a second YAML file via another `--config` flag instead.
+
+In input validation, each source is evaluated against the app's values schema independently. We avoid replicating any values merging logic (as known from helm or the app platform) into the kubectl-gs CLI code base.
+
+## Progress display during creation
+
+If the user chooses to create the cluster immediately (by not specifying an output destination or format), we want to provide detailed progress on the creation. The user should be encouraged to quit watching the progress at any time, without any effect on the provisioning progress.
+
+We may even introduce another subcommand to continue watching the creation progress at any time.
```
pipo02mix, line 162: I would vote for this one, or -w flag either

cluster-creation-cli/README.md
marians: We can assume in the CLI command implementation that the org name is specified in `.global.metadata.organization` and derive the namespace from that. However, if the cluster app schema changes and the org name field moves into a different place, this implementation will break and have to be updated, with the usual quirks of maintaining backward compatibility and supporting several versions.

pipo02mix: can we make a test to find in case it breaks?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+Ex. 2: Simple cluster creation based on standard input
+
+    kubectl gs create cluster --provider capa <<EOF
+    global:
+      metadata:
+        servicePriority: lowest
+        name: test01
+        organization: testorg
+        description: Just a test cluster
```
stone-z, line 105: In an ideal case, the CLI wouldn't require passing any config unless the user wants to override defaults.

For example, it would be very friendly if this were a correct and complete way to start a cluster:
```
kubectl gs create cluster --provider capa --name test01
```

Based on the discussion in https://github.com/giantswarm/roadmap/issues/2999, `name` will be required and can be supplied via the CLI, but IIUC the others listed here could be defaulted (priority, org) or left empty (description). Is that true?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+    EOF
```
vxav, line 106: Opening a comment here to not sidetrack from the point of @stone-z.
I wanted to point out that for CAPV and CAPVCD we use the name of the cluster app as the cluster name. (I don't know if CAPA/CAPZ moved to this method too).

e.g. https://github.com/giantswarm/cluster-cloud-director/blob/main/helm/cluster-cloud-director/templates/cluster.yaml#L4

cluster-creation-cli/README.md
vxav: I agress on the approach of having a valid config by providing minimal config, however, there will be a few infra-specific parameters (really not a lot) for the onprem providers that we can't default.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+While ex. 1 appears simpler and allows for easier re-use of a file-based configuration for a single user, ex. 2 has the benefit of being completely self-contained. It's a pure one-liner without the need for a separate file. This way it's easy to share the command, or copy the command from a website or web UI and paste it into a terminal.
```
vxav, line 108: I can see this getting outdated pretty quickly as the schema changes, how would we maintain it?  Would it need to be included into the release notes?

cluster-creation-cli/README.md
marians: @stone-z What you describe is really the intended behaviour. We aim to ask only the required details from the users. However, with `kubectl gs create cluster --provider capa --name test01`, the one thing I'm missing is the organization. How can we set a default for organization (and thus namespace)?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+## Basic syntax
+
+The command should be named `create cluster`, hence the entire command, without any flags, would be:
+
+    kubectl gs create cluster
+
+The most important flags would be:
+
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
+  - TODO: How do we restrict and validate the provider list per installation?
```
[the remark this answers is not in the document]
marians: In general, we want to catch mistakes like typos early. The experience of waiting for ~20 minutes, only to find out that you misspelled an argument and you'll have to clean up and try again is what we'd like to avoid if possible.

cluster-creation-cli/README.md
marians: In terms of logic, I think it's the other way around. The CLI command should take the cluster name form user input and use it for the App CR `.metadata.name` field.

cluster-creation-cli/README.md
marians: That's an interesting point. In a previous draft document I had an [entire section](https://docs.google.com/document/d/1l_jK3F9ePEe_Jab0UcLO7u_B24bZNbBwwzPA0hSu-LA/edit#heading=h.w3qtd4lxetxk) about this, but decided to skip it here in order to not bloat the RFC. I'll copy it here:

> ## Cluster app schema evolution and incompatibilities
>
> As cluster apps' schemas evolve, we sometimes introduce incompatible changes. For example, we move properties into a different place. With the specification described above, we will encourage users to keep their cluster configuration values YAML files around for re-use, which will eventually lead into users facing incompatibilities.
>
> Most important in this situation is that these incompatibilities get caught through the validation executed by the `kubectl gs create cluster` command. This way, the command will fail with an error. However, this requires that the schema used for validation does not allow arbitrary properties (additionalProperties) on any object.
>
> Outside the scope of this spec, we may think about further improvements to provide a smooth experience when creating clusters using config files form previous app versions. Some ideas:
>
> - Add values migration into the CLI as a separate command, or even
> - Automatically migrate values from previous schemas into the selected app version's schema
> - Have a schema version and require/allow configuration to specify the schema version to validate against

For now, I think that the best approach we had was the `yq` migration command embedded in the changelog ([example](https://github.com/giantswarm/cluster-aws/blob/main/CHANGELOG.md#0500---2023-12-04)).

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+- `--config`: Path to a config file (optional). Can be used multiple times. See "Configuration layering" below. As an alternative, or in addition, configuration can be passed via standard input.
+- `--app-version`: the cluster-PROVIDER app version to use. If not specified, the latest version is used.
```
puja108, line 77: we need to see how this would interact with the proposed release concept where the cluster version is defined by a release cc @nprokopic

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Examples
+
+Ex. 1: Simple cluster creation based on a file
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml
+
+Ex. 2: Simple cluster creation based on standard input
+
+    kubectl gs create cluster --provider capa <<EOF
```
AverageMarcus, line 99: Instead of this syntax of piping in from stdin I'd rather be specific about what it is related to and use something like `kubectl gs create cluster --provider capa --config -` (note the `-` indicating stdin)

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --set global.metadata.name=test02
+
+Ex. 4: Dry run
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --dry-run
+
+Ex. 5: Write the result to a file
+
+    kubectl gs create cluster --provider capa --config myconfig.yaml --output manifest.yaml
```
AverageMarcus, line 120: As I commented above, I think this would be better to make use of the `--dry-run` and redirect to a file like with other kubectl commands. e.g.
```
kubectl gs create cluster --provider capa --config myconfig.yaml --dry-run > manifest.yaml
```

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+The command will create one ConfigMap per source, and one [App](https://docs.giantswarm.io/vintage/use-the-api/management-api/crd/apps.application.giantswarm.io/) resource for the cluster-app. In the App resource, the ConfigMaps will be referenced under the `.spec.extraConfigs` property, which is an array. This way, the number of ConfigMaps is not limited.
+
+The individual ConfigMaps' priority will be configured explicitly in the following order, where later sources overwrite earlier sources:
+
+1. Each file specified via the `--config` flag, in the order they are specified.
+2. Standard input (STDIN).
+3. Each single value specified via the `--set` flag, in the order they are specified.
+
+The ConfigMap representing STDIN always comes last and takes highest priority. This allows to override configuration form one file with a another file specified later in the chain, or with a value passed via STDIN.
```
AverageMarcus, line 138:
```suggestion
The ConfigMap representing STDIN always comes last and takes highest priority. This allows to override configuration from one file with a another file specified later in the chain, or with a value passed via STDIN.
```

AverageMarcus, line 138: I'm confused by this, you state that the order shown here is the order of priority but then go on to state that stdin has the highest priority even though it is listed second.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+In other words, we leverage the App platform's configuration layering principle to provide a flexible templating mechanism, where the underlying layer is always provided by the app's default values.
+As a usage example, this layering could be used to create a cluster using a template configuration file, and overriding some specific values via STDIN:
+
+    kubectl gs create cluster --provider capa --config template.yaml <<EOF
+    global:
+      metadata:
+        name: test02
+    EOF
+
+Alternative syntax for above example, using the `--set` flag:
```
AverageMarcus, line 149: Does each `--set` property generate a new ConfigMap or do that all get merged into one?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+    kubectl gs create cluster --provider capa --config template.yaml \
+      --set global.metadata.name=test02
+
+This enables the target scenario as described earlier, by re-using most configuration values from one file, and providing the required difference (here: the cluster name) via standard input. The same could of course be achieved by passing a second YAML file via another `--config` flag instead.
+
+In input validation, each source is evaluated against the app's values schema independently. We avoid replicating any values merging logic (as known from helm or the app platform) into the kubectl-gs CLI code base.
```
AverageMarcus, line 156: Wont this break in your above example unless the `template.yaml` has a default name that is always overwrote by a later config?

E.g. I'd really want to have a `template.yaml` that I pass in to all clusters that have the non-changing values and then I can do something like `--set global.metadata.name=test02` for each cluster. But if the validation is applied to the `template.yaml` before the `--set` is added it'll fail because name isn't set.

Or maybe I've misunderstood what this mean?

cluster-creation-cli/README.md
AverageMarcus: I don't think this should be included in this RFC. It's scope creep in my opinion. Watching for cluster creation isn't "technically" related to applying of the cluster app. There's a lot of layers and processes that are involved to make that happen. Instead, we already have the `kubectl gs get clusters` that should be preferred in my opinion.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Progress display during creation
+
+If the user chooses to create the cluster immediately (by not specifying an output destination or format), we want to provide detailed progress on the creation. The user should be encouraged to quit watching the progress at any time, without any effect on the provisioning progress.
+
+We may even introduce another subcommand to continue watching the creation progress at any time.
+
```
AverageMarcus, line 163: This RFC seems to be missing anything about how a cluster would be upgraded/modified after creation 
(can I re-run the create command and it'll overwrite? Will it block me because a cluster with the same name already exists? If it doesn't block, how can I avoid accidentally overwriting someone else's cluster named `test`?)

I think this RFC needs to discuss how / if / where this command would fit into a gitops workflow. We're pretty actively encouraging our customers to use gitops and this approach seems to go against that advice.

cluster-creation-cli/README.md
puja108: I'd not add a default org for now and it would also be a bit weird when a user is a member of several orgs but not all. Only in the single org member case we can clearly set a default.

cluster-creation-cli/README.md
puja108: Hmm, from what I'm reading here, it seems we have an org field separate from the org namespace? Why? This feels wrong to me to duplicate and by that generate implementation blockers for moving and changing namespaces. K8s resources should not have dependencies on their namespace.

cluster-creation-cli/README.md
puja108: Honestly, thinking of this as a kubectl command, kubectl has a default/selected namespace, I kinda feel we should stay true to that if at all.

cluster-creation-cli/README.md
vxav: > if the cluster app schema changes and the org name field moves into a different place, this implementation will break

Isn't it also true for all the other flags?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+Production clusters, in contrast, are created much more rarely, often based on the experience gathered through testing with various clusters. Production clusters are also more likely to be provisioned through GitOps, while test clusters are often created and deleted ad-hoc.
+
+## Requirements
+
+- Users should be able to **re-use a configuration** they have used for cluster creation previously, optionally overriding only a few details.
+
+- A **dry run option** should allow "previewing" the result without actually writing any resources to the management cluster.
+
+- A **connection with the management cluster** should only be required to write resources. For a dry run, no connection should be necessary. (However, an internet connection will be required to access public information from GitHub repositories.)
```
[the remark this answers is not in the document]
vxav: There was a requirement to be logged in an MC of the same provider as the WC to create to get the latest cluster-app and default-apps-app versions if they weren't specified in the flags. I always found this behaviour counter-intuitive so I agree with the proposal. It is even more relevant with regards to hybrid environments (e.g. CAPZ MC with CAPV WC)

cluster-creation-cli/README.md
marians: > Hmm, from what I'm reading here, it seems we have an org field separate from the org namespace?

Well yes, it's part of the cluster config schema. We might get rid of it (outside the scope of this RFC) but then we still need a solution to specify the org namespace.

>> if the cluster app schema changes and the org name field moves into a different place, this implementation will break

> Isn't it also true for all the other flags?

Which other flags? There should not be any duplication between flags and cluster config. There is the `--set` flag, which takes the full property path and the value, so the command implementation does not need knowledge about the schema. Example: `--set global.metadata.name=test02`.

cluster-creation-cli/README.md
marians: IMO everything coming from `--set` should end up in one common ConfigMap.

cluster-creation-cli/README.md
marians: The validation I was thinking of was simply validating config YAML against the config schema. It seems you are thinking of a validation of a rendered manifest, right? Not sure the latter is needed on the CLI side.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+    kubectl gs create cluster
+
+The most important flags would be:
+
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
+  - TODO: How do we restrict and validate the provider list per installation?
+- `--config`: Path to a config file (optional). Can be used multiple times. See "Configuration layering" below. As an alternative, or in addition, configuration can be passed via standard input.
+- `--app-version`: the cluster-PROVIDER app version to use. If not specified, the latest version is used.
+- `--dry-run`: only validate the configuration and print the resulting manifest.
+- `--output`: file path to write the result to. If not specified, resources are written to the management cluster. The option `STDOUT` can be used to print the result to the console.
```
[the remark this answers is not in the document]
marians: I checked back to see how kubectl v1.29 behaves here with `create` commands. The command `kubectl create serviceaccount my-sa --dry-run=client` only prints this:

    serviceaccount/my-sa created (dry run)

To get a YAML manifest printed, I have to add `--output=yaml`.

So the specified flag names are in line with `kubectl create`, but the values are not. First we don't specify different values for `--dry-run` (client, server), second the `--output` values differ. I don't see this being a problem though.

cluster-creation-cli/README.md
marians: Since it's so conventional with `kubectl` to use the `--namespace/-n` flag, probably the most straightforward would be to just to ask for that flag. However, there are some caveats:

- If not specified, the typical (expected by users) default namespace value is `default`. This means a user could accidentally try to create a cluster in that namespace (preventing this would be a matter of permissions or admission control and outside of the CLI's jurisdiction). It also means that a different defaulting behaviour to set some org-namespace would be specific to this command and unexpected. I am not in favour of this.

- As long as there is [`.global.metadata.organization` in the schema as a required property](https://github.com/giantswarm/cluster/blob/v0.29.0/helm/cluster/values.schema.json#L1450), we would ask for redundant input. I am also not in favour of this.

cluster-creation-cli/README.md
marians: See also my comment https://github.com/giantswarm/rfc/pull/98#discussion_r1627104913

cluster-creation-cli/README.md
marians: That's a mistake caused by me adding the `--set` option later. Thanks for pointing it out!

marians: Fixed in 9911b567a65f912e1aeb7bed2330fca0380dd07c

cluster-creation-cli/README.md
marians: Clarified in 9911b567a65f912e1aeb7bed2330fca0380dd07c

cluster-creation-cli/README.md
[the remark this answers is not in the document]
marians: Oh, sorry, I was blind and didn't read my own spec well! The current spec uses `--format` where kubectl uses `--output`. I agree that this is bad and must be fixed. 🙇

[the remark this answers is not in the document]
marians: Cleaned up in fea89c14548f95797d8627ab91203ebfff1b1a68

cluster-creation-cli/README.md
marians: This is how it's specified now.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Users should be able to **re-use a configuration** they have used for cluster creation previously, optionally overriding only a few details.
+
+- A **dry run option** should allow "previewing" the result without actually writing any resources to the management cluster.
+
+- A **connection with the management cluster** should only be required to write resources. For a dry run, no connection should be necessary. (However, an internet connection will be required to access public information from GitHub repositories.)
+
+- Configuration provided by the user should be **validated** against the apps's schema, and in case of validation errors, users should receive detailed feedback.
+
+- Users should decide whether to actually create the cluster, or print a manifest only.
```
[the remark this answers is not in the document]
marians: To me, it's two separate requirements. The solution to both is the `--dry-run` flag.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Basic syntax
+
+The command should be named `create cluster`, hence the entire command, without any flags, would be:
+
+    kubectl gs create cluster
+
+The most important flags would be:
+
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
```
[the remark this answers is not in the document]
marians: That's an important question and I have no answer for it yet.

cluster-creation-cli/README.md
puja108: ok, not part of this RFC to fix this, but we should definitely see that we revisit this org field, I'm still a bit surprised we duplicate info like this and tie things into our proprietary business logic

cluster-creation-cli/README.md
marians: Clarified in 2754756677d4deac32c2b3939bfe9543a44c26e7

marians:
This draft got some great feedback so far. In the recent commits I addressed the parts I could fix/improve. Please read the commit messages for details.

The main comment now includes a list of open questions.

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,334 @@
...
+- **Configuration** in the context of this RFC refers to the configuration of a workload cluster. This configuration is a set of properties that define the desired state of the cluster.
+
+- **Cluster app** refers to a Helm chart that defines the resources to be created in the workload cluster. The chart is maintained by Giant Swarm and is versioned.
+
+## Background
+
+So far, Giant Swarm users created workload clusters mainly using these two methods:
+
+- Via the `kubectl gs template cluster` CLI command
+- Via the web user interface (happa)
```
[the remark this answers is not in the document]
marians: More exactly it is `kubectl gs gitops add workload-cluster`.

Since it has been asked how this spec helps in a GitOps scenario, one open task is still to address that question.

marians:
## Regarding: **Explain how this command fits into a GitOps scenario**

The dry-run output mode named `config-yaml is suited well for use in a GitOps scenario.

Taking our workload-clusters-fleet repository as an example, the config [here](https://github.com/giantswarm/workload-clusters-fleet/blob/main/management-clusters/gazelle/organizations/giantswarm-production/workload-clusters/operations/cluster/cluster_config.yaml) is very close to what the specified command would produce.

There is a difference though, and to add a workload cluster to a GitOps repository, more is needed.

The difference: We use placeholders like `${organization}` in the repo. These are not mandatory though.

For adding a cluster to a GitOps repo, it takes a kustomization like [this one](https://github.com/giantswarm/workload-clusters-fleet/blob/main/management-clusters/gazelle/organizations/giantswarm-production/workload-clusters/operations/cluster/kustomization.yaml) and that customization has to be registered in some list.

Adding these pieces is currently handled by the command `kubectl gs gitops add workload-cluster`. That command can definitely be improved, also to make it more clear where to add a configuration (potentially created through `kubectl gs create cluster --dry-run --output=config-yaml ...`). We might even extend the `kubectl gs gitops add workload-cluster` command to accept the config input options that `kubectl gs create cluster` does. But this is beyond the scope of this RFC.

Another GitOps use case will be creating a cluster via a UI (Backstage scaffolder). In that case, no CLI interaction would be required. The user should either interactively fill in config details via the hierarchical form (based on schema), or -- as a shortcut -- paste/upload config YAML stored locally / obtained from somewhere. This, too, is outside the scope of this RFC.

marians:
As discussed in KaaS sync, my understanding is that users can specify the release version (if they don't want to use the latest, which is the default) instead of specifying a cluster app version. Hence I just replaces the `--app-version` flag with `--release` in https://github.com/giantswarm/rfc/pull/98/commits/bc125d1df171f16dbf4d04d3318b5db66343f586

marians:
With the latest commit https://github.com/giantswarm/rfc/pull/98/commits/0afc0f13a0de675a3165ae1d1c2a7532998e7a8a I've changed the proposal to provide a `--namespace` flag to specify the target namespace. This resolves the open question regarding how to specify the owner organization and/or namespace.

cluster-creation-cli/README.md
marians: I think this is settled now. `--app-version` got replaced by `--release`.

cluster-creation-cli/README.md
marians: After some talking in todays KaaS sync I went with the `--namespace` flag. I dislike that it's one more flag the user will have to set, as "default" will result in non-functioning clusters, but all other solutions seem too involved for now.

cluster-creation-cli/README.md
marians: Marking this as resolved. True, the metadata field in the schema should be deprecated.

pipo02mix (approved):
LGTM

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,353 @@
...
+
+- `--provider`: (required) provider name. For example: capa, capvcd, capz, eks.
+  - TODO: How do we restrict and validate the provider list per installation?
+- `--config`: Path to a config file (optional). Can be used multiple times. See "Configuration layering" below. As an alternative, or in addition, configuration can be passed via standard input.
+- `--release`: the version number of the release to use. The release specifies the version of components to use in the cluster. If not specified, the latest release is used.
+- `--namespace` / `-n`: the namespace in which to create the cluster. This is expected to be an organization namespace, like `org-myorg`.
+- `--dry-run`: only validate the configuration and optionally print the resulting manifest.
+- `--output`: This flag can be used to specify the format of the output when `--dry-run` is set. See the "Dry run output" section below for details.
+  - `yaml` for a Kubernetes manifest that includes an App resource and one or several ConfigMap resources.
+  - `config-yaml` for merged configuration YAML.
```
AndiDog, line 82: What is the difference between an App+ConfigMap manifest vs. "merged configuration YAML"?

cluster-creation-cli/README.md
```diff
@@ -0,0 +1,353 @@
...
+    kubectl gs create cluster \
+      --provider capa \
+      --namespace org-test \
+      --config myconfig.yaml \
+      --dry-run \
+      --output yaml > manifest.yaml
+
+## Dry run output
+
+Combining `--dry-run` and `--output`, users can decide to print a representation of the desired cluster instead of creating it immediately.
```
AndiDog, line 137: For existing users, it's confusing that the command defaults to creating a cluster instead of printing a manifest. In the best sense of a Unix tool, it should do only one thing well, and rely on other tools (namely `| kubectl apply -f -`) to do out-of-scope actions. With GitOps being the overall goal for most users, it's quite surprising and unsafe to have the new command create a workload cluster immediately. CLIs should not use `--something` to change to a different mode – here, I'd expect `--output yaml` to print a manifest, which clearly is a different mode from applying a cluster manifest.

Also, we have lots of existing users of `kubectl gs template cluster` who already rely on certain flags. We'd need to migrate them to the new command and deprecate the command line arguments, or maintain the old command forever. Therefore, old and new command should be compatible. Same for internal users.

AndiDog (commented):
Veto against defaulting to immediate cluster creation 😉

cluster-creation-cli/README.md
marians: The `config-yaml` should only include the values part (like `values.yaml`, not a ConfigMap), while the `yaml` would be several manifests for App, ConfigMap. "Merged" means that when using several sources and flags as an input, the output is only one YAML document.

cluster-creation-cli/README.md
marians: > For existing users, it's confusing that the command defaults to creating a cluster instead of printing a manifest.

I think this is the case when you think of `template cluster` as a reference.

Since this is a new command named `create` cluster, I think it's fair to assume that users understand the intention is to actually create a cluster.

marians:
> Veto against defaulting to immediate cluster creation 😉

@AndiDog With my reply, how do you think about this? Do you still want to veto this?

AndiDog:
Let's please drive this forward in a larger-scale synchronous meeting instead. Opinion juggling won't help here 😉.

puja108:
might need sig arch cc @JosephSalisbury

JosephSalisbury:
@marians feel free to drop in sig arch agenda, but possibly more targeted meeting / andreas chatting with you / honeybadger directly to unblock is easier

marians:
Since Andreas and I obviously have contradicting opinions on this, I would prefer some more people as tie breakers. I can't make it to the SIG architecture sync on Wednesday (the slot usually never suits me). So the next best thing would be SIG product next Monday.

JosephSalisbury:
added as topic to sig product 30/09 - @marians & @AndiDog if you could both join, that'd be grand <3

cluster-creation-cli/README.md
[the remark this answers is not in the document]
AndiDog: Better make it required for now and not use the MC to implicitly determine the type. I'm saying this because we may have more CAPx-WC-on-CAPy-MC mixed cases in the future, and requiring the argument leads to the least surprise.

cluster-creation-cli/README.md
AndiDog: As mentioned in the SIG product call: `yaml` vs. `config-yaml` is unclear. I suggest `--output {app,values}`. And we should also reword the RFC here – "merged configuration YAML" should be turned into an unambiguous explanation.

AndiDog (commented):
After yesterday's SIG Product meeting, we seem to be clear that we really want to put priority on implementing this, so I'm fine to agree on either of the proposed technical implementation details (i.e. remove my veto) in order to avoid minor things to block this.

pipo02mix:
Do we still want to have this command?

marians:
> Do we still want to have this command?

Yes we still want this

PR #99, opened 2024-07-15
Improves template placeholders

JosephSalisbury:
Makes the template placeholders more explicit

### DIFF
--- a/.github/rfc-format/check-rfc-format.py
+++ b/.github/rfc-format/check-rfc-format.py
@@ -233,6 +233,10 @@ def main():
     rfcs_json = []  # sorted in alphabetical order of directory names
 
     for entry_name in sorted(os.listdir(args.rfc_repo_dir)):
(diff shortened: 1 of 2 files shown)


.github/rfc-format/check-rfc-format.py
```diff
@@ -241,6 +241,10 @@ def main():
         if not os.path.exists(readme_file_path):
             continue
 
+        # The RFC template has special formatting, so ignore it in the check
+        if rfc_dir.endswith("template"):
```
AndiDog, line 245: Fine, but let's do this instead at the beginning of the loop so we bail out early

```
if entry_name == 'template':
    continue
```

snizhana-dynnyk (approved):

JosephSalisbury:
@snizhana-dynnyk we're good to merge here once @AndiDog approves 🚀

AndiDog (approved):

PR #106, opened 2024-09-26
First version of app prefix RFC

pipo02mix:
Towards https://github.com/giantswarm/workload-clusters-fleet/pull/802

### DIFF
--- a/app-prefix/README.md
+++ b/app-prefix/README.md
@@ -1,0 +1,46 @@
+---
+creation_date: 2024-09-26
+issues:
+- https://github.com/giantswarm/workload-clusters-fleet/pull/802
+last_review_date: 2024-09-26
+owners:
+- https://github.com/orgs/giantswarm/teams/team-honeybadger
+state: review
+summary: Decide whether the apps in CAPI cluster should have or not a prefix, and if we need to enforce it.
+---
+
+# App prefix
+
+## Intro
+
+When we moved to Cluster API, we decided that all resources would live the organization namespace. That decision means apps, including the cluster app, from different clusters coexist in the same namespace. It also means, it forces us to name the apps differently to avoid conflicts.
+
+## Scope
+
+This affects `kubectl gs` since we have a mix of functionality. In `gitops` subcommand we append the cluster id as a prefix on the output but in `template` subcommand the user needs too add explicitly the `--app-name` with the prefix on it.
+
+When users use the GitOps template, we advocate relying on the `namePrefix` variable to prefix the apps and config maps automatically.
+
+Lately, happa has also prefixed the apps automatically.
+
+We don't validate or mutate the app name via the management cluster API.
+
+## Solution
+
+### API enforcement
+
+We could validate and enforce the name of the apps when deploying to the organization to avoid collisions and conflicts that can unexpectedly affect customers. That way, we could notify customers when they submit an app with no prefix or a prefix that does not match any existing cluster.
+
+At the same time, we should unify the behavior across all the tools and commands (GitOps templates, kubectl and portal).
+
+## Alternatives
+
+### Do not enforce
+
+Leave the customer the freedom to name the apps as they want. This will impact neglected customers who forgot to prefix their apps, potentially unintentionally uninstalling apps or moving to clusters. We will let them choose what type of convention they take.
+
+We could potentially hide it from the customer. This would make the user experience more friendly, but it would also make debugging harder. Our `kubectl-gs` command could get the app without the need for a prefix, or the portal could show the app without containing the prefix.
+
+## References
+
+- [Discussion](https://github.com/giantswarm/workload-clusters-fleet/pull/802)


app-prefix/README.md
```diff
@@ -0,0 +1,46 @@
...
+- https://github.com/orgs/giantswarm/teams/team-honeybadger
+state: review
+summary: Decide whether the apps in CAPI cluster should have or not a prefix, and if we need to enforce it.
+---
+
+# App prefix
+
+## Intro
+
+When we moved to Cluster API, we decided that all resources would live the organization namespace. That decision means apps, including the cluster app, from different clusters coexist in the same namespace. It also means, it forces us to name the apps differently to avoid conflicts.
```
teemow, line 16: @pipo02mix a bit more context is needed here as well:

In the future we would like to separate apps and clusters. Meaning that a development team has access on the management cluster to deploy and manage their app but not a cluster. Maybe even the app can be migrated automatically between clusters by the platform team. So independence between clusters and apps becomes more important.

app-prefix/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This affects `kubectl gs` since we have a mix of functionality. In `gitops` subcommand we append the cluster id as a prefix on the output but in `template` subcommand the user needs too add explicitly the `--app-name` with the prefix on it.
+
+When users use the GitOps template, we advocate relying on the `namePrefix` variable to prefix the apps and config maps automatically.
+
+Lately, happa has also prefixed the apps automatically.
+
+We don't validate or mutate the app name via the management cluster API.
+
+## Solution
```
teemow, line 28: Regarding my comment before and independence of apps and clusters. An alterntive solution would be to do something like pod names for replicasets. Just append some unique identifier to the name of the app.

app-prefix/README.md
pipo02mix: yeah @piontec will help me (next week). I think it can influence the decision, but still we can do what you describing having or not the prefix

app-prefix/README.md
pipo02mix: But then it is same problem the app is not the same as you defined in the manifest. Or we need to create a new resource for the `app instance` entity

piontec:
Hey!
It's hard to provide a strong opinion here, as in general nothing forces us to this naming convention and no other kubernetes objects have that kind of naming convention, even if they live in the same namespace. Today, apps targeting a WC are designated by the `cluster` label - and this is enforced by `app-admission-controller`. Should we put the same information into the app's name and restrict the format of app's name? Well, maybe. We definitely can. Does it make sense? I'm not sure. 

What we're in honeybadger are wondering about: what about other objects? Should we prefix them as well? `HelmReleases`, 'ConfigMaps` etc? 

This restriction seems a bit artificial, as no other kubernetes objects have it, but if we think this will improve user experience, let's do that, the validation code is anyway already in `AAC`. But I would rather make it a recommendation.

uvegla:
FTR admission controllers can also return warnings but still accept objects: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response

pipo02mix:
yeah I dont want to influence in the decision but I would like to have an agreement across the company so we can have standard explanation to customers. I am happy with not enforce or apply it automatically. Using an admission controller to block or validate the apps that collision seems correct to me. @piontec WDYT about developing your points in the RFC?

piontec:
I think that we should just choose the "do not enforce option" as a result and that's it. One last question maybe is to ask if we want "to prefer the cluster-prefix in the app name". Then, we won't enforce it, but we can return a warning as @uvegla suggested.

app-prefix/README.md
pipo02mix, line 16:
```suggestion
When we moved to Cluster API, we decided that all resources would live the organization namespace. That decision means apps, including the cluster app, from different clusters coexist in the same namespace. It also means, it forces us to name the apps differently to avoid conflicts. [More information in the following discussion](https://github.com/giantswarm/workload-clusters-fleet/pull/802#issuecomment-1946218047)
```

pipo02mix:
Okay, @piontec. Can we go ahead and state the decision in the RFC? It means we will modify the `kubectl gs gitops` command to avoid the automatic prefix with the cluster name. We can add the warning though

piontec:
Works for me.

pipo02mix:
is this something HN can do it within this quarter?

piontec:
If you mean "can we add the warning to App Admission Controller", I think we can - doesn't seem to be a big task

app-prefix/README.md
```diff
@@ -0,0 +1,46 @@
+---
+creation_date: 2024-09-26
+issues:
+- https://github.com/giantswarm/workload-clusters-fleet/pull/802
```
pipo02mix, line 4:
```suggestion
- https://github.com/giantswarm/giantswarm/issues/31646
```

pipo02mix:
Please @piontec @teemow approve or comment :)

piontec (approved):
LGTM

PR #107, opened 2024-11-07
Fix broken link CRD RFC

pipo02mix:


luca-rui (approved):
the link is linking 👍

PR #108, opened 2024-12-06
Configuration rendering controller RFC

uvegla:
Towards: https://github.com/giantswarm/roadmap/issues/3773


configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,221 @@
...
+need to do something very similar to fetch and assemble the sources, we could cut this corner now and keep
+using Flux.
+
+There is potential to later implement custom DSL to describe how to merge configurations, which could make sense
+if we want to introduce many different layers in the future (catalogs, workload-cluster, app instance, etc. specific
+configs). For now, it makes sense to keep our current logic as is to reduce time to implement and friction.
+
+### Defining scope of configs to render
+
+TBD
```
piontec, line 210: That's the interesting part. Here I mean by scope "which parts of the source config repos to apply". Let's try to split it into 2 sub-cases: what we want to support now and what we think might be useful in the future (to not block it by decisions we take now).

The only current scope is `Installation` and `App`. I think it's OK to keep it like that to get started. This means, we get exactly the same config for each app as long as we render it for the same installation. We can save the resulting CM&S into the `giantswarm` namespace and have a reasonable starting point, that also mimics how `collections` work now.

In the future, we want:
- WC scope (atlas has a use case for that already) - different config for each WC, but the same App. Currently, this can be generalised as "different config for different app instances". We can think about matching by app CR name (like a name regexp pattern, but that's fragile) or by some labels present on the App CR. 

- maybe org level, but I don't think I've seen a use case for that

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+It would potentially be best to not depend on Flux either, but since we already have it implemented, and we would
+need to do something very similar to fetch and assemble the sources, we could cut this corner now and keep
+using Flux.
+
+There is potential to later implement custom DSL to describe how to merge configurations, which could make sense
+if we want to introduce many different layers in the future (catalogs, workload-cluster, app instance, etc. specific
+configs). For now, it makes sense to keep our current logic as is to reduce time to implement and friction.
+
+### Defining scope of configs to render
```
piontec, line 208: If we mean here "for which App CRs to render the config", I think we should mdoel it after `selectors` in general, based on `nameSelector` or `labelSelector`. The default value is `- '*'` and can be overriden to something more specific, like:

```
target:
  - labelSelector:
      key: app.kubernetes.io/name
      value: flux.*
```

uvegla: It is not about App CRs. The config renderer does not know anything about App Platform. It is to define a sub scope of the config structure to render.

configuration-rendering-controller/README.md
uvegla: I am looking for some common expression language maybe, or ensuring extensibility by setting type to scopes, e.g:

```
scopes:
  - type: managment-cluster
    expression: *

# or

scopes:
  - type: workload-cluster
     expression: operations

# or

scopes:
  - type: app-instances
    expression: kyverno*|cert-manager|external-dns
```

In case of multiple scopes defined, its always additive. All configs that match any of the rules would be rendered.

Another approach would be to define a different CRD per kind of configuration. That is very flexible open for expansion. We can start with 1 for the installation use case we have now, just name it accordingly and later we can add more for catalogs, WCs and instances. With `kubebuilder` we can even have the go structs "inherit" the shared parts to avoid duplication.

uvegla: Detailed this in a new section.

configuration-rendering-controller/README.md
piontec: Yeah, not App CRs, app names from the config repo, so this can't be based on labels. I wonder if we can do it better/more future-proof than just matching on the app config name. What I'm a bit afraid of is that in almost every usage scenario, the decision will be `.*` and the controller will generate a shitload (all) possible CM&S every single time and that can become a performance problem. Maybe it's worth passing the full "path" in the config repo of the most specific file merged in the config, so something like `my-repo/my-mc/(my-namespace)/my-app`? But this can be hard to work with.

Another problem: imagine we introduce, in the future, "namespace" scope of the config repo structure. 
- How do we select which namespace scoped configs are selected? From `targetNamespace`? From an extra selector (so an `and` between app name and namespace name)?
- How do we render the config? Do we just skip namespace scoped configs and render every config matching the app name pattern anyway or do we don't render it, as there's no namespace name match?

Again, I'm not saying this is wrong, but I'm a bit worried that rendering configs "blind", without any info about what's in the cluster, might cause some hard to solve problems in the future. The separation between config generation and usage is really nice, but it's not the key part of the solution here...

configuration-rendering-controller/README.md
```diff
@@ -209,7 +209,174 @@ configs). For now, it makes sense to keep our current logic as is to reduce time
...
+  target:
+    namespace: giantswarm-configuration
+  configuration:
+    cluster:
+      name: gauss
+    applications:
+      regexMatchers:
+        - ".+"
+status:
+  inventory:
```
piontec, line 244: Hmm, is this `inventory` object not too big? We already have something in the range of hundreds of apps, I think? This means this will be huge, hard to work with and changing all the time. Maybe this level of details should be in metrics only?

uvegla: Just metrics is also fine, tho some representation / overview would be nice on the resources. Its simply faster to check than going to Prometheus / Grafana. Tho we could also set up a nice dashboard.

It also came to my mind to generate report CRs, like Trivy or Kyverno.

configuration-rendering-controller/README.md
uvegla: I feel that new / different config levels should potentially be a new CRD. That is partly covered here, thats why the current CRD example is `ManagementClusterConfiguration`, cos that is what we currently do.

stone-z:
I don't have time this week for a more nuanced and in-depth review, but wanted to offer some high-level thoughts before this got old:

1. The general approach is very classic Giant Swarm and I think it could work technically and might meet enough needs for now
2. I generally agree with the direction of offering Giant Swarm abstractions, behind which we can automate things for customers
3. I wonder if there is realistically enough capacity to dedicate to fully migrating to this approach. With the long tails for our releases, I worry about this becoming a _third_ configuration option (obligatory: https://xkcd.com/927/). 
4. Our platform needing both "collection-ish" and "gitops-ish" routes is questionable -- maybe it could just be one or the other. The need for this controller (and I agree, there is current a need) reinforces the idea that our gitops setup does not adequately support our platform. This or a similar controller is fine, and gitops is fine, but having both is a smell. I would probably rather see this implementation work first given to git automation, and using "pure" Flux constructs, and then revisiting any remaining needs to be implemented in GS controllers. (For the record, this isn't a GS-limited problem; I think the current state of the whole gitops landscape is insufficient for managing complicated platforms. My point here is: if we're all-in on gitops -- which I'm told we are -- consider building the stuff that makes gitops sufficient instead of building parallel approaches that conflict with gitops)

5. Not feedback for the controller, but this RFC mentioned something interesting that I want to draw attention to. No action item for this RFC, but:

> For App CRs however, we do have a rule in `app-admission-controller` that rejects an App CR if it references a CM or Secret that does not exist... In an ever-changing Kubernetes environment, it is fair to expect the system to become eventually consistent and self-healing.

Shield has similar problems with admission controllers and platform automation. For instance, it is convenient for developers to store PolicyException-type resources inside the Helm chart for their applications, but exceptions _must_ be deployed strictly _before_ their related workloads, or the entire Helm chart will fail. As a result, our users have to find out-of-band approaches to deliver exceptions, and do a number of other things that should be platform-native. Basically, our platform can't reason about the final state of a workload without deploying it and seeing what happens.

I agree that we can get away with the App CM/Secret case being eventually consistent, but it is another example of this limitation.

General thought: a mature platform should have some degree of awareness of workloads and their needs before deploying them to a cluster. This is beyond the scope of this RFC, but is worth mentioning because I think designing for that capability also informs the design of the configuration delivery.

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+We can reuse `konfigure` as a library in a Kubernetes operator that reconciles a custom CR to pre-render all or a
+given subset of configuration to a destination namespace.
+
+This way, anything is open to mount the configuration config maps and secrets whether they are Giant Swarm App CRs,
+Flux HelmReleases or whatever comes next.
+
+App Platform supports cross namespace references to some extent. It is simply RBAC restricted for workload cluster
+operators. The unique `app-operator` can read CM and Secrets from all namespaces. We can implicitly deploy one instance
+of the new CRD with a scope covering all configured apps to a Giant Swarm restricted namespace, be it `giantswarm` or a new `giantswarm-configs`
```
QuentinBisson, line 91: Is there a specific reason why the configuration CRD could not allow setting the namespace where the config should be with a default being the namespace where the configuration CR is in?

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,481 @@
...
+and self-healing. This is the practice in other operators as well. The problem should be raised when the App CR
+does not get eventually healthy, for which we already have alerts in place. Also, this does not raise a new problem
+but one we already have and is native to the ecosystem we are working on, just think of Flux and
+its CRDs with which it operates but also eventually manages itself with.
+
+Additionally, with the configuration system decoupled, App Platform could also be open for extension of being able to
+detect expected configurations or to react when they change. For example by supporting emitting Kubernetes event
+that App Platform or any component could watch and reconcile the resource earlier.
+
+In my opinion, it is strongly discouraged to automagically mount generated configuration for multiple reasons:
```
QuentinBisson, line 125: I so not specifically agree on this one. Customer quite like the auto mounting for apps that are part of releases because it's harder to maintain them with flux

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,481 @@
...
+- we simply cannot mount it to all App CRs let say, because they do not have them now. Changing their configuration
+  is already a good reason, but also we could also randomly roll apps as well
+- `app-admission-controller` is already burdened with auto mounting logic. That one, as an admission controller
+  however would only react when and App CR gets created or updated. Watching all CMs and Secrets could be proven
+  too much and making changes to App CRs in reaction to that seems like a bad idea from a webhook with time
+  restrictions on them imposed by Kubernetes.
+- `app-operator` also has some auto mounting logic for user values and secrets. If anywhere, maybe in this operator
+  some clever logic could be implemented but that is separate from the scope of the current subject of this RFC
+
+It is important, that new controller for configuration management has absolutely zero knowledge of App Platform or
```
QuentinBisson, line 138: +1

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+There is potential to later implement custom DSL to describe how to merge configurations, which could make sense
+if we want to introduce many different layers in the future (catalogs, workload-cluster, app instance, etc. specific
+configs). For now, it makes sense to keep our current logic as is to reduce time to implement and friction.
+
+### Defining scope of configs to render
+
+```yaml
+apiVersion: configuration.giantswarm.io/v1alpha1
+kind: ManagementClusterConfiguration
```
QuentinBisson, line 224: Why do we need a ManagementClusterConfiguration and not a simple Configuration CRD?

QuentinBisson:
I like the general approach of having a decoupled configuration management and to provide better abstractions which customers could also use. 

I'm stuggling to understand how this would fit as a whole solution for platform teams or customers to use (to say configure cluster chart apps like kyverno only on some specific clusters). Would you be able to draw a schema about all the mechanisms we have now and how this would replace it?

configuration-rendering-controller/README.md
uvegla: We potentially want to extend the layers the configuration system can render that are currently not possible but there is a need for it, e.g. catalog configs, workload cluster configs, app instance configs (same app deployed multiple times and you want different configurations for them).

A single CRD may not be suitable to cram all that into it. We aim to cover what we currently do with the config system and that is technically configuration for MC apps. Therefor a name that represents that feels more future proof.

QuentinBisson: I would have thought this could be done using a hierarchical system of configs as well.

You could have:
- a main config
- a mc specific config
- an app instance config

And the apps would point to a base config so this would be as flexible but less specific in semantics?

configuration-rendering-controller/README.md
uvegla: For one, we can do that tho cos only a portion of apps using a specific deployment methods gets the configs now but there are many other apps. Also there are for example Helm Releases where we cannot add this to the controller but we still want it to take advantage of the config system.

For another, it is crucial that the config system does not do it / has no knowledge of the platform. Later, we can extend App Platform to mount the configs in some ways. E.g. we could support publishing events from the config system and app-operator could watch them. Or other ways too. It is a separate problem.

QuentinBisson: That's true :)

stone-z:
TL;DR - just make a thing that securely merges ConfigMaps and let other tools figure out how to put them there.

---
From the discussion in SIG Architecture, some thoughts on making this a first-class platform feature.
This problem space overlaps with Shield's (and Kyverno's) issues with exception management, so my suggestion is heavily informed by designs used there.

I suggest using Flux (or whatever else) natively to deliver configmaps from git.
This new controller's sole purpose is to merge those ConfigMaps and Secrets according to some logic. (It seems like this should already exist, so gladly point me to the project if it does. I'm pretty sure a chunk of the actual merging logic could be stolen from app-operator already)

Implement:
- a namespace-scoped CRD (let's say MergePool)
- a cluster-scoped CRD (let's say ClusterMergePool)
- a namespace-scoped CRD (let's say MergeConfig)
- a cluster-scoped CRD (let's say ClusterMergeConfig)

MergePools define sets of MergeConfigs.
MergeConfigs define sets of ConfigMaps/Secrets to merge.

IMO this could either be an ordered list, or a selector with some priority mechanism -- dealer's choice.

For example, given this MergePool:
```
apiVersion: configuration.giantswarm.io/v1alpha1
kind: MergePool
metadata:
  name: gauss-configuration-mp
  namespace: giantswarm
spec:
  mergeConfigs:
    - [ordered list of MergeConfigs]
  destination:
    # namespace: not-giantswarm # destination namespace only for ClusterMergePools
    name: some-gauss-configmap
```
or, with "discovery" of future, unknown configs:
```
apiVersion: configuration.giantswarm.io/v1alpha1
kind: MergePool
metadata:
  name: gauss-configuration-mp
  namespace: giantswarm
spec:
  selector:
    - [standard k8s selector logic here]
    - let's say, label=configuration.giantswarm.io/mergepool=gauss-config
  destination:
    # namespace: not-giantswarm # destination namespace only for ClusterMergePools
    name: some-gauss-configmap
```

Then, we create a MergeConfig that fits the selector:
```
apiVersion: configuration.giantswarm.io/v1alpha1
kind: MergeConfig
metadata:
  name: gauss-configuration
  namespace: giantswarm
  labels:
    configuration.giantswarm.io/merge-pool: gauss-config
    configuration.giantswarm.io/merge-priority: 50
spec:
  configMaps:
    - [ordered list]
  secrets:
    - [ordered list]
```
Selection logic is probably insecure for MergeConfigs, so I'd suggest a static list here.

Maybe specifying both CMs and Secrets is unsupported, or the priority is always CMs then Secrets, and the output is always a Secret. Maybe the list contains a priority for the listed CMs/Secrets. Again, dealer's choice.

The controller reconciles the [Cluster]MergeConfigs, and renders an output ConfigMap/Secret.

Additionally:
- exclude all functionality related to secret management (SOPS, vault, etc.)
- exclude all functionality related to cross-cluster config (no following kubeconfigs)
- exclude all knowledge of the config structure, or any git repositories at all
- exclude any features that rely on Apps, including even their names

The result is that the API is
1. additive - Giant Swarm and customers can both provide arbitrarily many levels of defaulting
2. RBAC-controlled - users can be granted access to the output CM without having access to any prior steps or the ability to control the merge. Namespaced MergePools and MergeConfigs can be safely used for ConfigMaps within a single namespace by users granted permission, but they can't abuse "discovery" features to merge their own ConfigMaps.
3. Independent of cluster topology, git structure, and App platform constructs

Giant Swarm can ship and merge as many ConfigMaps as we want without knowing what they are ahead of time. Teams can opt their own ConfigMaps into the merge by adding a MergeConfig with the right pool and priority. Customers can override our values by adding their own MergeConfigs, and if they mess something up, we can add another MergeConfig with a higher priority to override them in a pinch without fighting gitops.

Side implementation note - one way this system breaks is if two existing ConfigMaps are being merged with priority x and x+1, but a user needs to add a new layer between them. If you want to be cool and mathy, maybe there's a way to ensure the priorities are evenly spaced.

configuration-rendering-controller/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+Note: I am not super sure about how the certificates are actually handled for Vault on vintage. On vintage
+`kustomize-controller` instances they are mounted to the controller pod. Would need further research if this is the
+correct approach.
+
+#### About .spec.destination
+
+It is intentionally singular. I would recommend allowing only a single destination per resource.
+
+The simplest destination now is `namespace` within the same cluster. We could potentially allow workload clusters
```
uvegla, line 317: Omitting `kubeConfig`  references for now.

uvegla:
TODO finalize based on comment thread (e.g. remove `.status.inventory` and other small changes)

stone-z:
@uvegla what are your thoughts on https://github.com/giantswarm/rfc/pull/108#issuecomment-2613238517

uvegla:
@stone-z 

> This new controller's sole purpose is to merge those ConfigMaps and Secrets according to some logic. (It seems like this should already exist, so gladly point me to the project if it does. I'm pretty sure a chunk of the actual merging logic could be stolen from app-operator already)

There might be some confusion here cos there are 2 kinds of config "merging":

- one - that this RFC is related to - is config generation that is using go templates and value files. Parts of it are `shared-configs` repo and CCR repos. The component that handles that rendering for MC specific apps is [konfigure](https://github.com/giantswarm/konfigure/) (and on `vintage`, soon to be departed, is [config-controller](https://github.com/giantswarm/config-controller)). Konfigure is literally this logic stripped out from `config-controller` and turned in to a CLI + KRM function to serve under `kustomize-controller`
- and the actual config merging which is merging different layers of configuration, e.g. catalog config, GS defaults, customer overrides, detailed here: https://docs.giantswarm.io/tutorials/fleet-management/app-platform/app-configuration/. This is done in `app-operator`, but to be exact in a separate package: https://github.com/giantswarm/app

> I suggest using Flux (or whatever else) natively to deliver configmaps from git.

Based on the above, I feel a little lost here. Which config maps do you mean? The rendered ones from `shared-configs` + CCR. One problem here is rendering these to git:

- vault usage is still present in some MCs. Rendering these on CI would need access to the Vault. These might go away soon or can easily be SOPS replaced, so this one is no big deal
- the bigger problem is that when `shared-configs` changes all the CCR repo config rendering pipeline would need to trigger, render the configs and merge them to main for Flux to pick them up. We looked into such a triggering mechanism when we did the config split / MCF split migrations and it seemed like a very complicated and hard to track system. Just for one aspect: `shared-configs` should be public (it is not yet, but it depends on vintage IIRC) so a pipeline there should have no knowledge of customer names to trigger CCR repo pipelines. It would also be hard to track and validate / fix issues.

> MergePools define sets of MergeConfigs.

I feel this aims to replace the logic in `app-operator` which would break the App CR interface. Not necessarily, but what is the point of the current config layers exposed by it if we split this merging logic out? If it is decoupled from there, we should simply allow a single secret to be mounted as the final values to pass down to the Helm install / upgrade as a values file, no?

Maybe I am missing something / dont fully understand your proposal. If so, can we please have a chat at some point? Thanks!

stone-z:
Let's have a chat. I think this is still too Giant Swarm-specific / missing the big picture.

stone-z:
Laszlo and I had a nice chat. Basically, we disagree and commit.

Laszlo taught me some stuff and I was surprised by how limited the off-the-shelf capabilities of Flux and Kustomize were for rendering layers of configuration; I had the impression that was more industry standard. I still fully maintain that we _would_ be able to use a controller like the one I described to allow us to use native Flux and offer dynamic config, but concede that it would require some more work to get around dealing with variables and limitations of ConfigMaps (namely, that they need to be flat). Those could be overcome, and I'd argue the complexity is not more than the other behaviors this controller will have, but the RFC suggestion is still a net benefit over the current state, and the two aren't mutually exclusive.

The approach in the RFC:
- is still Giant Swarm-specific, but is more usable by customers because they can define their own target repositories, which is a Good Thing :tm:
- does not address Shield's need for more dynamic config*
- requires/assumes git is the only source of configuration, which is a good goal, but not the case in practice
- allows us to get rid of our kustomize plugin, which is a Good Thing :tm: 

 \* and, we thought of a way Shield can potentially abuse this controller to do what we want.

uvegla:
@stone-z About this part:

> The approach in the RFC:
> ...
> - requires/assumes git is the only source of configuration, which is a good goal, but not the case in practice

Git does not have to be the only source. It can be mixed on the App CR / HR with any other source (CM/Secret), e.g. a user created CM or a secret created by ESO mounted via `extraConfigs` of an App CR or added to the `valuesFrom` of a HelmRelease.

The config rendering does require git at the moment, but thats just because thats how our configuration structure was designed. Could also be extended in the future if it makes.

PR #112, opened 2025-01-30
Add: Location metadata for clusters

marians:


piontec (approved):
I like the idea, but using the standard k8s label, not our own

marians:
Thanks @piontec for the review! FYI: I made some substantial changes after your review, mainly regarding the default location ID system for provide for on-prem.

vxav:
That makes sense to me. Although would we use generic characters if we want to specify the city but not the subdivision? Otherwise the `split(-)[2]` won't always refer to the same category.

pipo02mix (approved):
LGTM

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,332 @@
...
+- easy to use and to understand
+- hierarchical
+- adaptable/customizable
+
+It should be up to the customer to decide if they want to deviate from the system and populate the label mentioned above with whatever string they like. For example, let's just assume a customer has a numeric system to enumerate their facilities. In that case, they should be able to use that identifier instead of our default system.
+
+For our default system, here is a synopsis of the format:
+
+```
+<CONTINENT_CODE>-<COUNTRY_CODE>[-<SUBDIVISION_CODE>[-<CITY_NAME>]]
```
stone-z, line 83: This looks fine to me. I agree with Xav that there could be ambiguity about the meaning of the optional parts, but if we need to rely on that for some reason, that use case probably warrants its own proper support.

Requiring the country to be indicated for all clusters will help with compliance stories. Borders might change, but the label can follow.

I'd offer a preemptive suggestion that there should probably be a required, incrementing number for each location path. The first time a customer has two locations in a given city, we'll have `ab-cd-ef-gh` (which is the first location) and `ab-cd-ef-gh-1` which is, counterintuitively, the second location. It' easy now to start with `-1` and increment.

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,332 @@
...
+  - `sa` for South America
+- `COUNTRY_CODE` is the two letter [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code of the country the cluster is located in.
+- `SUBDIVISION_CODE` is the code for a subdivision of the country, like it might makes sense for larger countries like the U. S. If CITY_NAME is specified, the SUBDIVISION_CODE must be given.
+- `CITY_NAME` is a concise form of the city name the cluster is hosted in, also optional.
+
+### Alternative solutions
+
+- Instead of `topology.kubernetes.io/region`, we could specify our own label in the `giantswarm.io` namespace. The latter would make sense in case we wanted to mark that there would be some logical difference between these labels.
+
+- Instead of strings, we could introduce a system that provides geo coordinates. (Since the cloud providers already work with string identifiers, this draft opts for a solution that extends this paradigm to on-prem.)
```
stone-z, line 112: Something like [what 3 words](https://what3words.com/clip.apples.leap) could also work and maybe be human-friendlier.

However, the exact locations of some sites are secret, so we would definitely have some opt-outs from this approach.

anvddriesch:
I think this makes a lot of sense since I've seen the same pattern in terms of how customers think about their clusters and it will only grow once we have more smart factory installations.
Most importantly imo
- the goal of the default system is for as many customers as possible to use it so it might make sense to get some direct feedback there as well
- automation, monitoring etc on our side should be possible to parse the default system
- customers have to be able to choose any other system they want to use

marians:
> > customers have to be able to choose any other system they want to use
> 
> I agree that it would be nice to allow customers to assign "friendly names" regardless of the provider, but I feel like that should be a separate field from "where is this infrastructure physically located?" I wonder if by allowing total flexibility for friendly naming, we limit the usefulness of the field for our own purposes

To me, friendly naming should happen on the UI level, if the customer is interested. Like this:

If they use our default system, where e. g. `eu-de-he-frankfurt` would be available as a code, they could also make use of our default translation config on the UI level, which translates this to `EU/Germany/Hessia/Franfurt-Main` (for example). If they want to maintain custom codes, they also would have to add custom UI config to get them translated. Otherwise they would be displayed in raw form.

That said, if a customer insisted in setting `topology.kubernetes.io/region` with strings like "Lille, France", they should be allowed to do so.

anvddriesch:
> To me, friendly naming should happen on the UI level, if the customer is interested. Like this:

>If they use our default system, where e. g. eu-de-he-frankfurt would be available as a code, they could also make use of our default translation config on the UI level, which translates this to EU/Germany/Hessia/Franfurt (for example). If they want to maintain custom codes, they also would have to add custom UI config to get them translated. Otherwise they would be displayed in raw form.

>That said, if a customer insisted in setting topology.kubernetes.io/region with strings like "Lille, France", they should be allowed to do so.

Those two solutions sound good to me. :)

marians:
With the recent commit https://github.com/giantswarm/rfc/pull/112/commits/a9d86ca7e8b01e15624e3f3167fed197f3d6e958 I added content related to populating the region label in metrics.

I would welcome feedback on this especially from @giantswarm/team-atlas .

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+- As a platform admin, I want to check the Kubernetes version of all clusters in a certain location.
+- As a service owner, I want to check the deployment status of certain apps in a certain location.
+
+With Cluster API on cloud providers we usually have location indicators in cluster resources, in various places. Each cloud provider uses their own identifier system, however they all make use of a string identifier. The proposed solution in this RFC would help simplify and unify location lookup for clusters by clients throughout cloud providers.
+
+In on-premises installations, no location information is currently available in Kubernetes resources. In metrics (PrometheusMimir) we annotate many series with a `region` label. However, in on-prem installations, the value for this label is always `onprem`. This RFC attempts to change that.
+
+### Key assumptions
+
+This RFC assumes that each cluster exists in one location only, and nodes of the same cluster are not distributed geographically.
```
QuentinBisson, line 33: Did we not just talk about scale to the cloud in the engineering workshop @JosephSalisbury ?

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The solution proposed here consists of two parts:
+
+- Setting consistent metadata on clusters
+- Providing a default location ID system for on-prem customers
+- Setting the `region` label in metrics accordingly
+
+#### Consistent metadata on clusters
+
+We should set the [`topology.kubernetes.io/region`](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion) label on the main cluster resource of all clusters. Currently (as of February 2025) this resource is of kind `Cluster`, group=`cluster.x-k8s.io/v1beta1`.
```
QuentinBisson, line 59: Was this not supposed to be a node label used for scheduling? I am not sure if the topology label makes sense here

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The `topology.kubernetes.io/region` label is documented as w well-known label, set by Kubernetes on `Node` and `PersistentVolume` resources.
+
+For clusters running in cloud providers, the value of the label should be the region identifier. E. g. `eu-west-1` for the AWS region in EU/Dublin.
+
+For clusters on premises, the value should follow a system that we establish for this purpose. The system may differ between customers, to allow fulfilling of customer-specific requirements.
+
+Regardless of the environment (cloud/on-prem), the value of the `topology.kubernetes.io/region` label should not include any whitespace and consist of the lowercase letters, numbers, and dash only (`[a-z0-9-]`).
```
QuentinBisson, line 67: Is there a reason why it needs to be lowercase?

marians: The goal here is to restrict the character space as much as possible. Cloud providers also use lowercase letters, so our system would resemble the existing identifiers better.

cluster-location-metadata/README.md
marians: The linked docs page states that it's used on Node, PersistentVolume. So far it seems it's not used on Cluster.

If we are uncertain about this, we can decide for a custom (giantswarm.io) label instead.

cluster-location-metadata/README.md
QuentinBisson: I would think label values restriction would bé sufficient but alright

cluster-location-metadata/README.md
QuentinBisson: I would rather we use well known labels tbh, i'm wondering if there are not some in the CAPI or kcp work that would fit better but it's still okay as is

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+summary: The document discusses the need for consistent location metadata for clusters, addressing customer communication and management needs. It proposes a unified location identifier metadata system on the cluster level, for cloud and on-premises clusters.
+---
+
+# Location metadata for clusters
+
+### Problem statement
+
+We have anecdotal evidence that customers think about our installations in terms of their physical location. They communicate in terms like these (paraphrased):
+
+- the Dublin management cluster
```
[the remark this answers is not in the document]
QuentinBisson: Alright I missed that

cluster-location-metadata/README.md
JosephSalisbury: yeah, i'm not 100% sure this is going to be true forever

however, even if we do start with some sort of "scale-to-the-cloud" solution in future, i think it's still reasonable to consider the cluster as being housed in one place (i.e: the control plane would still be colocated somewhere)

cluster-location-metadata/README.md
JosephSalisbury: looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set, there's also a maximum length

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Alternative solutions
+
+- Instead of `topology.kubernetes.io/region`, we could specify our own label in the `giantswarm.io` namespace. The latter would make sense in case we wanted to mark that there would be some logical difference between these labels.
+
+- Instead of strings, we could introduce a system that provides geo coordinates. (Since the cloud providers already work with string identifiers, this draft opts for a solution that extends this paradigm to on-prem.)
+
+### Implementation plan
+
+Team Honeybadger would assemble the default hierarchy, at least up to the second level (country), and partly up to the third level (subdivision code). We don't need global coverage with level three initially, as we can still extend our code table once new customers are onboarded or more locations are needed. The maintained code table should be applied as part of the cluster creation form(s), for quick location selection/entry.
```
JosephSalisbury, line 125: this is just for the on-prem clusters, right?

cluster-location-metadata/README.md
```diff
@@ -0,0 +1,343 @@
...
+
+The `topology.kubernetes.io/region` label is documented as w well-known label, set by Kubernetes on `Node` and `PersistentVolume` resources.
+
+For clusters running in cloud providers, the value of the label should be the region identifier. E. g. `eu-west-1` for the AWS region in EU/Dublin.
+
+For clusters on premises, the value should follow a system that we establish for this purpose. The system may differ between customers, to allow fulfilling of customer-specific requirements.
+
+Regardless of the environment (cloud/on-prem), the value of the `topology.kubernetes.io/region` label should not include any whitespace and consist of the lowercase letters, numbers, and dash only (`[a-z0-9-]`).
+
+#### Default location ID system for on-prem
```
JosephSalisbury, line 69: i'm not sure what the level of adoption would be here - i.e: do most customers already have their own naming / labelling system they'd use, and this is somehow unnecessary?

JosephSalisbury (commented):
overall, like the direction, some questions, no blocking

cluster-location-metadata/README.md
QuentinBisson: That's true. This just made me think that maybe we should show the node labels but that would bé for later I guess

cluster-location-metadata/README.md
marians: Right, the default language code system is for on-prem.

cluster-location-metadata/README.md
marians: I don't see any actual system in place in any customer, just strings in descriptions or cluster names. However, if we would simply present a system the customers could use and benefit of (because our UI would support it without the need to customize), I guess they would adopt it eventually.

JosephSalisbury: fair, let's try it

cluster-location-metadata/README.md
stone-z: Bumping the numbering suggestion here -- does anyone have thoughts for/against?

marians: The only downside I see is that the number isn't self-explanatory. But that's what labels are for.

When writing the proposal, I had in mind that the system would have to grow with the customer's needs. To use your example scenario, given `eu-fr-cityname`, they would have to change that to `eu-fr-cityname-foo` and add ``eu-fr-cityname-bar`. But of course it would be less complicated if changing wasn't necessary.

vxav: It makes sense to me to have a trailing numeric ID, it certainly won't hurt at least.
In that case do we need the subdivision code? It's starting to be quite a long string.

marians:
Update:

- I changed from the suggestion to re-use the `topology.kubernetes.io/region` label to introducing our own one, so it's clear that this is our system in our jurisdiction, and no future change in the Kubernetes project can interfere.
- Added the suggestion of a number suffix to custom location codes (optional)

cluster-location-metadata/README.md
marians: I clarified the character restrictions with the latest commit. The length restriction applies to all labels, so I didn't mention that. https://github.com/giantswarm/rfc/pull/112/commits/e73b89cfd869bcad29a27ae977363d5834bcbf97

stone-z:
Thanks for incorporating the number suffix :+1: I'll mull over the missing TOPIC.

Also, maybe we should include consideration of clusters that aren't on earth :stuck_out_tongue: 
... _mostly_ joking

PR #113, opened 2025-02-04
Add Silences GitOps

TheoBrigitte:
Towads: https://github.com/giantswarm/giantswarm/issues/30437

This PR adds a proposal to use MCB and CMC GItOps to manage and deploy Observability Silences CRs to management clusters.

This aims at fixing the current privacy and security issue with the current deployment mechanism and also brings the capability for customer to manage their own silences.


silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+  - name: cluster_id
+    value: some-cluster-id
+    isRegex: false
+```
+
+### Use cases
+
+- Create a silence for all clusters
+  - Add the Silence CR to MCB repository in `bases/silences/` directory
+  - Reference the new Silence CR in the `bases/silences/kustomization.yaml` file
```
QuentinBisson, line 82: I think we could do without this if se use a Kustomization CR right?

QuentinBisson: We should be able to apply the whole `bases/silences` directory

silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Solution proposal
+
+This solution proposes to re-use the current [management-cluster-base][2] (MCB) and customer management cluster (CMC) (e.g. [giantswarm-management-clusters](https://github.com/giantswarm/giantswarm-management-clusters)) GitOps mechanism in place in order to manage and deploy Silences CRs accross management clusters.
+
+This solution is laid out as follows:
+
+- Silence CRs common to all clusters are stored in the MCB repository, under the `bases/silences/` directory
+- Silence CRs added in MCB also need to be referenced in the `bases/silences/kustomization.yaml` file
+- Silence CRs specific to a cluster are stored in the corresponding CMC repository, under the `management-clusters/<management-cluster-name>/silences/` directory
+- Name collision between MCB and CMC Silences CRs is prevented by adding the special `global-` name prefix when deploying Silences CRs from the MCB repository
```
QuentinBisson, line 29: Good catch but I definitely do not like the global prefix (no clue what I would use tbh) but we use shared, common and so on in other places so this feels wrong somehow

silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Benefits
+
+- Silences are deployed and managed using standard GitOps mechanisms
+- Customers can manage their own silences
+- Fixes current privacy and security issues
+
+### Drawbacks
+
+- Rather than using the `targetTags` field to define where the Silence CR is deployed, user now needs to duplicate the Silence CR for each management cluster it needs to be deployed to
```
QuentinBisson, line 104: How is that a drawback? That feels like an advantage to me, no more messing up with regex and so on

silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Create a silence for multiple clusters in a multiple management clusters
+  - Add the Silence CR for each desired management cluster in `management-clusters/<each-management-cluster>/silences/` directory
+  - (Optionally) Add a matcher for the `cluster_id` tag in the Silence CR
+
+### Benefits
+
+- Silences are deployed and managed using standard GitOps mechanisms
+- Customers can manage their own silences
+- Fixes current privacy and security issues
```
QuentinBisson, line 100: Is there anything we cn get rid of?

QuentinBisson: Like the cronjob?

silences-gitops/README.md
TheoBrigitte: I would prefer not to have this kustomization.yaml file, but I haven't found a workaround this.

Only the [`Kustomization.kustomize.toolkit.fluxcd.io`](https://fluxcd.io/flux/components/kustomize/api/v1/) is capable to apply a directory by generating a `kustomization.yaml` for all the yaml resources found in a directory. But this resource works by looking up the specified path in a `sourceRef` (typically a GitRepo in our case), but we have no `GitRepo` for MCB, so we can't use this method.

The other [`Kustomization.kustomize.config.k8s.io`](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/) use a `resources` field which reference location for other `kustomization.yaml` files, they can be chained but the last one must explicitly list in resources files.

silences-gitops/README.md
TheoBrigitte: I am open to other prefix name suggestion here

silences-gitops/README.md
TheoBrigitte: I added this in the `What changes` section, IMO that's not a benefit for the end user.

silences-gitops/README.md
QuentinBisson: I may be wrong but if a silence needs to be in multiple customers then it should be a global silence anyway imo

silences-gitops/README.md
QuentinBisson: Oh I missed that

silences-gitops/README.md
QuentinBisson: Maybe someone from @giantswarm/team-honeybadger have an idea?

silences-gitops/README.md
QuentinBisson: I meant that I do not see a use case where we would have the same silence for 2 different customers that should not be deployed for all installations

QuentinBisson: It is more likely than it should be a provider, region or kaas release specific alert and we should strive to find a better common denomiator than the customer name (which should only be added when pushing data to grafana cloud anyway)

stone-z (commented):
I am fine with this because it _is_ more idiomatic for Giant Swarm, re-using gitops, allowing customer self-service, etc.

Some thoughts:
1. The `silences` repository includes validation of the silences, as well as clean up logic for expired silences. Will this be kept? How will it be managed across all the repositories? Who will own that?
2. It would be good to have a reliable indication of the source on the Silence CR itself. I'm not strongly opinionated how that happens, as long as it's usable by 3am oncall brain.
3. Only because this problem looks _extremely_ similar to a Shield one involving exception management, would you mind adding a short discussion of any alternative solutions you considered? I'm interested in getting an idea what would be required from a platform-level solution to safely deploying things out of band. In this case, Silences fit more nicely into GitOps because _we_ only need to get them onto MCs. However, if customers were to want to use them for their own alerts from inside a WC, or if we would try to re-use the same mechanism for exceptions, it wouldn't work.

QuentinBisson:
@TheoBrigitte I brought up this rfc in sig-arch and @mcharriere and @puja108 had some concerns about this which I hope they will bring up to this RFC :)

mcharriere:
sure thing.
First one is about splitting the silence. Without supporting tooling it can be a bit inconvenient for an oncaller to find the right repo for a given cluster. This could be solved probably when adapting our tools/opsgenie to the new structure.

The second one is that because of operational reasons the `silences` repo doesn't require approval for merging the PRs. But in MCB or CMC repos, the consequences of an unintended change can be expensive.
Is there a way to keep the convenience of not needing an approval but having the certainty that one is not deleting an MC at 3am?

stone-z:
> 1. It's only github actions, there's no reason why we cannot add it to mcb and the CMCs repos using this https://github.com/giantswarm/template-configs

The template solves initial creation, but we should be explicit about who is responsible for keeping it up to date.

> 3. ... The current silence operator is too opinionated for it's own good

Ok, so I take that to mean 1. we don't intend for this to be used for customer alerting and 2. there's no interest in a more general solution for out-of-band stuff at this point. That's fine. Again, just trying to be completely on the same page.

QuentinBisson:
I think having a général out-of-band solution in the next 6 months is important because we will have this exact same question about customer alerts dashboards and silences. I just don't think the silence operator as it is today is the future :)

QuentinBisson:
The initial need here was to avoid having to maintain a secret to pull secrets when we have flux anyway but an out-of-band solution might be there in the future

silences-gitops/README.md
uvegla: There is a feature in `kustomzie` to automatically add `.metadata.name` prefixes and suffixes that you could utilize and "enforce" this: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#prefixsuffix-transformer, for example: https://github.com/giantswarm/konfigure-operator/blob/0d480e9404625b1278e108ae817980405ddb714a/config/default/kustomization.yaml#L9

silences-gitops/README.md
uvegla: You need the `kustomization.yaml` in MCB and then reference that in CMC repos like: https://github.com/giantswarm/giantswarm-management-clusters/blob/bd8f624cb83574cb0467a5410ef0beb75d378f82/management-clusters/golem/catalogs/kustomization.yaml#L11. I think with remote bases there is no workaround that.

I am not a huge fan of the auto generate `kustomization.yaml`. That is a flux only feature and takes away rendering the manifests locally by running `kustomize`. The whole point of `kustomize` is to declare clearly what you want to apply. Flux merely runs the rendering and applies the result.

If sounds reasonable, I would also prefer a separate Flux Kustomization for silences as well. The `flux` Kustomization is getting a bit bloated. But up to you.

silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Use cases
+
+- Create a silence for all clusters
+  - Add the Silence CR to MCB repository in `bases/silences/` directory
+  - Reference the new Silence CR in the `bases/silences/kustomization.yaml` file
+
+- Create a silence for a single cluster
+  - Add the Silence CR to the `management-clusters/some-management-cluster/silences/` directory
+  - Add a matcher for the `cluster_id` tag with the cluster ID as value in the Silence CR
```
uvegla, line 86: Just a thought, to please avoid adding any `cluster_id` or user specific identifiers to silences in MCB. E.g. it could be tempting when you want to create silence across many customers and it would be "painful" to add to n CMC repos.

Another idea: does it make sense to have provider specific silences? We already have provider bases in MCB. That is pulled into the `flux` Kustomziation in all MC, so that kind of contradicts the separate KS for silences. Tho they do not have to be referenced in the provider k8s kustomization file and just added to each MC silences kustomization.yaml as a remote base as well.

silences-gitops/README.md
uvegla: Maybe maybe you could also add a patch to the kustomization to add a label / annotation that it is one from MCB for all silences CRs flowing thru that kustomization. 🤔

silences-gitops/README.md
```diff
@@ -0,0 +1,114 @@
...
+  - (Optionally) Add a matcher for the `cluster_id` tag with all cluster ID as regex value in the Silence CR
+
+- Create a silence for multiple clusters in a multiple management clusters
+  - Add the Silence CR for each desired management cluster in `management-clusters/<each-management-cluster>/silences/` directory
+  - (Optionally) Add a matcher for the `cluster_id` tag in the Silence CR
+
+### Benefits
+
+- Silences are deployed and managed using standard GitOps mechanisms
+- Customers can manage their own silences
```
uvegla, line 99: Well they currently don't necessarily have access to the CMC repos. On the other hand, I am not sure anything - if not RBAC - prevents them to deploy Silence CRs even now, just by other means e.g. from their own repos.

silences-gitops/README.md
uvegla: Proposed something for providers above. Also if a silence needs to be applied to multiple MC of a single customer, it just a matter of code organization: add a `bases/silences` folder in the CMC repo and reference the silences in the MC specific code.

uvegla:
> It's only github actions, there's no reason why we cannot add it to mcb and the CMCs repos using this https://github.com/giantswarm/template-configs

@QuentinBisson GitHub actions could also be shared from MCB as well to all CMC repos. This avoid the need to update everywhere all the time, example:

- https://github.com/giantswarm/management-cluster-bases/blob/fae59f6df60c51d2d05864dd946c9e1f8c562e17/.github/workflows/generate_diffs.yaml
- https://github.com/giantswarm/giantswarm-management-clusters/blob/bd8f624cb83574cb0467a5410ef0beb75d378f82/.github/workflows/generate_diffs.yaml

silences-gitops/README.md
TheoBrigitte: I am exactly using the `namePrefix` there https://github.com/giantswarm/management-cluster-bases/pull/177/files#diff-0a7c294daac0ed3af0691830ed653c0139a0da4b40254d70c701529a9732a64b

Yes having a patch to label Silences CRs with their source repository would be helpful as mentioned by Zach earlier.

silences-gitops/README.md
TheoBrigitte: What's your proposition for providers ?

silences-gitops/README.md
TheoBrigitte: Can you give an example on how the provider based Silences CRs would work ?

TheoBrigitte:
@mcharriere 

> sure thing. First one is about splitting the silence. Without supporting tooling it can be a bit inconvenient for an oncaller to find the right repo for a given cluster. This could be solved probably when adapting our tools/opsgenie to the new structure.

Do you mean that it would be challenging for an oncaller to find where to put the silence ? Do you have any suggestion ?
The assumption was that giantswarm staff is aware of the MCB and CMC structure, and finding the right CMC repo for a cluster should be straightforward as their are base on MC names.

> The second one is that because of operational reasons the `silences` repo doesn't require approval for merging the PRs. But in MCB or CMC repos, the consequences of an unintended change can be expensive. Is there a way to keep the convenience of not needing an approval but having the certainty that one is not deleting an MC at 3am?

True we would need to allow by passing PR approvals, and I don't know how we could prevent mistakes there.

TheoBrigitte:
@stone-z 

> > 1. It's only github actions, there's no reason why we cannot add it to mcb and the CMCs repos using this https://github.com/giantswarm/template-configs
> 
> The template solves initial creation, but we should be explicit about who is responsible for keeping it up to date.

Atlas would be responsible for keeping those github actions up to date, also Laszlo just proposed a nice solution to re-use actions from MCB to CMC repos.
 
> > 3. ... The current silence operator is too opinionated for it's own good
> 
> Ok, so I take that to mean 1. we don't intend for this to be used for customer alerting and 2. there's no interest in a more general solution for out-of-band stuff at this point. That's fine. Again, just trying to be completely on the same page.

No we haven't explored alternative solutions to deploy Silences CRs, honeybadger recommended this approach and we didn't wanted to re-invent the wheel.
Given the current direction of "opening the observability platform to customer", custom could in the future use our infrastructure for their own alerting needs, ideally they could deploy their silences this way but this might not be possible according to Laszlo.

mcharriere:
> Do you mean that it would be challenging for an oncaller to find where to put the silence ? 

Yes, it's not only about finding the repo but also they have certain structure and there isn't a unique way to create resources. So we move from something simple, 1 repo with a flat structure; to something people need to think and choose where to place the silence. That's cognitive load right there, that at 3am I'd rather want to avoid.

> Do you have any suggestion ?

One option I can think is to keep the silence repo with a directory structure. Let's say we have one folder per MC; then you can in the CMC for that MC create the flux resources to deploy that path.

Alternatively developing some tooling (maybe a opsctl command helper?) to handle the the repo split and making it uniform.

silences-gitops/README.md
uvegla: > What do you mean with a separate Flux Kustomization for silences ? Here is the concept I had so far
https://github.com/giantswarm/management-cluster-bases/pull/177/files
https://github.com/giantswarm/giantswarm-management-clusters/pull/932/files

This is good, mostly, but that `silences` Flux kustomization should reconcile a `kustomziation.yaml` in the CMC repo. For that you need the replace for the path. Look at how `catalogs` or `flux-extras` Flux Kustomization work for example.

silences-gitops/README.md
uvegla: You simply extend the provider base with a silences folder, add a `kustomization.yaml` in there and reference it in the CMC `kustomization.yaml` that is reconciled by the `silences` Flux Kustomization. So the CMC one would look something like:

```yaml
resources:
  - https://github.com/giantswarm/management-cluster-bases//bases/silences?ref=main
  - https://github.com/giantswarm/management-cluster-bases//bases/provider/capa/flux-v2/silences?ref=main
```

JosephSalisbury:
discussed in sig arch:

- no-one's against moving from silence-operator job approach to using gitops, that makes sense
- some concerns that the user experience would degrade, but to some degree, it's more of a change
- maybe we build some tooling in future to help, but possibly it's better to learn gitops
- plz write "how to silence - the doc" to help folks at 3am

JosephSalisbury (approved):
lfg

silences-gitops/README.md
TheoBrigitte: Alright I'll look into this and I'll ask your review in the implementation.

silences-gitops/README.md
TheoBrigitte: Changed the prefix to `common`

silences-gitops/README.md
TheoBrigitte: Taking this into consideration for the implementation phase

silences-gitops/README.md
TheoBrigitte: Taking this into consideration for the implementation.

TheoBrigitte:
Thank for all the feedback, I took good notes and will try to address what's possible in the implementation.

Here is my takeway:

- Add bases/silences in CMC to be able to have Silences in multiple customer MC
- Add capability for provider specific silences
- Add annotation to identify silences source
- Use kustomization.yaml in CMC repo to align with other Kustomizations (e.g. catalogs, flux-extras)
- Add CI validation to MCB and re-use it accross CMC repos, add atlas as owner
- Change repositories bypass settings to allow Silences creation
- Document how to create a silence
- Build tooling to help with silence creation

Implementation will follow as part of https://github.com/giantswarm/giantswarm/issues/30437

PR #117, opened 2025-04-29
Gateway API - Resource migration & Certs management

mcharriere:
- RFC proposing possible approaches for Ingress->Gateway API resource migration.
- RFC stating the issues we've encountered when dealing with Certs on Gateway API

https://github.com/giantswarm/roadmap/issues/3150

### DIFF
--- a/gateway-api-migration/README.md
+++ b/gateway-api-migration/README.md
@@ -6,7 +6,7 @@ owners:
 - https://github.com/orgs/giantswarm/teams/team-cabbage
 state: review
 summary: |
-    TBD
+    The RFC addresses the challenges of migrating from the current Ingress-based traffic management to the more modern Gateway API.
 ---
 
 # Title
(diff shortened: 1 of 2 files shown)


mcharriere:
As @QuentinBisson pointed out, both solutions described here are not mutually exclusive:
- The base code/logic for the CLI tool and the operator are similar.
- Even when going for the alternative solution, the CLI can be convenient for the initial stages (with a trimmed down output)
- Moving from CLI to operator when the core logic is the same should not be a problem

Based on that, the main question is we want to expose the customers to maintaining multiple CRs.

ubergesundheit:
I really like the solution presented as "preferred solution" because this aligns with the GitOps approache we're suggesting our customers. With a CLI tool, one could convert their whole GitOps repo in one go.
One thing that we need take care about is collecting "most used" Ingress configurations and take care of testing the translated CRs against a running Gateway API implementation

gateway-api-migration/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+The key advantage of this strategy is that it provides us with a high degree of control over how Ingress features are translated into the Gateway API ecosystem. By directly generating the required CRs, including those specific to our chosen Gateway API implementation (e.g., Envoy Gateway CRs), as well as related resources like Certificates and DNSEndpoints, based on the annotations present in the original Ingress, we can effectively bridge the feature parity gap and address the current variability in ecosystem maturity.
+
+However, a consideration is that this approach places the maintenance of these generated resources with our customers. They will be responsible for updating the generated manifests to incorporate new Gateway API features, adapt to specification changes, and potentially manage the lifecycle of the supplementary CRs.
```
puja108, line 41: I mean this is how they already are using Ingress resources, and I doubt we want to add our own level of abstraction in between there. I'd leave that open to the customer platform team to decide if they want an abstraction layer.

Additionally, this approach is independent of how the resources are applied to the cluster and can adapt to any CICD or GitOps process, as it only needs to translate the resources once.

That said, we might need to talk to actual customers to find out what type of tooling they prefer and what other preferences they have, e.g. what input the CLI should be able to act on.

mcharriere: Just to ble clear, I'm not talking about moving between implementations, but rather moving from **implementation specific CRs** to **standard gateway api CRs** when features are added the later in the future.

Let's put an example:
- Today, if you need to add auth to your HTTPRoutes you need a `SecurityPolicy CR` from Envoy Gateway. There's no way to do that with standard gateway api resources.
- There's [work](https://gateway-api.sigs.k8s.io/geps/gep-1494/) being done in Gateway API to include that feature; but AFAIK it's still not known which form would take. 
- Then customers will need to adapt their deployments from the `SecurityPolicy CR` to whatever standard CR comes from GEP-1494

This scenario is most likely to happen because implementators are covering the missing features with their custom CRs.

pipo02mix:
@mcharriere and I talked. This is now achieved using ingress2gateway (supports EG resourcces). Closing

PR #120, opened 2025-05-29
Add version upgrade operator

piontec:


version-upgrade-operator/README.md
```diff
@@ -0,0 +1,484 @@
...
+To configure an automated update for an App CR, we need answers to 3 key questions: 1) what should be
+upgraded; 2) when it should be upgraded; 3) which version should be used. The following steps have to be
+performed by a user and the new operator (this assumes that there's no GitOps interfering, deployment with
+GitOps is described below):
+
+1. The user deploys an application, either by using `App CR` and a related `Catalog` or `HelmRelease` and
+   `HelmRepository`
+   - at this point, Flux or another related deployment operator are already performing Helm Chart deployment
+1. The user has to define an automated version upgrade policy
+   - if it doesn't exist or a user wants to own the object, he creates an `ImageRepository` object, that
```
fhielpos, line 75: It's a bit confusing to me why it keeps referring to images when the targets are actual Apps or Helm charts.

lyind (commented):
Nice!

I wasn't present when this was discussed and thus assumed making Flux optional is done for a good reason.

stone-z:
For visibility, I am working on a suggestion :construction:

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,491 @@
...
+  validFrom: "2025-05-05T05:05Z" # time window when reconciliation is in power: [2025.05.05 05:05, ∞)
+
+  ## SCHEDULING options - exclusive, choose just one
+
+  # upgradeWindows
+  #
+  # Defines all the possible upgrade windows. The windows are ORed together, so upgrade will be triggered for any
+  # reconciliation that falls within any of the windows. Within a single window, the conditions are ANDed together,
+  # so `dayOfWeek: Wed ` and timeStart: 7:00` means "only on Wednesdays at 7:00".
+  upgradeWindows:
```
puja108, line 199: as the customer interaction we had a few weeks back mentioned, how would we model sth like "first tuesday of the first full week in a month?" 

related: would cron notation not make it easier as there's existing tooling to convert certain desired patterns into cron which might be harder to model into sth custom like we propose below.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,491 @@
...
+     defines the "which version?" part of the controller's input parameters.
+   - The user either creates or reuses `VersionUpgradeSchedule` object (see below). This part delivers the
+     "when?" part of automatic upgrades configuration.
+   - Then, the user creates a `VersionUpgradeConfig` object that is providing only the following information
+     - which `ImagePolicy` to reference and use as a source of truth about available tags
+     - which `VersionUpgradeSchedule` to reference and use as a source of information about when to update the
+       target Resources
+     - how to find target Resources and where to put the version info in them
+
+### GitOps integration
```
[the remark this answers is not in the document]
lyind: Good question. Would users of GitOps even want to use our update operator, prefer e.g. Flux' own auto-update facilities (lacking compared to the proposed operator) or just time the merge of their GitOps PR according to their update window?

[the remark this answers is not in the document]
puja108: We already know that customers with GitOps would like to have this. We have talked to our biggest customer about this already and this feature was actually designed based on pains we and this customer has with upgrades in specific maintenance windows.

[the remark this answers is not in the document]
lyind: @puja108 Thank you for clarification on this.

JosephSalisbury:
from sig arch

- joe: would be cool to see an example or two of the CRs we'd need to upgrade things
    - e.g: what CRs would we need to create to manage upgrades for the adidas clusters? and somehow also for apps: e.g: keep $APP-IN-WC up to date

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,484 @@
...
+- there's no value in the target's version property or it is `latest`
+- the upgrade CR has an optional `defaultVersion` field configured.
+
+If you need to avoid installing (or attempting installation) of the `latest` version (might happen between
+deploying an object and our new operator setting the new version), you might need to first commit the target
+object into repository with a flag like `suspend: True` and flip this flag only after the initial version is
+set.
+
+Please also note, that if you're a GitOps user, resolving directly to Flux's
+[auto upgrade capabilities](https://fluxcd.io/flux/guides/image-update/) might be a better solution for you.
```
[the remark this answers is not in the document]
piontec: This is a different feature, that's aimed to passing info from CI to IAC, but it only triggers reconciliation. You probably can try to hack something with it, but I'd like to avoid such hacky solution, as it means disabling reconciliation (or setting it to something super long).

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,491 @@
...
+
+  ## SCHEDULING options - exclusive, choose just one
+
+  # upgradeWindows
+  #
+  # Defines all the possible upgrade windows. The windows are ORed together, so upgrade will be triggered for any
+  # reconciliation that falls within any of the windows. Within a single window, the conditions are ANDed together,
+  # so `dayOfWeek: Wed ` and timeStart: 7:00` means "only on Wednesdays at 7:00".
+  upgradeWindows:
+    # dayOfWeek: expressed as common short names. Can be a single day ('Mon'), a single range of days ('Mon-Fri'),
```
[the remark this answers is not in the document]
piontec: Status will be provided. Current status proposal is quite verbose and will for sure tuned in the speccing/implementation phase.
As for overlap, I wouldn't block them. Imagine that you want an upgrade on each Monday and each 1st of a month. Should we block such config, because eventually 1st of a Month will be a Monday? I don't think so.

So, we keep the definition "if right now is within any window, we update".

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,491 @@
...
+    # or a comma-separated list of days ('Mon,Wed,Fri'). "*" is an alias for "every day".
+    #
+    # dayOfMonth: number of a day within a month. Can be a single day ('1'), a range of days ('1-5'), or a comma-separated list of days ('1,3,5').
+    #
+    # timeStart: time the window starts, in the format "HH:MMZ" (24-hour clock) (UTC).
+    #
+    # length: how long is the upgrade window opening.
```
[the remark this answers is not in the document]
lyind: Or `windowLength`, but that would be redundant (the object is already called `window...`).

version-upgrade-operator/README.md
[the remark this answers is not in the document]
lyind: Then my question remains, how to control the update window with Flux.

This might be one reason behind the creation of this RFC or out of scope entirely, resolving this.

version-upgrade-operator/README.md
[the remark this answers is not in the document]
piontec: If it's needed: with the way controllers work (events + internal reconciliation loops) we believe it's needed. Without it, some timing might get slippery. But we can also make it optional and by default set to "one reconciliation cycle". If it should be renamed to `duration`: I like it, easier to understand than `length`.

JosephSalisbury (approved):
i think i'm good to go with this, or at least let's start and we can iterate as needed

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,629 @@
...
+
+## Assumptions and requirements
+
+- has to work without GitOps, but must be able to integrate with it
+- has to support schedules
+- works on MC, but targets any resources (so including App CRs and HelmReleases targeting remote WCs)
+
+## Problems
+
+Won't work with the current cluster Release process, where chart version is not really the version deployed.
```
JosephSalisbury, line 28: @alex-dabija to drop it from my brain, i reckon cluster upgrades are the biggest user here, so we need to make sure we get along with the chart version issue

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,629 @@
...
+   - Then, the user creates a `VersionUpgradeConfig` object that is providing only the following information
+     - which `ImagePolicy` to reference and use as a source of truth about available tags
+     - which `VersionUpgradeSchedule` to reference and use as a source of information about when to update the
+       target Resources
+     - how to find target Resources and where to put the version info in them
+
+### GitOps integration
+
+To make it possible to use `VUO` for objects coming from a GitOps repo, the target object's version cannot be
+kept in the GitOps repo anymore (in case it's used). Instead, the operators acting on the target CRs must
```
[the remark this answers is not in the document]
piontec: In theory yes, flux works this way. In practice - this requires us getting write access to customers' gitops repos, which is a no-go in many cases. So, we prefer to assume the new operator is gitops-independent (but possible to integrate)

piontec:
OK, if no one screams in 2-3 days, I'm gonna accept and merge

version-upgrade-operator/README.md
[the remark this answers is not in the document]
uvegla: I think of it in a way that we provide the framework and customers can set it up for themselves / with our help. I think we should support some way to push back to git.

I have some uneasy feeling of updating manifest in the cluster, for example how is an initial version created for a gitops managed resource? It meantioned HelmReleases can have it empty which means latest - if tag exists - then we might downgrade it. But what about App CRs where its a mandatory field? You either need to create the resource in the cluster manually 1st with a given version, then create the gitops part of it + the automation that will eventually kick in. But then its unclear to me how the field ownership is taken over. You cant have a partial CR that is versionless initially in the repo cos that will be an invalid manifest and the CR will be rejected.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+    name: trivy-rc
+    namespace: org-testorg # optional; defaults to the same namespace as the VUC CR
+
+  # defaultVersion
+  #
+  # (Optional) When the operator detects a new matching target that has no version set or the version is `latest`, the version
+  # will be set to `defaultVersion`, even outside an upgrade version. This happens only once, when a resource is
+  # discovered for the first time.
+
+  defaultVersion: "1.0.0"
```
uvegla, line 373: Is this something that should be kept up to date? I guess we can have some checks if this version is within the supported range, but feels like this is a field we would usually need to know what version to apply before we reach the 1st update window, but then people forget about it.

What if the CR - lets say the App or HR CR - is gitops managed then deleted from the cluster. Flux will put it back but the info is lost what the version should be and the app is in a weird state before the next upgrade window.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+  # It could be something simple like `True`, which means we lock current version, or `False`,
+  # which means we don't. But it can be more sophisticated, say for example, it is an enum field
+  # with these values:
+  #
+  # "None" - user may change version of the app, controller won't react. When reconciliation time comes,
+  #          the target field version will be set by the controller.
+  # "Corrective" - user may change version of the app, but during reconciliation, the controller will
+  #                restore last version to it, but takes no responsibility for the damage that switching
+  #                the versions may have caused
+  # "Enforce" - user may not change version of the app, for controller installs countermeasures
```
uvegla, line 395: What kind of installs countermeasures do you have in mind?

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+  # Decides policy towards version changes of targeted objects, that were not introduced by the controller.
+  #
+  # It could be something simple like `True`, which means we lock current version, or `False`,
+  # which means we don't. But it can be more sophisticated, say for example, it is an enum field
+  # with these values:
+  #
+  # "None" - user may change version of the app, controller won't react. When reconciliation time comes,
+  #          the target field version will be set by the controller.
+  # "Corrective" - user may change version of the app, but during reconciliation, the controller will
+  #                restore last version to it, but takes no responsibility for the damage that switching
```
uvegla, line 393: So you mean it will correct it only during the upgrade window or even before the upgrade window?

How do we detect the drift? How do we know what was the version applied in the last upgrade window? What about the 1st upgrade window (tho that ties back to the initial CR creation issue I mentioned above)?

> but takes no responsibility for the damage

I get it but dont get it. What kind of responsibility the operator takes lets say even during the upgrade window?

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+    lastCheckTime: "2025-04-15T09:10:10Z"
+
+  # TODO: think if the status is not too verbose, some info is redundant
+
+  # upgradeAllowed indicates if, based on the suspend, current time and valid the controller
+  # is currently permitted to perform an upgrade if one is needed.
+  upgradeAllowed: false
+
+  # lastAppliedVersion shows what was the version tried during the last version-applying sync
+  lastAppliedVersion: "0.45.0"
```
uvegla, line 455: Maybe we can use this to some of the problems above, but it feels a little risky relying on the status of the object, becasue this is not something recoverable, but something that was and can only be added at special reconcilation loops - not all of them - that takes place during the upgrade window.

If the VUC is deleted then put back by Flux lets say, this info is gone forever.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+  defaultVersion: "1.0.0-rc1"
+  versionUpgradeScheduleRef:
+    name: upgrade-monday-morning
+  versionLock: "None"
+```
+
+The above configuration will:
+
+- set the trivy Helm Chart version to `1.0.0-rc1` on the first reconciliation of the `VUC` resource, even if
+  outside the upgrade window, because `defaultVersion` attribute is set
```
uvegla, line 638: This kind of answers my question from above about what to set on 1st reconciliation. But then this can lead to downgrades and an obsolete field after the rest of the 99.9999% of the lifecycle of this CR.

Would if it work if we set to the current matching latest version if there is no conclusive version available on what to apply? 🤔

uvegla:
I am leaning towards splitting into smaller sub systems, cos each problem is quite complex:

- discovery (ok, we can outsource to Flux CRs for now)
- scheduling
  - my main concern here is that the controller runtime loop / operator pattern does not fit the this scenario very well. Most loops dont do anything. Nothing guarantees in this framework that the resource will actually be reconciled in the upgrade window. On operator restarts all resources will be reconciled automatically, cos the operator does not know when they were last reconciled. You need to pass the requeue parameters at the end of the loop. Technically we can delay next loop until the upgrade window, but again, no guarantees it the pod will last that long. Having 100-150 resources mentioned below can clog the pipeline, but that remains to be seen. All I am saying here is that the operator pattern feels like quite a stretch for this job and there may be dragons and no guarantees to work as we excpect it to.
- applying
  - what to apply, what is the desired state and how to apply?
  - on the how to apply, could we define interfaces for different apply methods? E.g `in-cluster` to apply to a manifest directly on the cluster (this could also be extended in future to support kube config for remote cluster) or `git` to support pushing back to git. The latter would be complex, sure, but it really makes me cringe to lose the main selling point of gitops that is to have the source of truth in the repo.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+1. The user deploys an application, either by using `App CR` and a related `Catalog` or `HelmRelease` and
+   `HelmRepository`
+   - at this point, Flux or another related deployment operator are already performing Helm Chart deployment
+1. The user has to define an automated version upgrade policy
+   - if it doesn't exist or a user wants to own the object, he creates an `ImageRepository` object, that
+     points to a specific repository (application, helm chart) of an OCI registry. `ImageRepository` discovers
+     and knows all the tags in a remote repository of a registry. There's an exclude regexps list to skip
+     stuff you're sure you won't ever need.
+     - The relationship between `ImageRepository` and `HelmRepository` is very close, we can probably automate
+       `ImageRepository` creation so that users don't have to manage them
```
uvegla, line 83: I guess we could automate the creation of these from the targets in the VUC. But then the operator need to know how to parse that info from these CRs, so we cannot work with just any CR. Also if there are multiple targets, nothing really guarantees that they are all the same app actually, so we take the 1st on the list? 🤔 

@ljakimczuk Pointed out that with user creating the ImageRepository + ImagePolicy, with copy pasting users we can end up with applying the latest version of another app. If it fails cos that version does not exists, its okay. The same version existing for that app too is a more scary thing.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+   - The user now creates an `ImagePolicy` object, which determines the range of applicable versions, out of
+     all found by `ImageRepository`. `ImagePolicy` is a filter for `ImageRepository`. You give it a SemVer
+     range and it filters out everything and leaves only the newest tags matching your policy; it's a n:1
+     relationship with `ImageRepository`, so `ImageRepositories` can be shared. The `ImagePolicy` object
+     defines the "which version?" part of the controller's input parameters.
+   - The user either creates or reuses `VersionUpgradeSchedule` object (see below). This part delivers the
+     "when?" part of automatic upgrades configuration.
+   - Then, the user creates a `VersionUpgradeConfig` object that is providing only the following information
```
ljakimczuk, line 91: Now when I think of it, I don't really like the design which opens up a way for odd configuration, which is seen as wrong by all of us, I guess, from the very beginning. I am talking about VUC CR which aggregates a single ImagePolicy CR, pointing to a single application, and multiple targets, being App CRs and HelmRelease CRs, potentially not even instances of the same app.

The assumption behind VUC seems to be, these multiple targets user selects are expected to be instances of the same app and that the ImagePolicy CR he references applies to this single app too. And I guess some assumptions are okay, car producers too rather don't assume someone will jump off a cliff in their cars, but then someone also builds up these protective barriers on the roads, so maybe we should have some barriers too 😅 

Like for example, currently I can create a VUC CR that targets NGINX, Kong and Trivy apps and I can reference ImagePolicy CR for Hello World app. It does not make sense. It is not that this configuration will do much harm, if Hello World app versions set does not contain any common versions with the NGINX, Kong or Trivy versions, in such case it will only break these apps-related CRs, worse if some versions are the same as @uvegla pointed out.

Also what feels odd is, I have targets, which are App CRs and HelmRelease CRs, both pointing to a repository, a Catalog CR or HelmRepository CR, and yet I have to point to ImagePolicy CR, which points to some repository again, which is in fact source of the risk. Maybe we should adopt Flux tactics from handling HelmChart CRs and create ImageRepository and ImagePolicy CRs implicitly for discovered targets?

I mean, let's forget about referring an ImagePolicy CR in the VUC CR. Say I have targets, and for each target I can apply some tags-related policy, like for example: 

```yaml
...
  targets:
    - kind: App
      field: spec.version
      name: cluster3-trivy
      policy:
        semver:
          range: `>=1.0.0`
    - kind: HelmRelease
      field: spec.chart.spec.version
      name: cluster1-trivy
      policy:
        alphabetical:
          order: asc
    - kind: OCIRepository
      field: spec.ref.semver
      name: trivy
      policy:
        semver:
          range: `^1.x-0`
    - kind: HelmRelease
      field: spec.chart.spec.version
      labelSelector:
        app: trivy
        stage: production
        vucUpdate: true
...
```

And then the VUO reads such targets and creates necessary Image* CRs? It does not necessarily have to create multiple Image* CRs to not hammer the API server, but if user targeted more than one app, not in terms of CRs instances, then obviously more Image* CRs should get created, etc.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+     - how to find target Resources and where to put the version info in them
+
+### GitOps integration
+
+To make it possible to use `VUO` for objects coming from a GitOps repo, the target object's version cannot be
+kept in the GitOps repo anymore (in case it's used). Instead, the operators acting on the target CRs must
+accept the fact that they don't manage the configured version property, and that the automatically discovered
+version is placed there.
+
+Even though this approach limits (only a bit) the disaster recovery properties of the GitOps approach, it is
```
ljakimczuk, line 104: I wouldn't say it limits disaster recovery only a bit. I would say in some scenarios it may not be so obvious of how to recover properly, or extra harm can be done, I am talking specifically about cluster from-a-scratch re-creation scenario, so the one not involving ETCD backup.

In such case, if we have some Flux source with versionless HelmRelease CRs and re-creating a cluster, so the source is applied fresh, then there is no guarantee Flux will apply VUO-related CRs first, even if we could deploy them first, there is no guarantee VUO will pick up versionless HelmRelease CRs from the source before Flux does it, in which case they will get deployed in the latest version. Later VUO will I guess try to downgrade them, so it may lead to another problems.

Not that it is a common case, but mentioning it anyway.

version-upgrade-operator/README.md
```diff
@@ -0,0 +1,673 @@
...
+important to note that for the specific set of `VUO` CRs, `target CR` and the set of discovered tags, the
+calculated version is deterministic. We will also try to provide good reporting in the status field of `VUO`
+CRs, including info about what was discovered and what was updated.
+
+In case of GitOps deployments of App CRs, we have to change all the CRs so that Flux doesn't manage the
+version field of the App CR and doesn't fight with `version-upgrade-operator` to set it. This might also need
+some changes in `app-operator` and/or `app-admission-controller`.
+
+With `HelmReleases`, the problem is that when no version is provided, it will default to `latest`. It means,
```
ljakimczuk, line 113: I read this entire paragraph a bit like: "we make VUO to provide the default version you pointed out, but you know, the VUO may not be fast enough and your app may sometimes get deployed in the latest version, in which case it will have to be downgraded. Shame though there exists apps that don't handle downgrades well. If you are worried your app is one of these, you can always suspend it first and deploy it this way, so that you buy VUO time. But you know what, if you're Flux user best if you could simply refer to the Flux native tools" 😄 

If I was a user I would be like "what the heck is recommended way?". Can we maybe come up with one way of handling such cases? If we see the risk of VUO not being fast enough, can we maybe agree that the proper way of handling version management is suspending / resuming or maybe some other, better integrated with Flux way?

For example, Flux has its native mechanisms for updates done directly in the source repository. What it currently lacks are schedules. So maybe VUO instead of updating version directly according to schedule, it should suspend / resume ImageUpdateAutomation CRs according to user-provided schedule. So that VUO still decides of "when" something happens, but Flux does it its own native way. Maybe for Flux-heavy users the entire solution should be taken towards this direction?

version-upgrade-operator/README.md
ljakimczuk: From my side, when I think about it, this mechanism feels redundant imho. Here is why.

First thing is, the `defaultVersion` field has any meaning only at the App or HR CRs creation time, for it these CRs existed long before VUC has targeted them, both would have versions already set to something.

Second thing is, if I create App or HR CRs and I define some `defaultVersion`, then it means I want the app to be actually deployed even if "now" is outside the scheduled window. 

So if I really want the app to get deployed now with some default version, why the heck do I need another field, if I am breaking the "upgrade-in-scheduled-window" rule anyway? Cannot the VUO just recognize this app is its new target, what it knows from the inventory, and then apply normal tag policies, but escaping the "upgrade-in-scheduled-window" rule too? Why do we need another field for escaping the rule? Less fields to handle, especially having `defaultVersion` makes no sense after some initial point in time.

version-upgrade-operator/README.md
uvegla: > So maybe VUO instead of updating version directly according to schedule, it should suspend / resume ImageUpdateAutomation CRs according to user-provided schedule.

Hmm, thats actually a very interesting and fresh idea thats worth exploring imo.

version-upgrade-operator/README.md
ljakimczuk: The rough idea for corrective behaviour could be describe with these steps:
- controller detects current version of an app and saves it in the VUC status.
- VUC gets reconciled at specified intervals and part of the reconciliation is the corrective behaviour. If now is outside scheduled window and the app version mismatches the one saved in status, the controller corrects it.
- if now is inside scheduled window, there is no point in correction in general, I think, for the usual reconciliation and acting according to the policy is the actual correction then.

PR #121, opened 2025-06-13
Add an RFC for the runbook migration

pipo02mix:
After a SIG docs discussion, we have agreed we need to do something in relation to where to place our runbooks

### DIFF
--- a/runbooks-placement-strategy/README.md
+++ b/runbooks-placement-strategy/README.md
@@ -1,12 +1,12 @@
 ---
-creation_date: 2025-09-05
+creation_date: 2025-07-15
 issues:
 - https://github.com/giantswarm/roadmap/issues/2838
+last_review_date: 2025-09-05
 owners:
 - https://github.com/orgs/giantswarm/teams/team-planeteers
 state: approved
-last_review_date: 2025-09-05
-summary: This RFC addresses the strategic placement of operational runbooks to improve accessibility, maintainability, and security while ensuring consistent structure and content sanitization. Part of the work is to rename the "Ops recipes" to "Runbooks" for clarity and alignment with industry standards.
+summary: This RFC addresses the strategic placement of operational runbooks to improve accessibility, maintainability, and security while ensuring consistent structure and content sanitization. Part of the work involves renaming the "Ops recipes" to "Runbooks" for clarity and alignment with industry standards.
 ---
 
 # Strategic placement and standardization of operational runbooks
@@ -19,7 +19,7 @@ Our operational runbooks (currently referred to as "Ops recipes") are currently 
 2. **Lack of standardization**: Runbooks follow different formats and structures across locations
 3. **Security concerns**: Some runbooks may contain sensitive information that needs to be identified and sanitized
 4. **Discovery issues**: Engineers struggle to find the right runbook when they're distributed across different platforms
-5. **Maintenance overhead**: Managing runbooks in multiple locations increases maintenance burden
+5. **Maintenance overhead**: Managing runbooks in multiple locations increases the maintenance burden
 
 We need a unified decision for runbook placement that ensures:
 
@@ -34,14 +34,14 @@ We need a unified decision for runbook placement that ensures:
 
 ### Preferred solution
 
-Consolidate all runbooks in the intranet site with a (dedicated) section, following a standardized structure. Part of the work is to rename "Ops recipes" to "Runbooks" to align with industry terminology and improve clarity.
+Consolidate all runbooks in the intranet site with a (dedicated) section, following a standardized structure. Part of the work involves renaming "Ops recipes" to "Runbooks" to align with industry terminology and improve clarity.
 
 #### Key components:
 
 1. **Primary location**: Move all runbooks to a dedicated section in the intranet site (`docs/runbooks/`)
 2. **Content sanitization process**:
-   - In order to move the runbooks we need to audit the documents for sensitive information (credentials, internal URLs, customer-specific data)
-   - Create sanitized versions removing or abstracting sensitive content, moving the sensitive information to the customers' repositories.
+   - In order to move the runbooks, we need to audit the documents for sensitive information (credentials, internal URLs, customer-specific data)
+   - Create sanitized versions, removing or abstracting sensitive content, moving the sensitive information to the customers' repositories.
    - Establish guidelines for what constitutes sensitive information (this is already done)
 3. **Standardized metadata structure**: the runbooks should follow a structured format frontmatter, including:
    - **Title and description**
@@ -49,7 +49,7 @@ Consolidate all runbooks in the intranet site with a (dedicated) section, follow
    - **Last review date**
 4. **Migration strategy**:
    - Phase 1: Decide the placement of the runbooks
-   - Phase 2: Migrate all runbooks to the intranet adding ownership and last review date. Phase out the vintage runbooks.
+   - Phase 2: Migrate all runbooks to the intranet, adding ownership and last review date—phase out the vintage runbooks.
    - Phase 3: Ensure all alerts and notifications are updated to point to the new runbook locations
 5. **Governance model**:
    - Designate runbook owners for each recipe to ensure accountability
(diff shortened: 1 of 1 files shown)


runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,96 @@
...
+- Clear governance and maintenance processes
+
+### Decision maker
+
+Planeteers team.
+
+### Deadline
+
+30.07.2025
+
```
AndiDog, line 37:
```suggestion
```

We have the `owners` field for this, and the deadline isn't relevant in a few weeks, so I'd leave this out (and deadlines usually don't work in RFCs).

runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,96 @@
...
+### Who is affected / stakeholders
+
+- Platform Engineers (primary users of runbooks)
+- External community members (for public runbooks)
+
+### Preferred solution
+
+**Consolidate all runbooks in the public handbook with a (dedicated) runbooks section, following a standardized structure.**
```
AndiDog, line 45: Maybe also mention the rename from "ops recipes" to the more well-known term "runbooks"?

Please place this in `summary:` as well, which is supposed to describe the action.

runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,96 @@
...
+   - In order to move the runbooks we need to audit the documents for sensitive information (credentials, internal URLs, customer-specific data)
+   - Create sanitized versions removing or abstracting sensitive content, moving the sensitive information to the customers' repositories.
+   - Establish guidelines for what constitutes sensitive information (this is already done)
+3. **Standardized structure**: Implement a consistent format for all runbooks including:
+   - **Title and description**
+   - **Prerequisites and required access**
+   - **Step-by-step procedures** with clear numbering
+   - **Expected outcomes and verification steps**
+   - **Troubleshooting section** (optional)
+   - **Related runbooks and references** (optional)
```
AndiDog, line 60: I don't think such major rework should be part of the migration, and I'm worried that a prescribed format just won't work because of the various styles existing runbooks were written in. Suggestion: remove this step.

We could however use a structured header (as in our RFCs, see [check script](https://github.com/giantswarm/rfc/blob/main/.github/rfc-format/check-rfc-format.py)) to ensure that newly-added ones have certain information as desired below (owner, etc.).

runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+**Benefits:**
+
+- **Improved discoverability**: Single source of truth for all operational procedures
+- **Enhanced collaboration**: Public access enables community contributions and feedback
+- **Better maintenance**: Centralized location simplifies updates and version control
+- **Knowledge sharing**: Public runbooks benefit the broader Giant Swarm community
+- **Consistency**: Standardized format improves usability and will make AI agents more effective in assisting with runbook-related queries
+
+### Alternative solutions
```
AndiDog, line 77: +1 for making everything public

mcharriere:
I think we should consider a few things in this RFC as a long term goal if we really want to go fully public.

1. Discoverability: just making the documents public doesn't guarantee that people would benefit from them. We should to define a strategy for "search and catalog".
2. Quality: we should add some quality guidelines to what constitutes a good runbook. Standardized format is of course a first step but more about the structure not the content.
3. Measure: we should measure engagement and access to those runbooks to know the scale of the impact.

runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,96 @@
...
+- Platform Engineers (primary users of runbooks)
+- External community members (for public runbooks)
+
+### Preferred solution
+
+**Consolidate all runbooks in the public handbook with a (dedicated) runbooks section, following a standardized structure.**
+
+#### Key components:
+
+1. **Primary location**: Move all runbooks to a dedicated section in the public handbook (`docs/runbooks/`)
```
hervenicol, line 49: Side note: I don't see anything in the public docs to encourage public contributions. I'd expect at least a link to the github repo if we want customers to contribute.

runbooks-placement-strategy/README.md
pipo02mix: I have changed to a standardized metadata.

pipo02mix:
>Discoverability: just making the documents public doesn't guarantee that people would benefit from them. We should to define a strategy for "search and catalog".

Here, my approach will be to utilize an AI agent to gather that information. Our current agent cannot scrape the intranet, so the recipes are missing.

> Quality: we should add some quality guidelines to what constitutes a good runbook. Standardized format is of course a first step but more about the structure not the content.

yeah I will push for that (we have a style guide but needs to be adapted) but this is not the goal of this RFC. Here we only discuss where to put them (now those are split) 

> Measure: we should measure engagement and access to those runbooks to know the scale of the impact.

yeah that will be considered but not in this RFC

mcharriere:
I think we should consider my points in this RFC (as a stated intention or long term goal). The RFC states that public access enables community contributions, but I think that's not enough. 
For the community to benefit from our public runbooks, they need to be able to find it when they search for an specific topic (discoverability), the content needs to be relevant and updated (quality). Measuring is the only way we can know if implementing this RFC was worth the effort.

I'm do not intent to extend a discussion around the details, but setting the path forward doesn't hurt. 

About "discovery":

> Here, my approach will be to utilize an AI agent to gather that information. Our current agent cannot scrape the intranet, so the recipes are missing.

I meant SEO and how our content is discovered by the community.

runbooks-placement-strategy/README.md
```diff
@@ -0,0 +1,85 @@
...
+### Preferred solution
+
+Consolidate all runbooks in the public handbook with a (dedicated) section, following a standardized structure. Part of the work is to rename "Ops recipes" to "Runbooks" to align with industry terminology and improve clarity.
+
+#### Key components:
+
+1. **Primary location**: Move all runbooks to a dedicated section in the public handbook (`docs/runbooks/`)
+2. **Content sanitization process**:
+   - In order to move the runbooks we need to audit the documents for sensitive information (credentials, internal URLs, customer-specific data)
+   - Create sanitized versions removing or abstracting sensitive content, moving the sensitive information to the customers' repositories.
```
hervenicol, line 44: How do we find the exceptions in the customer's repositories when oncall?

ie: "Please follow these steps to fix [whatever], except for [specific customer] where it works another way".
In this case today I write this specific info in the opsrecipe (especially since "another way" is often the reason why it breaks).
Should the runbook then link directly to the customer's repo instead of the public repos? This has lots of implications.

pipo02mix:
> I think we should consider my points in this RFC (as a stated intention or long term goal). The RFC states that public access enables community contributions, but I think that making them public is not enough to be beneficial for them.
The community needs to be able to find the runbooks when they search for an specific topic (discoverability), the content needs to be relevant and updated (quality) and measuring is the only way we can know if implementing this RFC was worth the effort.

Yes, I agree with you, but the primary target user will be the GS engineer, at least initially. It would be great if external users could find and read our runbook content in the future. In my mind, the main purpose is to make the content available to current customers (share links) and to make it accessible for AI training.

The metadata we are adding to the runbook will help to improve quality (it is the same approach as we are doing with regular docs), and having them under the handbook will let us know the most relevant content (because it already has metrics)

Screenshot 2025-07-11 at 17 01 47

runbooks-placement-strategy/README.md
pipo02mix: I have done a quick research across the current recipes. We typically don't store customer/installation information there. However, I see we have long and detailed content for [installations](https://github.com/giantswarm/installations/tree/master/leopard). So my proposal will be the search system (AI agent) will use both sources to give you more accurate response and you can always related to the links of the sources

pipo02mix: Yeah, because we don't have an AI agent for the intranet. Have you tried the AI agent in the general docs? IMO works better than old way with ElasticSearch

pipo02mix: Not at this moment because we only have the public one. But if we introduce it in the handbook it will be role-aware.

PR #122, opened 2025-06-27
Added initial revision of wc-patches-via-flux

AverageMarcus:


njuettner:
Two things here:

If the GitRepository fetch fails (e.g., GitHub outage), does WC reconciliation block?

I know with the `ClusterResourceSet` we would have this challenge to it cleanup properly and IIRC the docs are also saying it doesn't deal with the cleanup. 

What’s the cleanup story here and how would a rollback strategy work? Just some thoughts, I think we need to try it out 🙂

yulianedyalkova:
Closing, as agreed in product sync.

PR #135, opened 2026-02-04
RFC: Securely using Claude Code CLI with 1Password secrets

lyind:
## Summary

This RFC documents a secure approach for using Claude Code CLI and IDE plugins with secrets stored in 1Password on Linux.

Accompanying Claude skill: https://github.com/giantswarm/claude-code/pull/20

Key components:
- Dedicated 1Password vault with minimal-privilege service account
- Bubblewrap (`bwrap`) sandboxing for filesystem isolation
- `op://` secret references with `op run --` for just-in-time secret injection

## Test plan

- [ ] Review security properties and sandboxing configuration
- [ ] Validate instructions work on a fresh Linux setup
- [ ] Consider team-wide vault vs individual vaults

🤖 Generated with [Claude Code](https://claude.com/claude-code)   <-- already sandboxed at the time

### DIFF
--- a/claude-code-1password-secrets/README.md
+++ b/claude-code-1password-secrets/README.md
@@ -49,24 +49,38 @@ Create a script (e.g., `~/bin/run-claude.sh`) that:
 - Launches Claude Code inside a bubblewrap sandbox
 - Passes secrets as `op://` URIs (not raw values)
 
-```bash
-#!/bin/sh
+```sh
+#!/usr/bin/env sh
+set -e
 PROJECT_DIR="$(realpath .)"
-if [ -s "$1" ]; then
+if [ -d "$1" ]
+then
     PROJECT_DIR="$(realpath "$1")"
 fi
 
-# Safety check: prevent running in home or root directories
-if [ "$PROJECT_DIR" = "$(realpath "$HOME")" ] || [ "$PROJECT_DIR" = "$(realpath /)" ]; then
-    echo "Error: don't give Claude access to high-level directories! Run it somewhere else." 1>&2
+if [ "$PROJECT_DIR" = "$(realpath "$HOME")" -o "$PROJECT_DIR" = "$(realpath /)" ]
+then
+    echo "Error: don't give claude access to high-level directories! Run it somewhere else." >&2
     exit 1
 fi
 
-# Configuration
-vault="ai-jonas"  # Change to your vault name
+# GH_TOKEN and similar below contain only paths into the 1Password vault assigned to the service account
+vault=ai-jonas
 account_token="$(op item get --vault 'Employee' "service-account-token-$vault" --reveal --fields credential)"
 
-bwrap \
+temp_passwd="$(mktemp)"
+trap 'rm -f "$temp_passwd"' EXIT
+getent passwd "$(id -u)" >> "$temp_passwd"
+
+# Ensure bind directories exist
+for dir in .m2 .gradle go .npm .java .cache/pip .cache/helm .cache/go .cache/go-build .local/share/claude; do
+    mkdir -p "$HOME/$dir"
+done
+
+exec bwrap \
+    --unshare-user-try \
+    --uid $(id -u) \
+    --gid $(id -g) \
     --ro-bind /usr /usr \
     --ro-bind /lib /lib \
     --ro-bind /lib64 /lib64 \
@@ -74,7 +88,8 @@ bwrap \
     --ro-bind /etc/resolv.conf /etc/resolv.conf \
     --ro-bind /etc/hosts /etc/hosts \
     --ro-bind /etc/ssl /etc/ssl \
-    --ro-bind /etc/passwd /etc/passwd \
+    --ro-bind /etc/alternatives /etc/alternatives \
+    --ro-bind "$temp_passwd" /etc/passwd \
     --ro-bind /etc/group /etc/group \
     --ro-bind "$HOME/.gitconfig" "$HOME/.gitconfig" \
     --ro-bind "$HOME/.1password" "$HOME/.1password" \
@@ -97,10 +112,13 @@ bwrap \
     --bind "$HOME/.cache/helm" "$HOME/.cache/helm" \
     --bind "$HOME/.cache/go" "$HOME/.cache/go" \
     --bind "$HOME/.cache/go-build" "$HOME/.cache/go-build" \
+    --bind "$SSH_AUTH_SOCK" "$SSH_AUTH_SOCK" \
     --setenv "OP_SERVICE_ACCOUNT_TOKEN" "$account_token" \
     --setenv "GH_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
     --setenv "GITHUB_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
     --setenv "OPSCTL_GITHUB_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
+    --setenv "SSH_AUTH_SOCK" "$SSH_AUTH_SOCK" \
+    --setenv "SSH_ASKPASS" "$SSH_ASKPASS" \
     --tmpfs /tmp \
     --proc /proc \
     --dev /dev \
@@ -192,6 +210,8 @@ Modify the `--ro-bind` (read-only) and `--bind` (read-write) lines to match your
 - Should we provide a centralized, team-managed vault instead of individual `ai-$NAME` vaults?
 - How should this integrate with CI/CD pipelines that also use Claude Code?
 - Could the `run-claude.sh` script be improved, e.g. to make it more compatible or protect the service account token better?
+- Mac support? Use docker or another sandboxing app?
+- Enterprise monitoring/logging?
 
 ## References
 
--- a/claude-code-1password-secrets/README.md
+++ b/claude-code-1password-secrets/README.md
@@ -1,0 +1,201 @@
+---
+creation_date: 2025-02-04
+issues: []
+owners:
+- https://github.com/orgs/giantswarm/teams/team-up
+state: review
+summary: A guide for securely using Claude Code CLI and IDE plugins on Linux by storing secrets in 1Password and running the tool in a sandboxed environment using bubblewrap.
+---
+
+# Securely Using Claude Code CLI with 1Password Secrets on Linux
+
+## Problem Statement
+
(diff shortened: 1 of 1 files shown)


lyind:
NOTE: It might make sense for 1Password to roll their own "AI sandbox with perfected secret injection" product.

It could even be just what the script does, but rolled into `op` itself.

puja108:
1. /sandbox feature of Claude Code automatically uses whatever sandboxing feature is available (native on Mac, bubblewrap on Linux,...). I wouldn't reinvent the wheel here. The CC teams iterates fast on these things and is on top of latest developments at least for local setups.
2. I do think creating a 1password vault/user for your sandboxed claude code can make sense in some cases, just to be able to give it e.g. github token or other auth it might need in a controlled way.

also re:
>  I have not succeeded in getting op (1Password CLI) to run inside the sandbox without the service account token, i.e. connected to the signed-in 1Password desktop app.

That should be exactly the expectation, or not? When sandboxed it should NOT be able to connect to the host 1password, cause that is your personal login. It should only access its own vault. The most secure automation/claw setups I've seen recently all create special users for the AI that they then use to share passwords with, like I would share a password with my kid.

lyind:
> That should be exactly the expectation, or not? When sandboxed it should NOT be able to connect to the host 1password

Yes, it is the expectation. I still tried it out to maybe offer a fallback for users who don't setup a 1Password service account/special vault. However, it's not feasible.

puja108:
> for users who don't setup a 1Password service account/special vault

but that but be quite insecure then, or am I missing something?

lyind:
Still more secure than using Claude Code without the password manager skill. Less secure than "sandboxed + 1Password service account". Without the 1Password service account, Claude could, if manipulated to do so, read all your secrets as it would on the filesystem.

stone-z:
I still think it's highly unlikely that the humans are consistent enough about keeping secrets organized into meaningfully separate contexts for this to ultimately be a significant security benefit. We really would need something more dynamic and auditable

lyind:
Is there a way to automate creation of those vaults?

A simple script run by the user or even in the sandbox script (see RFC) could perform the necessary setup, provided 1Password CLI is installed.

claude-code-1password-secrets/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+# Securely Using Claude Code CLI with 1Password Secrets on Linux
+
+## Problem Statement
+
+Claude Code CLI (and IDE plugins) can execute arbitrary commands, including those that require authentication tokens (e.g., `gh`, `devctl`, `opsctl`). Storing these secrets as plain environment variables or in dotfiles poses security risks:
+
+1. **Secret exposure**: Claude Code could accidentally leak secrets in logs, error messages, or when suggesting commands.
+2. **Overly broad access**: The tool might access files or directories outside the intended project scope.
+3. **Credential persistence**: Secrets stored in shell profiles or environment files remain accessible even when not needed.
```
TheoBrigitte, line 18: I fully agree on this, but I also think this problem goes beyond Claude/AI use cases. Having secrets populated into environment variables exposes them to any CLI tooling.

claude-code-1password-secrets/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+Make the script executable:
+
+```bash
+chmod +x ~/bin/run-claude.sh
+```
+
+#### OPTION B) Trust and use Claude's builtin sandbox support
+
+Install a sandbox provider (e.g. Bubblewrap mentioned above), start Claude Code (*without sandbox/normally*), enter `/sandbox` in Claude to perform initial configuration.
```
TheoBrigitte, line 153: I am not in favor of trusting Claude builtin sandbox, I think we should trust bubblewrap which is the sandbox tool. When we launch Claude through bubblewrap we have the guarantee that Claude is then run inside a sandbox.

claude-code-1password-secrets/README.md
```diff
@@ -0,0 +1,229 @@
...
+    --bind "$HOME/.cache/helm" "$HOME/.cache/helm" \
+    --bind "$HOME/.cache/go" "$HOME/.cache/go" \
+    --bind "$HOME/.cache/go-build" "$HOME/.cache/go-build" \
+    --bind "$SSH_AUTH_SOCK" "$SSH_AUTH_SOCK" \
+    --setenv "OP_SERVICE_ACCOUNT_TOKEN" "$account_token" \
+    --setenv "GH_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
+    --setenv "GITHUB_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
+    --setenv "OPSCTL_GITHUB_TOKEN" "op://$vault/OPSCTL_GITHUB_TOKEN/password" \
+    --setenv "SSH_AUTH_SOCK" "$SSH_AUTH_SOCK" \
+    --setenv "SSH_ASKPASS" "$SSH_ASKPASS" \
```
TheoBrigitte, line 129: IMO using `--setenv` does not help with accidentally leak secrets, Claude would then have access to those with a potential to leak them.

claude-code-1password-secrets/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+```bash
+chmod +x ~/bin/run-claude.sh
+```
+
+#### OPTION B) Trust and use Claude's builtin sandbox support
+
+Install a sandbox provider (e.g. Bubblewrap mentioned above), start Claude Code (*without sandbox/normally*), enter `/sandbox` in Claude to perform initial configuration.
+
+### Step 4: Configure Claude Code to Use `op run`
```
TheoBrigitte, line 155: Using `op run` seems the most secure option IMO, but I would rather have it configured via a `PreToolUse` hook, which would then rewrite the invoked command to prefix it with `op run`.

I am also a bit concerned about performances when running **every** tool call through `op run` which is calling 1Password over network for each command execution.

lyind: The linked 1Password skill PR (`giantswarm/claude-code`) addresses this, by educating Claude about when exactly to prefix with `op run --`.

claude-code-1password-secrets/README.md
lyind: The variables set using `--setenv` are placeholder only. They are only populated on getenv() calls from processes run under `op run --` launched by the sandboxed Claude Code CLI instance.

I agree that making secrets directly available to the sandboxed Claude Code CLI via `--setenv` is **unsafe** and probably ran by many developers at this moment.

The example above is as safe as it gets, as it relies on `op run --` to uncloak the actual secrets and the 1Password skill to limit when `op run --` is being used.

claude-code-1password-secrets/README.md
lyind: That is why this RFC suggests telling Claude, via a Skill, to call commands that need access to a secret through `op run --`, which transparently uncloaks available secrets in environment variables, when the wrapped application process calls `getenv()`.

claude-code-1password-secrets/README.md
TheoBrigitte: I don't think this belongs in a skill, it's rather specific setup/environment definition which Claude should ideally not need to know about. Claude should just be able to run commands without any overhead. I would like to avoid Claude having to additionally "think" about how it should run a command, this bring some additional complexity for the agent itself, which we could avoid IMO.

lyind: It could be put into a hook, as you suggested.

The hook would have to detect if the command to be run is one that requires authentication (e.g. `gh`).

However, for some uses of secrets (e.g. in `curl` commands), it's not enough to just check the command and one needs the context and some brains to decide if secrets should be provided, or not. Only the agent itself can meaningfully make that decision: thus the Skill.

claude-code-1password-secrets/README.md
puja108: `/sandbox` does use bubblewrap (and socat) on Linux (and Windows WSL2) and native Mac sandboxing on Macs

lyind: Yes, that's why I listed it as an option. Though, as Theo said, it seems self-defeating to trust the thing to be sandboxed to set up the sandbox.

Claude's developers may not see it that way, but I do want to sandbox *all components of Claude itself*, not merely the stuff it may pick up on the Internet and run.

lyind: Plus, keeping the sandboxing technique separate keeps it useful to reign in other applications, too.

puja108: I do understand that notion, but i kinda trust a public-facing Anthropic team that will be harshly scrutinized for any security-faux-pas by the whole internet a bit more on keeping uptodate with sandboxing tech than us. If the security or AI team however wants to support and maintain our own sandboxing I'm fine.

lyind: We don't have to maintain "our own sandboxing". The sandboxing technologies are available for use via established and well-known command line tools. What's required on top is configuration/command line arguments that are otherwise controlled by the TypeScript/NodeJS application [`srt (Claude's sandbox runtime)`](https://github.com/anthropic-experimental/sandbox-runtime/tree/main).

Furthermore, we'd like to keep as many things as possible independent from the specific AI vendor (Anthropic) as that vendor may go away or we may opt to use another provider in future.

The one thing `srt` does "better" than just launching `bwrap` is adding the network filtering/management in a user friendly way. That and the Mac users are why we should list it as a valid alternative.

puja108: And the whole config and arguments and also keeping uptodate on newer sandboxing tooling and alternatives (or switching tools) is what we will need to maintain, we also do not maintain any of our OSS apps in the traditional sense, they are upstream components, but still we have considerable work in each team "maintaining" these apps.

Sure, we might move away from Anthropic at some point, and until then we and the whole ecosystem will be a few steps further in the whole sandboxing journey (which is moving extremely fast due to the pressure openclaw put on everyone), so then we need to re-evaluate, but until then I would rely on things that do not add extra burden of ownership in our very thin org.

That said, this is just my opinion, and as long as the option of using in-built tech is mentioned I'm fine and do not block anything here.

lyind: We will make sure to always explicitly mention the Claude provided sandboxing mechanism as a quick default path to get a workable isolated setup.

stone-z:
> Is there a way to automate creation of those vaults?

My thinking was something more dynamic from the identity provider side. 1Password's XAM [might also nicely support](https://1password.com/blog/extended-access-management-for-ai-agents) what you're trying to do

lyind:
We haven't had a chance to evaluate 1Password's XAM and it doesn't look like we will deprecate "good old authentication schemes", for personal workstation use, anytime soon.

PR #144, opened 2026-04-02
RFC: workload identity as a platform feature

stone-z:


workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+7. No customer or entity may **impersonate or influence** another customer's identities, nor use the identities available to them to access another customer's system.
+
+### Assumptions
+
+- Every management cluster (MC) is per-customer, per-region. Customers have full access to their infrastructure. Giant Swarm operates within, but does not own it.
+- Workload clusters (WCs) are managed by the MC and run both customer and Giant Swarm workloads.
+
+### Challenges
+
+- **Short-lived credentials require nearby issuance.** With certificate TTLs of ~1 hour, the issuing authority must be continuously reachable.
```
QuentinBisson: Maybe this is something that could be configurable per cluster for clusters that do not have proper network connection or airgapped systems? I would in that case not make it a challenge :)

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+| 7 — No cross-customer impersonation | Separate root CAs per customer. Only GS leaf credentials in customer infrastructure — no company CA keys. |
+
+## Open questions
+
+1. **Semantic vs structural IDs.** Encapsulating customer and architecture information in an ID makes it easy to understand, but imposes rigidity, leaks information, makes and long-term assumptions that won't hold. Instead of `v1/giantswarm/platform/falco/v1-2-3/customer-x/eu-west-1/mc-name/wc-name`, consider `v1/giantswarm/platform/falco/v1-2-3`. The second more semantically identifies the application independently of the architecture, but additional work must be done to support cluster/customer/regions.
+
+2. **What's in a name?** Especially if using structural names, what are the dimensions that must be included in the path?
+
+3. **SPIFFE ID path segment ordering.** Should customer precede region (`/<customer>/<region>/...`) or follow it (`/<region>/<customer>/...`)? Zach is in favor of customer first, then region, if using structural IDs.
```
QuentinBisson, line 221: As the RFC notes, topology changes (cluster renames, region moves, MC migrations) invalidate all existing authorization policies. And it leaks internal architecture to anyone who can read the cert.

How about we instead use an hybrid approach `spiffe://wid.giantswarm.com/v1/app/platform/falco/v1-2-3/loc/customer-x/eu-west-1/cluster-abc` where the path has two explicit, versioned segments:

`/app/...` stable semantic identity (what it is)
`/loc/...` structural location (where it runs)

This lets relying parties write policies at any granularity:
- UC1: match on `/app/platform/falco/*` — any Falco, anywhere
- UC2: match on `/app/platform/falco/*/loc/customer-x/eu-west-1/*` — Falco in EU for customer X
UC4: match on `/app/platform/prometheus/*/loc/*/*/cluster-abc` — Prometheus in this specific cluster

The /app/ and /loc/ separators also make it explicit which part of the ID is "what" vs "where" — and in theory, you could evolve each independently (e.g. bump to /app/v2/... or /loc/v2/...) without breaking the other.

The tradeoff is that location info still leaks cluster names/regions to anyone reading certs, but that seems unavoidable if you want scoped authorization to actually work.

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+4. **Dual identities for Giant Swarm workloads.** Some workloads need both a global company identity (for external services) and a customer-local identity (for resources that must survive offboarding). It is possible to issue multiple identities to a workload, which must be aware of which to use. Is that useful? Should workloads receive SVIDs from both planes? What are the complexity and security implications?
+
+5. **How many customer SPIREs?** In theory, we could operate customers with only WC or MC SPIREs, and not require both. The MC SPIRE could issue for all WC agents, or WC SPIREs could nest directly under the global root. The most resilient would be to use both.
```
QuentinBisson, line 225: I think having one per WC would definitely reduce the risks similar to zot being deployed on WCs but maybe we could provide this as an opt-in?

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+8. **Support for external --> any customer authentication.** Use case 7 describes the possibility for an external service to authenticate "any instance of one|any application belonging to any Giant Swarm customer". This is currently very difficult to implement. The correct solution would involve identities signed by both Giant Swarm and the customer. This is not yet supported. So, how important is this use case?
+
+9. **Customer SPIRE root placement.** The customer root needs to live somewhere highly available. Options include a dedicated WC in the customer's primary region, co-location in the first MC, a lightweight non-K8s host, or customer-provided infrastructure. The dedicated WC is the current recommendation, but introduces a single-region placement for a cross-region dependency.
+
+10. **DNS and OIDC endpoint ownership model.** With a single trust domain per customer, there is one DNS name to manage (e.g. `wid.acme.example.com`). Should customers own this DNS and delegate to Giant Swarm during the relationship, or vice versa? Zach suggests customers own the DNS entry and point it back at their own infrastructure after offboarding.
```
QuentinBisson, line 235: I think it would be easier for offboarding if they own it right?

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+2. **What's in a name?** Especially if using structural names, what are the dimensions that must be included in the path?
+
+3. **SPIFFE ID path segment ordering.** Should customer precede region (`/<customer>/<region>/...`) or follow it (`/<region>/<customer>/...`)? Zach is in favor of customer first, then region, if using structural IDs.
+
+4. **Dual identities for Giant Swarm workloads.** Some workloads need both a global company identity (for external services) and a customer-local identity (for resources that must survive offboarding). It is possible to issue multiple identities to a workload, which must be aware of which to use. Is that useful? Should workloads receive SVIDs from both planes? What are the complexity and security implications?
```
QuentinBisson, line 223: I think it would be interesting to know how many of those use cases we have for this right. Can this be fixed by adding a sidecar that presents the correct svid depending on what workload is coming (based on the other workload identity)?

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+5. **How many customer SPIREs?** In theory, we could operate customers with only WC or MC SPIREs, and not require both. The MC SPIRE could issue for all WC agents, or WC SPIREs could nest directly under the global root. The most resilient would be to use both.
+
+6. **How/where does Giant Swarm run our regional SPIRE?** We can't (yet) run nested SPIRE servers within customer MCs because they would have access to key material that allows lateral movement to other customer trust domains. To keep "per-customer, per-region" failure domains, we would need to run a SPIRE server "next to" each MC, but in a place customers can't access. Where is that? (there is a future feature called "named path" support which may allow us to safely run it on the customer MC, but it's already years in the making).
```
QuentinBisson: If each customer MC runs its own GS regional SPIRE instance (not shared across customers), a customer extracting the intermediate CA can only forge GS identities scoped to their own path so the blast radius it limited to the customer (having access to the MC) sabotaging his own installation right?

It is the same trust model as the teleport agent running on the MC though is it not?

workload-identity/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+7. **Air gap?** I think it makes sense to assume we need this eventually. Are there airgap-specific use cases?
+
+8. **Support for external --> any customer authentication.** Use case 7 describes the possibility for an external service to authenticate "any instance of one|any application belonging to any Giant Swarm customer". This is currently very difficult to implement. The correct solution would involve identities signed by both Giant Swarm and the customer. This is not yet supported. So, how important is this use case?
+
+9. **Customer SPIRE root placement.** The customer root needs to live somewhere highly available. Options include a dedicated WC in the customer's primary region, co-location in the first MC, a lightweight non-K8s host, or customer-provided infrastructure. The dedicated WC is the current recommendation, but introduces a single-region placement for a cross-region dependency.
```
QuentinBisson, line 233: There are tradeoffs to having it running in a customer WC though

QuentinBisson:
Do you think it would make sense to rationalize all the workload identity mechanism we use today (IRSA, the one on Azure, Teleport and maybe others) in this RFC or is it fine to check this out later? There might be some overlap at some point

workload-identity/README.md
stone-z: That's true, it isn't really a challenge as much as a constraint. This refers to the duration of each individual workload cert. These _should_ be short-lived, but we meet that goal by having a local (WC or MC) SPIRE instance. I'll update the doc

workload-identity/README.md
stone-z: My tendency is to make it optional, but opt-out. To support air-gapped clusters, we would need to have WC SPIRE anyway. So I would make MC + WC the default, since it is most resilient and ensures we're good at running WC SPIREs, but let customers choose to consolidate to using only MC SPIREs if they want to reduce the footprint and they accept the risk. I don't _think_ it would be difficult to reconfigure a running setup, but I'd leave that to after the initial implementation.

workload-identity/README.md
stone-z: Definitely, and that's closer to what we already do with their domains in other contexts. It just means they have limited identity options until they set it up, and if they mess up DNS the blast radius is larger.

workload-identity/README.md
stone-z: That actually isn't an issue for the relying party. The issue here is when a given client (say, alloy) has multiple identities/certs available, it needs to support giving a different one to different services (like, "hello Grafana Cloud, I'm a Giant Swarm Alloy!" vs "hello Customer X Victoria Metrics, I'm a Customer X Alloy!"). IMO the alloy example is a great use case for multiple identities, but client support for it is relatively unknown to me.

workload-identity/README.md
stone-z: My suggestion is that customers don't run or have any access to GS regional SPIREs.

> a customer extracting the intermediate CA can only forge GS identities scoped to their own path

If we run a GS SPIRE in a customer MC, _currently_ the intermediate cert could be used for signing any path, not limited to the customer's own. Support for only signing certain paths is part of the spec, but not yet implemented.

> It is the same trust model as the teleport agent running on the MC though is it not?

They are different mechanisms and risks. During MC creation, we generate a token in the Teleport cluster which gets passed to the new MC. This token is used by teleport-operator to issue fresh tokens for all the attached WCs. But the initial MC-level token is still subject to Teleport RBAC and controlled centrally in the Teleport cluster. There are still some risk tradeoffs (see [here](https://github.com/giantswarm/giantswarm/issues/35198) if you're curious), but if a customer were to abuse their Teleport token it would be a much smaller risk than abusing an intermediate signing cert.

workload-identity/README.md
stone-z: Absolutely :100:

I think the best case is the customer already has a SPIRE root and we just plug into that. However, the vast majority won't have one, so from a customer experience perspective, I think us providing a root in a special WC is the second best overall for them and us.

WDYT?

stone-z:
> would make sense to rationalize all the workload identity mechanism we use today (IRSA, the one on Azure, Teleport and maybe others)

I hadn't planned on trying to survey all of the other methods specifically in this, _however_ I would love to hear from users of those other methods about their usage.

It would be great to replace the provider-specific identity mechanisms with a single common one for all the use cases where that's possible and makes sense. We probably still end up needing some provider-specific stuff for early infrastructure lifecycle stuff, but that's fine.

Teleport is also a good callout. In theory, an alternative to this RFC would be to invest in building identity around Teleport Enterprise. However, that is a lot of lock-in, plus additional cost. But it would be an option.

workload-identity/README.md
stone-z: Yes, including semantics and then structure in the ID is currently my working idea. But it has tradeoffs.
- wildcards are not universally supported (prefix would have more support, but still not 100%)
- examining any ID discloses architecture information, and we'd need to anonymize customer in the IDs, which makes it less human-friendly

Plus some others. However, so far, my suggestion would still be to use that rough structure (semantic identity, then structural qualifiers) because I do think it's very important to be able to write policies that target both customers and regions, and I don't see a way to offer that semi-universally without having it encoded in the ID.

I would love if a reviewer did have an idea for hiding the structure information from the ID but still making it publicly available for policy writing.

workload-identity/README.md
QuentinBisson: Sure that makes a lot of sense

workload-identity/README.md
QuentinBisson: Alright that makes sense. But then I do wonder if the initial implementation could not start with this approach and then try to find a more secure solution later on. it could also change mid implementation as well unless named path is something that could be pushed upstream? Do you have infos on that feature?

workload-identity/README.md
QuentinBisson: You're probably right but I would hope cloud providers could provide something for this and then this would be an onprem only thing :D

QuentinBisson:
The direction of the RFC is really good. There are still some opened questions but it should definitely not block the implementation of an MVP on giantswarm clusters to learn more about spiffe/spire and try to find solutions to the opened questions. Nice work @stone-z

workload-identity/README.md
stone-z: The [X509-SVID standard](https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md#42-name-constraints) says that the issue is most libraries doing the actual validation don't support the field, even though it's part of the X509 spec. If SPIRE issued a cert with a `permittedSubtrees` field (that's what it's called in the spec), and a library used by some server tried to check it, it would just fail. So in order to use that feature, many client libraries need to be updated to support it (and the SPIFFE folks are apparently already working on that). In theory it could show up any time, but I don't think we can influence the timeline.

workload-identity/README.md
stone-z: Interesting, I was thinking of going the other way and making SPIRE the default everywhere, and only using provider-specific options in cases where there was a clear indication for it. I'll add building default orchestration of each provider's thing as an alternative

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+- **Trust bundle:** The set of root CA public certificates that a verifier uses to determine whether a presented certificate is trustworthy.
+- **Trust domain:** The organizational boundary within which a single CA hierarchy issues identities. Two workloads in the same trust domain can authenticate each other directly. Workloads in different trust domains require trust federation.
+
+## Use cases for validation
+
+![Use Cases](../@rfc/workload-identity/003use-cases.png)
+
+The following use cases drive the design. Each combines a relying party (who validates), an identity holder (who proves identity), and a scope.
+
+1. **External service authenticates any Giant Swarm application globally.** An external API trusts "any Giant Swarm net-exporter" regardless of which customer cluster it runs in.
```
pipo02mix: I would use letter instead of number, beause first time I show the diagram I thought it was a sequence of actions

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+2. **External service authenticates Giant Swarm applications scoped to a customer or region.** A customer's S3 bucket trusts "Giant Swarm Falco in Customer X's clusters in Europe only."
```
pipo02mix: Is that possible? How can an S3 bucket verify a container credential? I am interested to learn about it

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+### Assumptions
+
+- Every management cluster (MC) is per-customer, per-region. Customers have full access to their infrastructure. Giant Swarm operates within, but does not own it.
+- Workload clusters (WCs) are managed by the MC and run both customer and Giant Swarm workloads.
+
+### Challenges
+
+- **External validation must also be scoped and available.** Discovery endpoints / trust bundles for validation must be reliably reachable by external verifiers (AWS STS, Azure Entra ID).
+- **Identity path design leaks information.** Paths encoding internal architecture (application names, customer names, regions) are visible to relying parties. The naming scheme must balance expressiveness with information exposure.
+- **Global vs. local identity.** Some Giant Swarm workloads need a global identity (e.g. Mimir reporting to Grafana Cloud), while others need a customer-local identity (e.g. Alloy scraping endpoints that must survive customer offboarding). A workload may need both, and this need may change during its lifetime.
```
pipo02mix: Why `local` should be different from `global`? I mean, can we use global to probe identity in local envs? If it is the case, we don't need that distinction, and every identity could be globally validated, and we have the benefit that can work in both cases

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+1. **Giant Swarm Platform identity plane** — One trust domain (`spiffe://wid.giantswarm.com`) for all Giant Swarm platform workloads across all regions and customers. Controlled entirely by Giant Swarm.
+2. **Customer identity plane** — One trust domain per customer (`spiffe://wid.customer-x.com`), running in the customer's infrastructure with customer-held CA keys.
+
+Federation between planes enables cross-domain authentication (use cases 3–6).
+
+### Company plane topology
+
+- A **central SPIRE root** hosts the OIDC discovery provider and public federation bundle endpoint.
+- **Regional SPIRE servers** receive intermediate CAs with 24-72h TTLs to mitigate temporary central outages.
+- In each customer cluster, a **company SPIRE agent** (not a nested server) connects to the regional server. Agents hold only leaf SVIDs — no CA key material, so a compromised customer cluster cannot forge identities for Giant Swarm or other customers.
```
pipo02mix, line 145: It means the agent needs to pull identities frequently so new fresh container can authN using the new identity? I have no idea how this part works (agent <-> container <-> API)

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Customer plane topology
+
+- A **customer SPIRE root** runs in a dedicated WC and serves as the single root of trust for the customer's trust domain. Root CA keys are stored in the customer's KMS. It hosts the customer's OIDC discovery provider and federation bundle endpoint.
```
pipo02mix, line 149: Why OIDC discovery provider and federation bundle has to run in a WC instead of MC?

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Credential lifetimes
+
+| Credential | TTL | Rationale |
+|---|---|---|
+| Root CA (company or customer) | 5–10 years | Rotation requires re-establishing trust across all relying parties |
+| Regional / MC intermediate CA | 24–72 hours | Survives temporary root unavailability; same pattern for both planes |
+| WC intermediate CA | 12–24 hours | Shorter than MC intermediate; bounded by MC intermediate lifetime |
+| X.509-SVID (workload cert) | 1 hour | Short-lived; automatic rotation via SPIRE agent |
+| JWT-SVID | 5 minutes | JWTs are not revocable; minimize the replay window |
```
pipo02mix, line 199: Do we need to consider the effect of having hundreds of pods renewing this token on the API?

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+| 5 — New customer, no change | New trust domain + new agents + one federation entry. Nothing existing is modified. |
+| 6 — On/offboarding, no authz change | Policies supporting wildcards are very flexible. All policies (even without wildcards) work for same trust domain. Cross-domain trust (e.g. a customer service trusting Giant Swarm applications) would need to be reconfigured. |
+| 7 — No cross-customer impersonation | Separate root CAs per customer. Only GS leaf credentials in customer infrastructure, no company CA keys. |
+
+## Open questions
+
+1. **Semantic vs structural IDs.** Encapsulating customer and architecture information in an ID makes it easy to understand, but imposes rigidity, leaks information, makes and long-term assumptions that won't hold. Instead of `v1/giantswarm/platform/falco/v1-2-3/customer-x/eu-west-1/mc-name/wc-name`, consider `v1/giantswarm/platform/falco/v1-2-3`. The second more semantically identifies the application independently of the architecture, but additional work must be done to support cluster/customer/regions.
```
pipo02mix, line 217: Should we take that decision, or can each customer use a different approach?

workload-identity/README.md
pipo02mix: Why not use global identities always? That way, they can be validated in a local environment or globally for external services. Maybe I don't understand it, but I suppose the difference lies in the SVID format, am I right?

workload-identity/README.md
QuentinBisson: AWS trust policies are for this I think

workload-identity/README.md
QuentinBisson: If customers follow different approach then this will make it really hard to have shared/global policies. Also I think this is about Giant Swarm SVIDs and not customers :)

workload-identity/README.md
stone-z: The global vs local is only one example of a more generic dual identity question (also see Open question 4). I would expect the vast majority to be global (simply meaning the trust/CA bundle is public) identities, except perhaps for air-gapped clusters or a few unusual use cases.

The challenge still remains if a workload needs multiple global identities or multiple local ones, as in the alloy example. We end up in cases where we need to choose whether a workload is "ours" or "theirs" when it might serve use cases for both (or we need to run separate "ours" and "theirs" instances).

workload-identity/README.md
stone-z: AWS can be configured to recognize the public trust bundle as an OIDC identity provider ([example/tutorial here](https://spiffe.io/docs/latest/keyless/oidc-federation-aws/#part-3-configure-aws-components)). Once that is set up, SPIFFE ids can be bound to IAM Roles and assume other roles just like human users.

workload-identity/README.md
stone-z: Right, there is an agent in every cluster (MC and WC) that identifies each workload and contacts the (nested) SPIRE server API to request / renew the certificates.

workload-identity/README.md
stone-z: Let's distinguish between the bundle and the SPIRE server itself.

The trust bundle and OIDC discovery document are static, and not synchronous requests to the SPIRE server. We could (and likely will) serve these statically from e.g. S3 in order to serve them regionally. They update in the order of every few hours by default.

The root SPIRE will serve synchronous API requests from the nested SPIRE servers, and nested SPIRE servers will serve synchronous API requests from the agents.

The SPIRE root shouldn't run in an MC to reduce both the security and operations blast radii. If the cluster running the root SPIRE were to be compromised, the cryptographic material there could be used to impersonate any workload in the fleet or create entirely new ones. Plus, customers with multiple MCs would need network routes to the "central" MC. It also means the lifecycle of the root server is tied to that of an MC. With our normal backups, it's much easier to recover a deleted WC than it is a deleted MC.

workload-identity/README.md
stone-z: SPIRE is used behind the scenes in some very large systems, but it's definitely something to be aware of. Workloads themselves don't request certs; the SPIRE agents do, so as long as the agents are well-behaved, we shouldn't have issues with e.g. crashlooping pods. However, that is one reason why it's useful to have WC-scoped nested SPIREs.

workload-identity/README.md
stone-z: Right, this question is primarily about how to design our own SVIDs. We should have a recommendation, but the decision is ultimately up to each customer (and, if we are re-using a SPIRE root that already exists, the decision was already made). It would be nice if every customer chose the same format so we could more easily automate default policies, but I don't think we can assume that will be the case.

workload-identity/README.md
stone-z: Good idea, I've re-labeled the use cases alphabetically instead

workload-identity/README.md
teemow: I'm not yet sold on the idea to use dedicated WCs for management functionality. And I would like to discuss the pros and cons before we make such a decision.

Cons are:
* Even more costs and overhead for a GS installation 
* East West traffic between WCs
* Even more complex bootstrapping of installations for us. And more management overhead.

Maybe it is also good to discuss with customers?

teemow:
This overall looks well designed with a lot of considerations. I haven't had the time to look into this deeper and I don't know if I have the time in the near future. I only skimmed over it. This is not because it is not valuable for giantswarm and not because I do not respect all your work here. This is a complex topic. I'll not block this. Please go ahead if things have been discussed enough from your pov.

I do like a system that is independent of cloud providers and it would fit Giantswarm very well but I also think that the solutions coming from the cloud providers are the ones that most customers want to use. So in an ideal scenario we are cloud provider native as much as possible. We had similar solutions in the past and this always leads to a lot of discussions, more complexity for us and the customer and often less functionality or value as the open source solutions lack integration with the cloud provider compared to the providers solution.

Eg what would be the benefit for a customer to use spiffe/spire if they can simply use the workload identity feature in EKS? Can we convince customers that our spiffe/spire solution is worth the extra complexity and effort?

teemow:
Btw I am happy to connect you with amutable as they are working on the link between tpm, the OS and workload identities to make the whole chain verifiable. I anyway promised them that we'll give feedback as soon as they have things in place.

QuentinBisson:
That's a good question. We do need workload identity for us but can our spire connect with whatever customers want to us instead of the root spire? If yes, maybe we should focus on our implémentation and provide a customer spire when needed? This is must have onprem though

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+- **(D)** **Customer workload authenticates Giant Swarm applications locally.** A customer application allows "Giant Swarm Prometheus in this cluster" to scrape it.
+- **(E)** **Giant Swarm service authenticates a customer application broadly.** A Giant Swarm API trusts "any Customer X Muster."
+- **(F)** **Giant Swarm service authenticates specific customer application instances.** A Giant Swarm API trusts "Customer X's Jenkins in this cluster."
+
+Bonus: connections where both sides have federated trust in the other can optionally be secured with mTLS.
+
+### Discussion use cases:
+
+- **(G)** **External service authenticates any instance of any customer application.** An external service trusts "any Jenkins belonging to any Giant Swarm customer".
+Implementing (G) presents significant challenges with the current feature set of the proposed solution. Let's discuss if this is important.
```
puja108, line 70: this would currently mean with the current proposal, setting it up to trust each customer separately, right? which is not super bad IMHO

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+1. The failure domain of any identity component must be limited to a **single region**.
+2. The failure domain of any identity component must be limited to a **single customer**.
+3. Giant Swarm is itself a customer of the platform.
+4. If a customer leaves, their identity system must **remain operational** outside Giant Swarm.
+5. Onboarding a new customer must require **no changes** to existing customers.
+6. Customer on- or off-boarding must require **no changes** to pre-existing authorization policies.
+7. No customer or entity may **impersonate or influence** another customer's identities, nor use the identities available to them to access another customer's system.
+8. Certificate issuance and validation must be sufficiently available to support secure certificate usage.
+  a. Short-lived credentials (TTL of 5 minutes for JWTs) must be supported.
+  b. Implication: Short-lived credentials require a nearby issuance.
```
puja108, line 96: and HA in some sense of the word

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Assumptions
+
+- Every management cluster (MC) is per-customer, per-region. Customers have full access to their infrastructure. Giant Swarm operates within, but does not own it.
```
puja108, line 100: hybrid MCs are not really limited to a single region like with manufacturing customers where WCs are in other infras, but should not be a blocker AFAICS

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+- Workload clusters (WCs) are managed by the MC and run both customer and Giant Swarm workloads.
+
+### Challenges
+
+- **External validation must also be scoped and available.** Discovery endpoints / trust bundles for validation must be reliably reachable by external verifiers (AWS STS, Azure Entra ID).
+- **Identity path design leaks information.** Paths encoding internal architecture (application names, customer names, regions) are visible to relying parties. The naming scheme must balance expressiveness with information exposure.
```
puja108, line 106: this is only exposed on actual usage, right? or can it be queried from outside?

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+- **Global vs. local identity.** Some Giant Swarm workloads may need multiple identities for differeny use cases. For example, a workload may need a global identity (e.g. Mimir reporting to Grafana Cloud), while others need a customer-local identity (e.g. Alloy scraping endpoints that must survive customer offboarding). A workload may need both, and this need may change during its lifetime.
+- **Customer lifecycle edge cases.** Departure-and-return, organizational splits, and acquisitions all affect trust domain naming and federation state.
+- **Customers access MCs.** Customers would have access to key material stored in a management cluster.
```
puja108, line 109: but only limited to the MCs intermediary and the things it signs, right? as we kinda consider the MC to "belong to the customer" we do accept that as long as it is possible to ensure limited blast radius and permissions from MC components, which from how I understand Workload identity would be inherently possible, with the caveat of it actually needing to be somehow enforced/implemented.

workload-identity/README.md
stone-z: The actual location of all SPIRE roots will definitely get more discussion. We also need to figure out where to run ours. It may turn out that all SPIRE roots are some entirely different thing outside of our normal platform (like Teleport).

Ideally, customers would run their own SPIRE root however/wherever they run their most heavily protected infrastructure. This should be a discussion with them during onboarding.

Putting it in a WC would be an attempt to enable ourselves and customers who don't/won't run their own. If it's a goal to use strong cryptographic identities for auth from day 0 (also for things like authenticating our tools with their clusters), then we need to give them something on day 0. But I agree it might not be a WC, it could be something different we don't have a name for yet.

It shouldn't be on a management cluster because we've opened those up to usage by arbitrary users, and are running other workloads there that greatly increase the likelihood of compromise.

Root signing keys would typically included in the highest risk classification and protection level in a risk framework because compromise is so catastrophic. In organizations that already use X509 for anything else, even their cluster admins wouldn't have access to that material.

So, an MC wouldn't satisfy risk management frameworks, plus some more minor downsides of coupling the root with the lifecycle of a given region. This root would need to outlast all MCs' lifecycles. I haven't looked into the details of moving key material between regions, but it would in any case be a risky operation. In a WC we could at least heavily restrict access and move the cluster between management clusters in the future, but I agree it could be that something else is actually needed here.

stone-z:
> Eg what would be the benefit for a customer to use spiffe/spire if they can simply use the workload identity feature in EKS? Can we convince customers that our spiffe/spire solution is worth the extra complexity and effort?

> That's a good question. We do need workload identity for us but can our spire connect with whatever customers want to us instead of the root spire? If yes, maybe we should focus on our implémentation and provide a customer spire when needed? This is must have onprem though

One of the named alternatives is orchestrating the provider-native mechanisms instead. That could work too, it just requires far more work to productize. With SPIRE, we'd need to build very little in order to automatically issue every customer workload an identity, with little or no involvement from them, and that identity could be used across both our fleet and theirs. It also means auth itself is provider-independent; IRSA wouldn't be natively usable without extra work for authenticating with resources in other clouds, SaaS systems, or on-prem resources in hybrid setups. Orchestrating the provider-specific mechanisms would require us to build more integration, and we'd still need to run SPIRE for ourselves and for on-prem customers (most likely).

It isn't possible to use SPIRE to just "drive" a provider system directly, because one system has to control the signing keys. I haven't looked to see if it's possible to issue an intermediate signing cert from e.g. IRSA and then give that to SPIRE, but that wouldn't be using the provider tools anymore anyway (and I doubt it's possible).

All that said, customers could still choose to use an alternative on top of the identities we issue, this wouldn't limit them. They could use both in parallel, or for different purposes.

workload-identity/README.md
stone-z: Not specific to (G), but in general we would need one trust _federation_ between our SPIRE and each customer, and any services without GS or customer identities would need to be configured to trust each SPIRE (us and each customer) individually.

workload-identity/README.md
stone-z: That's true. I don't think SPIRE makes the failure domain any worse in hybrid cases; they already have a dependency on the MC, and WC-local SPIREs would keep things running in temporary failure cases. We just might need to treat the WCs as semi air-gapped if customers have "frequently offline" WCs.

workload-identity/README.md
stone-z: It is exposed when an application presents its own identity. There isn't an index or anything to look up all possible identities, as far as a I'm aware (that would be strange and concerning).

workload-identity/README.md
stone-z: See also the note about Named Paths / Name Constraints. It isn't currently possible to limit the scope of an intermediate signing key to e.g. "only workloads under this MC". The signing key could be used to sign a certificate for any workload in the customer's fleet, even if it's in another cluster. This will change "someday" but it's the reality for now. So the nested SPIRE is in practice a mini root. For that reason, we definitely can't put our own there, and I guess customers also _shouldn't_, but this should be refined and will be also up to their risk tolerance.

stone-z: That's actually a really good point. Until Name Constraints are possible, we probably _should_ have isolated root _and_ nested SPIREs for both us and customers :thinking: that's potentially a lot of extra infrastructure.

workload-identity/README.md
stone-z: For sure. I was hoping that was implied by "sufficiently available", but I could put in an explicit mention if it's helpful

workload-identity/README.md
puja108: I think the use case is anyway rare and if at all rather limited to we have "3 customers with Jenkins"

workload-identity/README.md
puja108: ah, right, seems I skipped reading that and went directly to read only a and b ;)

workload-identity/README.md
puja108: yes, that can and is already sometimes the case in some of our customers, so I guess that's fine

workload-identity/README.md
puja108: then I think this might be fine, customers that have a high privacy requirement here, would need to use some kind of code naming scheme in place, but I think that's overthinking it

workload-identity/README.md
puja108: hmm, right, and if it's only to limit blast radius

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+| 5 — New customer, no change | New trust domain + new agents + one federation entry. Nothing existing is modified. |
+| 6 — On/offboarding, no authz change | Policies supporting wildcards are very flexible. All policies (even without wildcards) work for same trust domain. Cross-domain trust (e.g. a customer service trusting Giant Swarm applications) would need to be reconfigured. |
+| 7 — No cross-customer impersonation | Separate root CAs per customer. Only GS leaf credentials in customer infrastructure, no company CA keys. |
+
+## Open questions
+
+1. **Semantic vs structural IDs.** Encapsulating customer and architecture information in an ID makes it easy to understand, but imposes rigidity, leaks information, makes and long-term assumptions that won't hold. Instead of `v1/giantswarm/platform/falco/v1-2-3/customer-x/eu-west-1/mc-name/wc-name`, consider `v1/giantswarm/platform/falco/v1-2-3`. The second more semantically identifies the application independently of the architecture, but additional work must be done to support cluster/customer/regions.
```
puja108, line 217: I'd like to gather use cases where per cluster/customer/region would make sense. have we asked customers about use cases? If we do not see any it's fine to omit I guess.

Question would also be if it would be possible to add later on if needed, or at least once a new customer would have a use case.

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+2. **What's in a name?** Especially if using structural names, what are the dimensions that must be included in the path?
+
+3. **SPIFFE ID path segment ordering.** Should customer precede region (`/<customer>/<region>/...`) or follow it (`/<region>/<customer>/...`)? Zach is in favor of customer first, then region, if using structural IDs.
+
+4. **Dual identities for Giant Swarm workloads.** Some workloads need both a global company identity (for external services) and a customer-local identity (for resources that must survive offboarding). It is possible to issue multiple identities to a workload, which must be aware of which to use. Is that useful? Should workloads receive SVIDs from both planes? What are the complexity and security implications?
```
puja108, line 223: I'd prefer to keep it simple and always use the GS ID for GS components but I know that would increase off-boarding efforts. Question is how high those efforts would be?

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+5. **How many customer SPIREs?** In theory, we could operate customers with only WC or MC SPIREs, and not require both. The MC SPIRE could issue for all WC agents, or WC SPIREs could nest directly under the global root. The most resilient would be to use both.
+
+6. **How/where does Giant Swarm run our regional SPIRE?** We can't (yet) run nested SPIRE servers within customer MCs because they would have access to key material that allows lateral movement to other customer trust domains. To keep "per-customer, per-region" failure domains, we would need to run a SPIRE server "next to" each MC, but in a place customers can't access. Where is that? (there is a future feature called "Name Constraints" support which will someday allow us to safely run it on the customer MC, but it's already years in the making).
```
puja108, line 227: This might be becoming a general issue, where we have things we might not want to expose to customers but would be on MCs. I'm thinking the future of MCs might need to be multi-cluster, which at least in the case of cloud would be fine as long as we use sth like EKS, but in case of onprem it would be too much overhead most probably. Or we'd need to look into other compartmentalization techniques. Or we do it like we used to do with Vault, where we had it running next to the MC in some VM IIRC. Might be a topic to discuss in SIG arch

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+7. **Air gap?** I think it makes sense to assume we need this eventually. Are there airgap-specific use cases?
```
puja108, line 229: With current customers we have at least the case where WCs might be offline for a few days.

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+8. **Support for external --> any customer authentication.** Use case G describes the possibility for an external service to authenticate "any instance of one|any application belonging to any Giant Swarm customer". This is currently very difficult to implement. The correct solution would involve identities signed by both Giant Swarm and the customer. This is different than dual identities, this is a single identity which is signed by two parties, which is not yet supported by SPIRE. So, how important is this use case?
```
puja108, line 231: I kinda like to keep per customer addition of auth for external use cases, it makes any addition or removal of a customer's app a conscious decision

workload-identity/README.md
puja108: I would also default to them owning it

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+9. **Customer SPIRE root placement.** The customer root needs to live somewhere highly available. Options include a dedicated WC in the customer's primary region, co-location in the first MC, a lightweight non-K8s host, or customer-provided infrastructure. The dedicated WC is the current recommendation, but introduces a single-region placement for a cross-region dependency.
+
+10. **DNS and OIDC endpoint ownership model.** With a single trust domain per customer, there is one DNS name to manage (e.g. `wid.acme.example.com`). Should customers own this DNS and delegate to Giant Swarm during the relationship, or vice versa? Zach suggests customers own the DNS entry and point it back at their own infrastructure after offboarding.
+
+11. **Agent attestation method.** More work is needed to investigate the technical means by which applications will attest in a cluster. Is someone interested in doing this?
+
+12. **Network path: agent to regional server, and MC/regional to customer root.** How will agents reach their issuing servers? The MC→customer root connection is cross-region if the root is in a different region.
```
puja108, line 239: for x-region we might also need to think about how much traffic it would incur

workload-identity/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+13. **Service mesh integration.** SPIRE can feed SVIDs to Envoy and Cilium. No investigation has been done on the technical details for that. Is someone interested in doing this?
+
+## Alternatives
+
+- Prefer provider-native identity mechanisms first. Instead of defaulting to a SPIRE identity, we instead use native platforms like AWS IRSA and Azure Workload Identity. The downside of this is the heterogeneity and complexity of maintaining parallel identity minting systems for each provider. Plus, we would still need to offer a solution for on-prem clusters.
+- Build around a single alternative offering. Teleport would be the most obvious choice here, but in theory we could also choose AWS or Azure identities. We see the lock-in as a major downside. If using Teleport, we'd need to upgrade to Teleport Enterprise, which has unknown cost. The effort to maintain such a solution is unknown but may be less than SPIRE.
```
puja108, line 246: This would also be super hard to get through with all customers. There will always be customers blocking the use, especially if it is connected to the US, now with all the sovereignty hype going on

workload-identity/README.md
puja108: as I mentioned in https://github.com/giantswarm/rfc/pull/144/changes#r3073434770 the problem of running secure/isolated workloads in an MC(-like) place is becoming more and more important and might deserve a special session in SIG arch

workload-identity/README.md
lyind, line 217: I read https://spiffe.io/docs/latest/spiffe-specs/spiffe-id/#2-spiffe-identity and didn't find any mention of the *query* URI part, in the SPIFFE docs.

Usage of the query part could allow attaching unordered attributes, as key-value pairs.

For example, would `spiffe://giantswarm.io/v1/giantswarm/platform/klaus/v1-2-3?wc=operations&mc=gazelle&team=planeteers` be a valid ID?

lyind:
After reading the RFC and realizing the large scope of this, I wonder:
- how well this will mesh with classic identity systems (Entra, etc.).
- what UIs are going to be used, if those already exist and how mature they are. Since managing, displaying and analyzing the trust relationships and entities could get demanding (compare classic systems).
- why it is projected to demand that much ressources (CPU/RAM, will tuning be required?): https://spiffe.io/docs/latest/planning/scaling_spire/

piontec:
Since this is forcing us to run our own CA root, should we also reconsider our assumption, that for artifact signing we want the "keyless" mode of `cosign`? Our main goal so far was to avoid managing the CA, but if we change that, does it make the situation any different? Is there any kind of integration possible between `cosign` and `spiffe`, like sharing common trust root? I don't think this is changing a lot, but maybe you've seen something related.

stone-z:
for @lyind:

> how well this will mesh with classic identity systems (Entra, etc.).

It parallels them, for machine identity use cases. We shouldn't e.g. create Google users for a machine, and we won't issue SVIDs for a human. However, a system (say, some API) can trust both Google and SPIRE as identity providers in order to authenticate humans or machines. Most large SaaS and infra providers also support this federation, which means we could authorize one of our workloads bearing a SPIFFE identity to access supported systems without pre-sharing any credentials.

> what UIs are going to be used, if those already exist and how mature they are

There is at least [an official UI](https://github.com/spiffe/tornjak). I haven't used it yet, but that would be a starting point if you have a particular use case in mind.

> why it is projected to demand that much ressources

I don't really know the exact consumption patterns, but I would assume the largest computational work is cryptographic operations, and I'd assume the largest memory demand is [de]serialization of certificate document structures. Some tuning will definitely be required as we figure out how to size each level of the tree. JWTs also require more compute than X.509s, so the consumption will vary based on usage. The [Deployment Sizing Considerations section](https://spiffe.io/docs/latest/planning/scaling_spire/#deployment-sizing-considerations) gives an overview, but the nesting must be taken into account. We will certainly have 10,000 workloads on thousands of agents, but the load associated with managing them can be distributed among multiple nested SPIRE instances.

---
and for @piontec

> should we also reconsider our assumption, that for artifact signing we want the "keyless" mode

Yeah good thinking. cosign does natively support SPIFFE identities. IMO we should stick to keyless at least for now, because although 1. all members of the trust domain can implicitly authenticate each other and 2. it would be possible to sign build artifacts with a SPIRE-provided cert, we would still need to handle key distribution to the signing points and clients would still need to be able to validate the signatures. We would likely still need to build some glue and end up needing to distribute public keys ourselves.

I'd suggest we use federation to trust e.g. the GitHub or CircleCI authority for the keyless signing case, at least to start. If we were to move builds into Tekton, or if SaaS build systems introduce native integration points for SPIFFE identities, it would make sense to revisit it.

PR #145, opened 2026-04-28
Add RFC: Silence-based maintenance windows

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/35664

This PR adds a new RFC proposal for silence based maintenance windows based on a new CLI and the already existing Alert timeline dashboard.

### DIFF
--- a/silence-based-maintenance-windows/README.md
+++ b/silence-based-maintenance-windows/README.md
@@ -93,6 +93,10 @@ A Go service deployed between Alertmanager and PagerDuty that reads routing rule
 2. **Document the maintenance workflow** -- write a runbook explaining how to use temporary silences together with the Alerts Timeline dashboard during maintenance operations.
 3. **Communicate to on-call engineers** -- announce the new workflow and tooling.
 
+## Future work
+
+- **Scheduled silences** -- extend the `Silence` CR with a start date so that silences can be created ahead of time and activated automatically at a given moment. This would let engineers prepare maintenance windows in advance instead of having to be at the keyboard exactly when the operation starts.
+
 ## Communication plan
 
 - Post the RFC PR to `#news-ops` and relevant team channels for review.
--- a/silence-based-maintenance-windows/README.md
+++ b/silence-based-maintenance-windows/README.md
@@ -44,7 +44,7 @@ While alerts are silenced and not routed to PagerDuty, they can still be observe
 
 ### CLI tool: `silencectl`
 
-A new CLI tool to make creating and managing temporary silences straightforward.
+A new CLI tool to make creating and managing temporary `Silence` CRs straightforward.
 
 **Core functionality:**
 - Create a temporary silence for a given installation with a specified duration
--- a/silence-based-maintenance-windows/README.md
+++ b/silence-based-maintenance-windows/README.md
@@ -1,0 +1,100 @@
+---
+creation_date: 2026-04-28
+issues:
+- https://github.com/giantswarm/giantswarm/issues/35664
+owners:
+- https://github.com/orgs/giantswarm/teams/team-atlas
+state: review
+summary: Use temporary Alertmanager silences and the existing Alerts Timeline dashboard instead of building alert routing to PagerDuty during maintenance operations like cluster upgrades.
+---
+
+# Silence-based maintenance windows
+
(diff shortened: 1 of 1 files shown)


silence-based-maintenance-windows/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Implementation plan
+
+1. **Build `silencectl`** with support for creating temporary silences per installation, duration-based expiry, graceful cleanup on interruption, and a non-interactive mode.
+2. **Document the maintenance workflow** -- write a runbook explaining how to use temporary silences together with the Alerts Timeline dashboard during maintenance operations.
+3. **Communicate to on-call engineers** -- announce the new workflow and tooling.
+
+## Communication plan
+
+- Post the RFC PR to `#news-engineering` and relevant team channels for review.
```
QuentinBisson:
```suggestion
- Post the RFC PR to `#news-ops` and relevant team channels for review.
```

silence-based-maintenance-windows/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+@TheoBrigitte (Team Atlas)
+
+## Who is affected / stakeholders
+
+- Team Atlas (owns the observability platform)
+- All on-call engineers (affected by alert noise during upgrades)
+- Any engineer performing cluster upgrades or other risky maintenance operations
+
+## Preferred solution
```
QuentinBisson, line 29: I still think adding a new silence type or adding the start date to the Silence CR would also help so we can schedules silences in advance. Can this be added as a next step?

silence-based-maintenance-windows/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Replace per-installation alert routing with **temporary Alertmanager silences** and the **Alerts Timeline dashboard** for situational awareness.
+
+### How it works
+
+When an engineer is about to perform a risky operation (e.g. a cluster upgrade), they:
+
+1. **Create a temporary silence** covering the target installation for the expected duration (e.g. 2-3 hours) using a new CLI tool (`silencectl`).
```
JosephSalisbury, line 37: does this create a silence in alertmanager, or a silence cr?

silence-based-maintenance-windows/README.md
TheoBrigitte: Yes sure this makes sense I will add it into next steps.

silence-based-maintenance-windows/README.md
TheoBrigitte: Our idea is to create a Silence CR, and add it into the correct GitOps rep

JosephSalisbury: okay cool, if the tool is (just?) managing silence crs, this all makes sense to me :D

Gacko (approved):

QuentinBisson:
@TheoBrigitte if you address the reviews i'm fine :)

QuentinBisson (approved):

TheoBrigitte:
I've addressed the reviews and did some minor updates.

teemow:
Just a general comment on this. I think watching dashboards is a significant step back. With the advent of automated rollouts and auto-releases people should not have to set silences and watch dashboards manually. This should have rather been taken forward to auto-route alerts to teams or individuals. 

So I consider this RFC only a temporary solution. This needs a follow-up.

QuentinBisson:
@teemow you're right that we need auto routing of alerts and I think we are not that far from it. 
What this RFC addresses though, is the lack of routing rules that opsgenie had in pagerduty so this use case should be for maintenance windows only (e.g. i'm upgrading MC A and I know this will page) to avoid paging other team. It could also bé used to silence non kaas alerts during cluster upgrades so that say, atlas, is not paged before phoenix when it's related to the cluster not upgrading properly.

TheoBrigitte:
As Quentin mentioned we are quite limited with PagerDuty, we tried to find ways to auto-route alerts to be on par with what we had before with opsgenie, but we could not find anything without over engineering a new complex and in-house solution. This is what we found to be most acceptable regarding time and effort for implementing this. We anyway want to re-evaluate our paging solution later on, so I will make sure we take this into consideration then.

PR #148, opened 2026-06-02
fix: explicitly mention the source of the time stamp for dev tags

piontec:
Update the semVer automatic spec with info about the source of the dev tag's time stamp.

### DIFF
--- a/semver-based-automatic-upgrades/README.md
+++ b/semver-based-automatic-upgrades/README.md
@@ -134,7 +134,7 @@ not semVer compliant.
 This RFC introduces two related parts. The second one, automated app upgrades, will become possible as a
 configuration of a standard flux feature. It is blocked by implementing the first one.
 
-### Proposed default tagging schema
+### Tagging schema
 
 In order to become semVer compatible and our tags to reflect software maturity, we will introduce the
 following tagging schema:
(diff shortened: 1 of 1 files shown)


semver-based-automatic-upgrades/README.md
```diff
@@ -147,7 +147,8 @@ following tagging schema:
   make these builds to have a tag that allows to identify the branch it is coming from and to make them
   sortable according to semVer. The proposed schema is thus to append suffix `-dev.[BRANCH].[YYYY-MM-DD].[HH-MM-SS]`
   where `BRANCH` is the name of the branch the build is coming from and the suffix is a time stamp with date
-  and time parts. Date and time use hyphens as separators (e.g. `2026-01-27` and `09-49-59`) so that each
+  and time parts, where the time stamp comes from the current commit. Date and time use hyphens as separators
+  (e.g. `2026-01-27` and `09-49-59`) so that each
```
uvegla: This makes me think: when we calculate this, does it make sense to convert all the timestamp on the commits to UTC? 🤔 Would the sorting be messed up if people across multiple time zones contribute? I am sure the commit contains the time zone info, but we need to convert.

uvegla (approved):

semver-based-automatic-upgrades/README.md
piontec: Good point, I will add explicitely that we convert to UTC

PR #150, opened 2026-06-08
feat: spec the new dev tag format

piontec:
We often use tags as kubernetes label values, but with verbose git branch names, the current naming schema easily crosses the 63 chars per label limit. This PR redacts the branch title and adds commit hash to make them unique and to aid tag-to-branch-and-commit mapping.

https://github.com/giantswarm/giantswarm/issues/35122

### DIFF
--- a/semver-based-automatic-upgrades/README.md
+++ b/semver-based-automatic-upgrades/README.md
@@ -1,12 +1,15 @@
 ---
 creation_date: 2025-12-11
 issues:
-- https://github.com/giantswarm/giantswarm/issues/24237
-last_review_date: 2026-05-20
+  - https://github.com/giantswarm/giantswarm/issues/24237
+last_review_date: 2026-06-08
 owners:
-- https://github.com/orgs/giantswarm/teams/team-honeybadger
+  - https://github.com/orgs/giantswarm/teams/team-honeybadger
 state: approved
-summary: We want to use flux and flux-operator's automatic upgrades capabilities to create automatic upgrades for different release stages, so we don't have to manually or through extra automation care about those rollouts.
+summary:
+  We want to use flux and flux-operator's automatic upgrades capabilities to create automatic upgrades for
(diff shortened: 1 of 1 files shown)


semver-based-automatic-upgrades/README.md
```diff
@@ -144,15 +147,25 @@ following tagging schema:
...
-  where `BRANCH` is the name of the branch the build is coming from and the suffix is a time stamp with date
-  and time parts, where the time stamp comes from the current commit's commiter date converted to UTC
-  (so the date displayed with `git log --format='%ci'` format, not the author's date).
-  Date and time use hyphens as separators (e.g. `2026-01-27` and `09-49-59`) so that each
-  part is a semVer alphanumeric pre-release identifier, which allows leading zeros and maintains correct
-  lexicographic sort order. For example, if the last stable tag in history is `1.9.1` and the branch name is
-  `my-feature`, the build results in a tag like `1.9.2-dev.my-feature.2026-01-27.09-49-59`.
+  make these builds to have a tag that allows to identify the branch and the commit it is coming from and to
+  make them sortable according to semVer. The proposed schema is thus to append suffix
+  `-dev.[BRANCH].[YYYY-MM-DD].[HH-MM-SS].h[COMMIT_SHA]` where `BRANCH` is the (sanitized) name of the branch
```
QuentinBisson, line 152: Why not -rc?

QuentinBisson (approved):

semver-based-automatic-upgrades/README.md
piontec: For simplicity, we assumed there's just 1 way to use RC suffixes and that is `-rc.NUM`. We won't run past 63 chars here and they are manually created, like stable tags, so there's no need for branch name.

QuentinBisson: Alright thanks

PR #151, opened 2026-06-22
fix(check-rfc-format.py): front-matter style warnings

lyind:
Small front-matter indentation fixes, to avoid  `check-rfc-format.py` throwing errors.

### DIFF
--- a/semver-based-automatic-upgrades/README.md
+++ b/semver-based-automatic-upgrades/README.md
@@ -1,15 +1,13 @@
 ---
 creation_date: 2025-12-11
 issues:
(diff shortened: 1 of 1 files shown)


piontec (approved):
LGTM, but this list formatting style is bad IMO and against majority of MD editors. Check with md validators - I use markdownlint with vim and it always reformats that style

lyind:
> Check with md validators - I use markdownlint with vim and it always reformats that style

Yes, I noticed that. This is just to be able to merge RFC updates in handbook. Permanent solution would be to improve that Python-based style checker/validator (`check-rfc-format.py`).

PR #153, opened 2026-07-07
Add app-testing contract RFC

QuentinBisson:
Defines a harness-neutral contract so the same app test files run under both app-test-suite (fast checks per PR on kind) and apptest-framework (full e2e nightly on workload clusters).

We are keeping both harnesses on purpose: they are a deliberate fast/slow split, not redundant tools. The motivation is not deduping existing tests (there is little to dedup); it is that the two harnesses had divergent authoring models, so covering both cadences cost 2x from scratch and adoption of both was close to zero. This contract unifies the authoring model: one directory, one idiom, both runners.

Core of the proposal:

- one conventional directory per repo (`tests/app/`), one Go module or one Python project, presence is the opt-in; committed lockfiles, installed frozen/offline
- three peer test types (`smoke`, `functional`, `upgrade`, matching ATS's existing StepTypes); the upgrade flow runs only `upgrade`-typed tests, before and after the upgrade (`APP_TEST_UPGRADE_STAGE`), never re-running smoke/functional; the pre run is the baseline that attributes a post failure to the upgrade; pure side-effect seeding goes in a `pre-upgrade` hook
- inputs via `KUBECONFIG` + canonical `APP_TEST_*` env vars (legacy `ATS_*` exported alongside); tests never provision
- runner guarantees: app deployed and settled, env exported, ordered execution, "no tests for this type" is a pass but zero collection is never silent, junit output
- cadence and feedback latency: cloud-only tests run nightly, detached from the PR; the contract keeps that a conscious choice (visible in the diff and collected-count) and gives an on-demand `/run` full-flow trigger to pull the signal forward
- the upgrade flow is inferred from the presence of `upgrade`-typed tests (no config switch); optional `expectedTypes` makes named types mandatory, so a typo'd tag fails instead of silently skipping
- a conformance suite (fixture app + guarantee assertions) is the acceptance gate both runners run in CI, including a cross-runner parity check (same fixture, both runners, matching outcome) that guards against settled-semantics drift (Helm-installed vs App-CR-deployed); team-tenet stewards the contract
- harness-specific tests remain: runtime skips on environment properties inside the shared dir, and in-process atf suites for anything needing MC/bundle/cloud machinery

Grew out of the discussion around unifying chart and e2e test authoring (pilot groundwork: giantswarm/muster#954). Stakeholders: team-honeybadger (app-test-suite), team-tenet (apptest-framework), team-bumblebee.

### DIFF
--- a/app-testing-contract/README.md
+++ b/app-testing-contract/README.md
@@ -1,6 +1,6 @@
 ---
 creation_date: 2026-07-07
-issues: []
+issues: []j
 owners:
   - https://github.com/orgs/giantswarm/teams/team-bumblebee
   - https://github.com/orgs/giantswarm/teams/team-honeybadger
@@ -15,11 +15,17 @@ summary:
 
 # The app-testing contract
 
+## Glossary
+
+- test toolkit: the test runner and orchestrator, either `app-test-suite` (ATS) or `apptest-framework` (ATF)
+- test framework: software framework used to implement the tests, e.g. Ginkgo or pytest
+- test suite: a collection of tests, usually in one repo, that test a specific chart (app)
+
 ## Problem
 
-We test our helm charts delivering managed apps using two frameworks: `app-test-suite` (ATS) and
-`apptest-framework` (ATF). So far, it was not clear which framework teams should use and how to write the
-tests. We propose to keep both frameworks, but specialize them to the two most frequent use cases:
+We test our helm charts delivering managed apps using two toolkits: `app-test-suite` (ATS) and
+`apptest-framework` (ATF). So far, it was not clear which toolkit teams should use and how to write the tests.
+We propose to keep both toolkits, but specialize them to the two most frequent use cases:
 
 - **app-test-suite (ATS)** - should provide rapid local feedback on PRs and in local dev environments. Its
   goal is to deploy the helm chart under tests as fast as possible and start testing the chart's
@@ -31,25 +37,25 @@ tests. We propose to keep both frameworks, but specialize them to the two most f
   runs the full suite, preferably nightly.
 
 With this in mind, it's clear that to provide a comprehensive test coverage and to follow the "fail fast"
-principle, we need to use both frameworks.
+principle, we need to use both toolkits.
 
-The trouble is writing the tests. The tests author doesn't want to implement the same or very similar set of
-tests twice, once for each framework. The goal of this doc is to propose a test implementation spec that will
-allow both ATS and ATF to run the same test suites in their respective environments. Full parity of tests
-might not be possible, as test itself might depend on the environment, but we want to get as close as
-possible. A local kind cluster is fast but can't do cloud identity, real storage, or upgrades; a workload
-cluster can. If a test suite needs to be aware of these differences, the author needs to get the possibility
-to gate on the capabilities it needs.
+The trouble is writing the tests (test suites). The tests author doesn't want to implement the same or very
+similar set of tests twice, once for each toolkit. The goal of this doc is to propose a test implementation
+spec that will allow both ATS and ATF to run the same test suites in their respective environments. Full
+parity of tests might not be possible, as test itself might depend on the environment, but we want to get as
+close as possible. A local kind cluster is fast but can't do cloud identity, real storage, or upgrades; a
+workload cluster can. If a test suite needs to be aware of these differences, the author needs to get the
+possibility to gate on the capabilities it needs.
 
-This goal of this RFC is not to dedupe existing tests. It's to define how the two frameworks are different,
-what environments they provide and what is the convention that, when respected, can allow to fit both
-frameworks with the same test code.
+This goal of this RFC is not to dedupe existing tests. It's to define how the two toolkits are different, what
+environments they provide and what is the convention that, when respected, can allow to fit both toolkits with
+the same test code.
 
 ## Decisions
 
 ### ATS and ATF
 
-We keep both frameworks, but specialize them to the two most frequent use cases:
+We keep both toolkits, but specialize them to the two most frequent use cases:
 
 - `ats` is meant for rapid feedback testing on PRs and in local dev environments. It deploys the helm chart
   under tests as fast as possible and starts testing the chart's functionality. It sacrifices all the possible
@@ -58,22 +64,24 @@ We keep both frameworks, but specialize them to the two most frequent use cases:
   It creates a real workload cluster on a GS installation, installs the chart using the App Platform, and runs
   the full suite as a batch run, preferably nightly.
 - we want a test development convention that will allow app maintainers to write tests once and run them under
-  both frameworks, with the same test code.
+  both toolkits, with the same test code.
 
 ### Test Development Convention
 
 #### Assumptions
 
 To avoid forcing a specific test framework or technology on test authors, we decided to make a convention that
-tests are executed as a separate process by the test frameworks. The only requirements are that:
+tests are executed as a separate process by the test toolkits. The only requirements are that:
 
-- we group all the tests into:
+- for each test suite, we group all the tests into:
   - **smoke**: very basic tests that check if the app is deployed and running, and if the main functionality
     is working and worth even trying actual functional test. They should be fast, simple and reliable.
   - **functional**: tests that check the actual functionality of the app, and that it behaves as expected.
     They should be more complex and cover more scenarios than smoke tests.
   - **upgrade**: tests that check if the app can be upgraded from a previous version, by default the last
-    stable version available in the OCI registry.
+    stable version available in the OCI registry. The tests from this groups are executed twice, once before
+    the upgrade (the "old" version, that we start the upgrade test with) and once after the upgrade (the "new"
+    version, under the test).
 - a single test can be of multiple types, for example a test that checks if the app is deployed and running
   can be both smoke and functional; a test that checks if the main page of an app loads can (and probably
   should) be functional and upgrade.
@@ -82,10 +90,10 @@ tests are executed as a separate process by the test frameworks. The only requir
   (golang) or markers (python).
 - all the information about the test environmenrt is passed to the tests via environment variables, and the
   tests should not depend on any other external information (like a config file or a specific cluster setup).
-  The test frameworks are responsible for setting up the environment and passing the information to the tests.
-- exit code `0` means the test run passed, exit code `5` that the run was successful, but only because it
-  executed no test and any other non-zero exit code means the test run failed.
-- test developer can deliver hooks that are executed by the test frameworks (see below).
+  The test toolkits are responsible for setting up the environment and passing the information to the tests.
+- exit code `0` means the test run passed, including the case where it executed no tests, and any other
+  non-zero exit code means the test run failed.
+- test developer can deliver hooks that are executed by the test toolkits (see below).
 
 #### The conventional directory
 
@@ -113,13 +121,13 @@ already has:
 | `upgrade`    | upgrade flow, before and after | checks the app still works across an upgrade |
 
 `upgrade` is its own type, not a flag on the others. The upgrade flow doesn't re-run smoke and functional; it
-runs the `upgrade` tests, once on the old version (`APP_TEST_UPGRADE_STAGE=pre`) and once after upgrading
-(`=post`). The pre run is the baseline: if it passes and post fails, the upgrade caused it, not something that
-was already broken.
+runs the `upgrade` tests, once for the "old" version (the version to test the upgrade from, usually the last
+stable version) and once after upgrading to the "new" version (the version under test). The first run is the
+baseline: if it passes and post fails, the upgrade caused it, and the upgrade test fails.
 
 Most upgrade checks are symmetric ("the app answers") and assert the same thing both times. When some state
-has to survive the upgrade, either seed it in the `pre-upgrade` hook and check it in the post run, or keep it
-in one test that branches on the stage:
+has to survive the upgrade, either seed it in the `pre-run` hook and check it in the post run, or keep it in
+one test that branches on the stage:
 
 ```go
 if os.Getenv("APP_TEST_UPGRADE_STAGE") != "post" {
@@ -127,14 +135,14 @@ if os.Getenv("APP_TEST_UPGRADE_STAGE") != "post" {
 }
 ```
 
-Seeding is a side effect, so it's a hook; checking is an assertion, so it's a test. Upgrade tests and the
-`pre-upgrade` hook also get `APP_TEST_UPGRADE_FROM_VERSION` / `APP_TEST_UPGRADE_TO_VERSION`.
+Seeding is a side effect, so it's a hook; checking is an assertion, so it's a test. Upgrade tests and hooks
+hook also get `APP_TEST_UPGRADE_FROM_VERSION` / `APP_TEST_UPGRADE_TO_VERSION`.
 
 #### Hooks
 
 Hooks do things with side effects (install a prerequisite, create a pod, clean up); tests check things. Hooks
 are delivered as executables in the `tests/app/hooks/` directory, and are optional. They should be implemented
-in platform-independent way, preferably in bash or python. Hooks are executed by the test framework. Splitting
+in platform-independent way, preferably in bash or python. Hooks are executed by the test toolkit. Splitting
 them is what lets the upgrade flow seed state without re-running a suite. Hooks get the test information
 through environment variables, the same as tests. The hooks are:
 
@@ -149,53 +157,58 @@ A missing hook is ignored. A non-zero exit fails the test run.
 
 #### Full Test Flow
 
-The test framework (`ats` or `atf`) runs the tests using this flow:
+The test suite is always invoked by the test toolkit (`ats` or `atf`). The toolkit runs the tests using this
+flow:
 
-1. Test framework detects tests are present in `tests/app/`.
-1. Test framework prepares the cluster used for testing (installs tools or dependencies it needs to execute
-   tests).
+1. Test toolkit detects that tests are present in `tests/app/`.
+1. Test toolkit installs software dependencies for the test suite (go or python) using the lockfile in
+   `tests/app/`.
+1. Test toolkit prepares the cluster used for testing (installs tools or dependencies it needs to execute
+   tests; might be a no-op, depends on the test toolkit).
 1. If present, the `setup` hook runs after the cluster is ready, before the app is deployed.
 1. The app is deployed using the passed helm chart and the installation is settled (chart install exists
    cleanly).
-1. For each test type of `smoke`, `functional`, and `upgrade` (where `upgrade` tests are executed twice, first
-   for old version, then after the upgrade, for the new version), in this order:
+1. For each test `type` in `smoke`, `functional`:
    1. `pre-run` hook runs for `type` tests (if present).
    1. Tests are executed for `type` type, using either `go test` or `pytest`, depending on the detected module
       type.
    1. `post-run` hook runs for `type` tests (if present).
-   1. If it's an `upgrade` type test and the execution for `old` version succeeded, the app is upgraded to the
-      `new` version.
-1. If present, the `teardown` hook runs after all tests, before the harness tears anything down.
+1. If there are `upgrade` tests:
+   1. If the version under test (new) is already installed in the cluster, it is uninstalled.
+   1. The "old" stable version of the app is installed.
+   1. `pre-run` hook runs for `upgrade` tests (if present).
+   1. `upgrade` type tests are executed for the "old" version.
+   1. `post-run` hook runs for `upgrade` tests (if present).
+   1. The app is upgraded to the "new" version.
+   1. `pre-run` hook runs for `upgrade` tests (if present).
+   1. `upgrade` type tests are executed for the "new" version.
+   1. `post-run` hook runs for `upgrade` tests (if present).
 1. The app is uninstalled.
+1. If present, the `teardown` hook runs after all tests, before the harness tears anything down.
 
 #### Inputs
 
-Tests get everything from the environment. They don't provision anything: no clusters, no chart installs, no
-App CRs.
+Tests get all the information about what is being tested from the environment variables. The following
+variables are guaranteed to be set by the test toolkit before the tests are executed:
 
-| Variable                                                        | Required          | Meaning                                                                                                                                                                                                                                 |
-| --------------------------------------------------------------- | ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `KUBECONFIG`                                                    | yes               | kubeconfig of the cluster the app is deployed on (never the MC)                                                                                                                                                                         |
-| `APP_TEST_TYPE`                                                 | yes               | the type currently being run                                                                                                                                                                                                            |
-| `APP_TEST_RELEASE_NAME`                                         | yes               | Helm release name of the app under test                                                                                                                                                                                                 |
-| `APP_TEST_RELEASE_NAMESPACE`                                    | yes               | namespace the app is deployed into                                                                                                                                                                                                      |
-| `APP_TEST_CHART_VERSION`                                        | yes               | version of the chart under test                                                                                                                                                                                                         |
-| `APP_TEST_CLUSTER_TYPE`                                         | yes               | topology of the cluster the app runs on: `kind` (local single-node), `capi` (a CAPI workload cluster), or `external` (a cluster the runner did not provision). Describes shape, not capability; gate on `APP_TEST_CAPABILITIES` instead |
-| `APP_TEST_CAPABILITIES`                                         | yes               | comma-separated capabilities the cluster actually provides, e.g. `cloud-identity,persistent-storage,load-balancer`; empty is valid. The runner sets it from what it provisioned or was handed. This is what a test gates on             |
-| `APP_TEST_KUBERNETES_VERSION`                                   | optional          | Kubernetes server version                                                                                                                                                                                                               |
-| `APP_TEST_VALUES_FILE`                                          | optional          | values file the app was deployed with                                                                                                                                                                                                   |
-| `APP_TEST_UPGRADE_STAGE`                                        | upgrade flow only | `pre` or `post`: which side of the upgrade this `upgrade`-type run is on                                                                                                                                                                |
-| `APP_TEST_UPGRADE_FROM_VERSION` / `APP_TEST_UPGRADE_TO_VERSION` | upgrade flow only | versions on either side of the upgrade; available to the `pre-upgrade` hook and `upgrade`-typed tests                                                                                                                                   |
-| `APP_TEST_EXTRA_*`                                              | optional          | harness extras, for example `APP_TEST_EXTRA_GITOPS_ENGINE`                                                                                                                                                                              |
+| Variable                                                        | Required          | Meaning                                                                                                                                                                                                                                       |
+| --------------------------------------------------------------- | ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `KUBECONFIG`                                                    | yes               | kubeconfig of the cluster the app is deployed on (never the MC)                                                                                                                                                                               |
+| `APP_TEST_TYPE`                                                 | yes               | the type currently being run (`smoke`, `functional`, `upgrade`)                                                                                                                                                                               |
+| `APP_TEST_TOOLKIT`                                              | yes               | The name of the toolkit running the tests (`ats` or `atf`)                                                                                                                                                                                    |
+| `APP_TEST_RELEASE_NAME`                                         | yes               | Helm release name of the app under test                                                                                                                                                                                                       |
+| `APP_TEST_RELEASE_NAMESPACE`                                    | yes               | namespace the app is deployed into                                                                                                                                                                                                            |
+| `APP_TEST_CHART_VERSION`                                        | yes               | version of the chart under test                                                                                                                                                                                                               |
+| `APP_TEST_CLUSTER_TYPE`                                         | yes               | type of the cluster the app runs on - a label: `kind` (local single-node), `capi` (a CAPI workload cluster), or `external` (a cluster the runner did not provision). Describes shape, not capability; gate on `APP_TEST_CAPABILITIES` instead |
+| `APP_TEST_CAPABILITIES`                                         | yes               | comma-separated capabilities the cluster actually provides, e.g. `cloud-identity,persistent-storage,load-balancer`; empty is valid. The runner sets it from what it provisioned or was handed. This is what a test gates on                   |
+| `APP_TEST_KUBERNETES_VERSION`                                   | optional          | Kubernetes server version                                                                                                                                                                                                                     |
+| `APP_TEST_VALUES_FILE`                                          | optional          | values file the app was deployed with                                                                                                                                                                                                         |
+| `APP_TEST_UPGRADE_STAGE`                                        | upgrade flow only | `pre` or `post`: which side of the upgrade this `upgrade`-type run is on                                                                                                                                                                      |
+| `APP_TEST_UPGRADE_FROM_VERSION` / `APP_TEST_UPGRADE_TO_VERSION` | upgrade flow only | versions on either side of the upgrade                                                                                                                                                                                                        |
 
-The prefix is `APP_TEST_`, which reads the same under either harness. ATS publishes these under the old `ATS_`
-prefix today. Where a name maps straight across (`ATS_X` to `APP_TEST_X`, e.g. `ATS_RELEASE_NAME`,
-`ATS_CHART_VERSION`, `ATS_CLUSTER_TYPE`, `ATS_EXTRA_*`) the runner exports both, so nothing breaks and new
-tests use `APP_TEST_`. Dual export isn't free (two names to know and grep for), so `ATS_` is deprecated and
-will be removed in a later change once repos have migrated.
-
-Four names are renamed, because the old ones were unclear or, for the upgrade stage, were never test-facing
-under one name to begin with:
+**Note**: These values come originally from `ats`, but are renamed here to match the test framework
+independence. Four names are renamed, because the old ones were unclear or, for the upgrade stage, were never
+test-facing under one name to begin with:
 
 | Legacy                                                            | Canonical                     | Note                                                                                                                                                                 |
 | ----------------------------------------------------------------- | ----------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


piontec:
Pushed a commit (`d72db9f`) folding **test-time controllers** into the contract — a declarative way for a repo to name the controllers (Flux, Argo, External Secrets, …) that must be running to reconcile the CRs its app creates, so any conforming harness bootstraps exactly those before deploy. It's the declarative sibling of the `setup` hook.

What it adds:
- **`### Prerequisite controllers`** section: schema in the shared config (`name` + `semver` with Masterminds/semver v3 semantics + optional per-harness `harness[]` values files), significant install order, reuse-if-present with a version check (out-of-range → fail, never modify a controller we find), install-only lifecycle, once per run.
- **Runner guarantees**: controllers bootstrapped and ready before the `setup` hook and the app deploy.
- **Implementation + Conformance**: both runners parse and bootstrap; the fixture declares a controller and the parity check covers it.

Three of the edits are **design changes to review**, not just additions:
1. **Shared config moves `tests/app/config.yaml` → `.apptest/config.yaml`** (code stays in `tests/app/`, declarations + values files live in `.apptest/`). Both harnesses already needed a shared, tool-neutral declaration file for controllers, and we wanted one file rather than two.
2. **The rejected `values:{ats,e2e}` alternative is softened**: still rejected for the app's deploy values, but the `controllers` section names per-harness values files via a `harness[]` list (harness name is a field, not a map key) — a bounded exception for runner-bootstrapped infra.
3. **Ordering**: controllers are bootstrapped *before* the `setup` hook (controllers are foundational infra the setup hook or app may depend on).

Happy to split any of these out or adjust if you'd rather keep the shared file under `tests/app/` or handle per-harness controller values differently. Companion ATS-side specs: giantswarm/app-test-suite#679.

app-testing-contract/README.md
```diff
@@ -0,0 +1,296 @@
...
+| `ATS_CLUSTER_VERSION`                                             | `APP_TEST_KUBERNETES_VERSION` | straight rename                                                                                                                                                      |
+| `ATS_EXTRA_UPGRADE_TEST_STAGE` (tests) / `ATS_HOOK_STAGE` (hooks) | `APP_TEST_UPGRADE_STAGE`      | value also changes: `pre_upgrade`/`post_upgrade` become `pre`/`post`. The runner does not alias the value, so existing upgrade tests reading the old one must update |
+
+`ATS_CHART_PATH` and `ATS_TEST_DIR` have no `APP_TEST_` equivalent; the contract doesn't expose them and they
+stay ATS-only. `APP_TEST_CAPABILITIES` is new, computed by the runner, with no `ATS_` predecessor.
+`KUBECONFIG` stays as-is; it's the standard name, not ours.
+
+#### Capabilities
+
+The cluster intended for PR testing is `kind`. As a simple cluster instance, it lacks features like cloud
```
QuentinBisson: I don't think that's true,  ATF is not running kind

app-testing-contract/README.md
```diff
@@ -0,0 +1,296 @@
...
+| `APP_TEST_RELEASE_NAMESPACE`                                    | yes               | namespace the app is deployed into                                                                                                                                                                                                            |
+| `APP_TEST_CHART_VERSION`                                        | yes               | version of the chart under test                                                                                                                                                                                                               |
+| `APP_TEST_CLUSTER_TYPE`                                         | yes               | type of the cluster the app runs on - a label: `kind` (local single-node), `capi` (a CAPI workload cluster), or `external` (a cluster the runner did not provision). Describes shape, not capability; gate on `APP_TEST_CAPABILITIES` instead |
```
QuentinBisson, line 202: For ATS all clusters are external so do we really need APP_TEST_CLUSTER_TYPE and APP_TEST_KUBERNETES_VERSION?

app-testing-contract/README.md
```diff
@@ -0,0 +1,296 @@
...
+      - file2.yaml
+  toolkitSpecific: # mutually exclusive with chartConfig.shared
+    - name: ats # ATS reads its own entry
+      valueFiles:
+        - file1.yaml
+        - file2.yaml
+    - name: atf # apptest-framework reads its own entry
+      valueFiles:
+        - file1.yaml
+        - file2.yaml
```
QuentinBisson: I would remove this and auto detect the files here if we wanted that but did we not prefer the circle CI matrix approach instead? If needed, I would create folders and detect them over adding a static config

app-testing-contract/README.md
piontec: Good question... My idea here was to just give the test devs the visibility into what they can expect of the type of the cluster, like "can we deploy and test HA mode", which might not make sense on `kind`. That info won't come from just kubernetes cluster version.

app-testing-contract/README.md
piontec: Thanks, that was meant for `ats` only, will fix

app-testing-contract/README.md
QuentinBisson: Should this not be a capabilities then?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+We propose to keep both toolkits, but specialize them to the two most frequent use cases:
+
+- **app-test-suite (ATS)** - should provide rapid local feedback on PRs and in local dev environments. Its
+  goal is to deploy the helm chart under tests as fast as possible and start testing the chart's
+  functionality. This means that it will sacrifice all the possible real cluster features to achieve this
+  goal. The cluster environment it is meant to run is `kind`, although it doesn't make any assumption about
+  the cluster type.
+- **apptest-framework (ATF)** takes the opposite approach: it chooses environment realism over the time needed
+  to execute the tests. It creates a real workload cluster, installs the chart using the App Platform, and
+  runs the full suite, preferably nightly.
```
mcharriere: we don't currently run nightly E2E for apps. Are you implying here switching from on-demand to nightly runs?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+  both toolkits, with the same test code.
+
+### Test Development Convention
+
+#### Assumptions
+
+To avoid forcing a specific test framework or technology on test authors, we decided to make a convention that
+tests are executed as a separate process by the test toolkits. The only requirements are that:
+
+- for each test suite, we group all the tests into:
```
mcharriere, line 76: I don't see how this maps with atf model:

- We can run multiple suites, each on a different WC in parallel
- A suite can be standard or upgrade
- Each suite is run per provider (or the provided defined for that suite)

Could you provide an example?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+  - **functional**: tests that check the actual functionality of the app, and that it behaves as expected.
+    They should be more complex and cover more scenarios than smoke tests.
+  - **upgrade**: tests that check if the app can be upgraded from a previous version, by default the last
+    stable version available in the OCI registry. The tests from this groups are executed twice, once before
+    the upgrade (the "old" version, that we start the upgrade test with) and once after the upgrade (the "new"
+    version, under the test).
+- a single test can be of multiple types, for example a test that checks if the app is deployed and running
+  can be both smoke and functional; a test that checks if the main page of an app loads can (and probably
+  should) be functional and upgrade.
+- tests need to be runnable with `go test` (golang) or `pytest` (python), and are using test filtering to run
```
mcharriere, line 88: atf works with ginkgo. the pipeline assumes certain things that wouldn't be there if the tests were not structured following that framework. (e.g. printing results)

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+  only the tests of a specific type (smoke, functional, upgrade). The test filtering is done with build tags
+  (golang) or markers (python).
+- all the information about the test environmenrt is passed to the tests via environment variables, and the
```
mcharriere, line 91: atf uses several config files to define multiple aspects of the configuration. why limiting it to envvars?

also, in the case of ats, how could I configure the kind cluster where the test is executed?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+`ATS_CHART_PATH` and `ATS_TEST_DIR` have no `APP_TEST_` equivalent; the contract doesn't expose them and they
+stay ATS-only. `APP_TEST_CAPABILITIES` is new, computed by the runner, with no `ATS_` predecessor.
+`KUBECONFIG` stays as-is; it's the standard name, not ours.
+
+#### Capabilities
+
+The cluster intended for PR testing with `ats` is `kind`. As a simple cluster instance, it lacks features like
+cloud identity, real storage, or load balancers. On the other hand, a "full" cluster, like created by `atf`,
+will probably be able to deliver some of these capabilities. To cope with these differences, tests that need
+those capabilities should gate on the `APP_TEST_CAPABILITIES` and only run nightly on a real workload cluster.
```
mcharriere: running E2E nightly for apps is worthless and a waste of money considering that:

- apps can have a quite dynamic release cycle
- often, on stable apps, a new PR means a new release the same day that the PR is merge (more so with automated releases). So testing nightly means testing after the release.
- an app can pass months without a new release, so testing every day is just burning $$

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+chartConfig:
+  shared: # optional configuration files, applied by the toolkit in the list order
+    valueFiles:
+      - file1.yaml
+      - file2.yaml
+```
+
+# TODO: just a config proopsal, discuss
+
+## Alternatives considered
```
mcharriere, line 290: Have you consider ditching `ats` and adding a `kind` or `vcluster` provider in `atf`? TBH adopting the same framework is way better than trying to make them match the contract.

mcharriere: I'd like to enumerate some advantages:

- atf is more flexible
- test metrics are unified
- results checks are uniform
- everything runs on the same infrastructure, debugging gets easier
- integration out of the box with agent analysis tooling that tenet is working on

app-testing-contract/README.md
QuentinBisson: Correct, and this shouldn't be in the contract at all. Scheduling is pipeline policy, not part of the test convention

QuentinBisson:
```suggestion
  runs the full suite.
```

app-testing-contract/README.md
QuentinBisson: I agree on cost, mostly. Apps release on merge, so a nightly run often tests an already-released artifact, and dormant apps would burn a WC per suite per provider per night. The one thing on-demand never catches is environment drift: 
- a suite that passed in March silently rotting against new k8s/provider versions with zero app commits. That argues for something scheduled but cheap, like weekly, or scheduled-only-if-not-run-since-last-release. Either way it's a pipeline decision per team, and should not be in the contract.

QuentinBisson:
```suggestion
those capabilities should gate on the `APP_TEST_CAPABILITIES`.
```

app-testing-contract/README.md
QuentinBisson: True today, but it's a small entrypoint change, not an architectural one. Two things make it cheap:
1. Ginkgo specs run under plain go test, so existing suites don't get rewritten. The contract explicitly allows Ginkgo (see Alternatives), it just stops requiring it.
2. The reporting path is JUnit XML, which isn't Ginkgo-specific: gotestsum --junitfile produces it from any go test run, and pytest emits it natively (--junitxml, ATS already relies on this). display-test-results keeps working unchanged.

app-testing-contract/README.md
QuentinBisson: atf's install path assumes App Platform or Flux HelmReleases and an MC. Local PR testing has no MC. A kind provider fixes provisioning but you'd still have to make the install layer optional/pluggable and strip the MC dependency from the core, at which point you've rewritten ATS inside atf so we are trying to not to that at all.

mcharriere: > path assumes App Platform or Flux HelmReleases and an MC

this is an implementation detail, in the test you don't need to know about the MC nor the deployment process.

> Local PR testing has no MC

true, but from the PoV of a test, you don't need to know about the MC.

> you'd still have to make the install layer optional/pluggable and strip the MC dependency from the core

why? this RFC claims that the goal is to gain speed, so being the MC where the cluster is running is not relevant.

> at which point you've rewritten ATS inside atf so we are trying to not to that at all.

why? keeping 2 separated toolkits or frameworks makes every new feature and improvement cost twice.

app-testing-contract/README.md
mcharriere: > The one thing on-demand never catches is environment drift

that's covered in the e2e cluster tests, which do run on daily basis with multiple scenarios and setups. It's true that that's duplicated code, but 1. it can be improved and 2. it's hard to escape that

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+test-facing under one name to begin with:
+
+| Legacy                                                            | Canonical                     | Note                                                                                                                                                                 |
+| ----------------------------------------------------------------- | ----------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ATS_TEST_TYPE`                                                   | `APP_TEST_TYPE`               | straight rename                                                                                                                                                      |
+| `ATS_APP_CONFIG_FILE_PATH`                                        | `APP_TEST_VALUES_FILE`        | straight rename                                                                                                                                                      |
```
uvegla, line 216: Value files in the name might be confusing. Does it mean the config file below - that can also contain a list of Helm value files - or just the Helm value file? There actually can be more than one of the latter, so I kind of assume it the config file.

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+| `KUBECONFIG`                                                    | yes               | kubeconfig of the cluster the app is deployed on (never the MC)                                                                                                                                                                               |
+| `APP_TEST_TYPE`                                                 | yes               | the type currently being run (`smoke`, `functional`, `upgrade`)                                                                                                                                                                               |
+| `APP_TEST_TOOLKIT`                                              | yes               | The name of the toolkit running the tests (`ats` or `atf`)                                                                                                                                                                                    |
```
uvegla, line 198: I get it that it can be useful to push this down, but assuming the tests should be able to run regardless of environment - sure, Helm values and App Test config files can make some differences - I mostly see this as an abuse surface to make the tests different based on where they run. Not having this could make it easier to enforce the test really run anywhere. Wdyt?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### Full Test Flow
+
+The test suite is always invoked by the test toolkit (`ats` or `atf`). The toolkit runs the tests using this
+flow:
+
+1. Test toolkit detects that tests are present in `tests/app/`.
+1. Test toolkit installs software dependencies for the test suite (go or python) using the lockfile in
+   `tests/app/`.
+1. Test toolkit prepares the cluster used for testing (installs tools or dependencies it needs to execute
```
uvegla, line 166: If possible, lets do it in a way that the tools are "installed" standalone to avoid polluting local machines especially and that the frameworks can easily use them instead of global ones.

One use case I have in mind is `konfigure-operator` that relies on Flux - well actually only `source-controller` - to fetch the source (tho it is extensible). So we need a way to install complex pre requisites and some of these, e.g. Flux I might not want to reinstall between runs, this this is more for the setup part. What I mean here is that now ATS installs Flux as an App CR. In these frameworks I assume its rather Helm directly that should do, so we might need a binary for that - or use a pure python (dunno if it exists) or the go client directly.

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+A missing hook is ignored. A non-zero exit fails the test run.
+
+#### Full Test Flow
+
+The test suite is always invoked by the test toolkit (`ats` or `atf`). The toolkit runs the tests using this
+flow:
+
+1. Test toolkit detects that tests are present in `tests/app/`.
+1. Test toolkit installs software dependencies for the test suite (go or python) using the lockfile in
```
uvegla, line 164: With `python` sometimes you need certain libs, binaries, header files or even a compiler to install certain packages. Might make sense to have a hook before this too. Problem can be especially local runs (ATS), where the mode of getting such dependencies greatly vary between the host OS.

app-testing-contract/README.md
QuentinBisson: > >path assumes App Platform or Flux HelmReleases and an MC

> this is an implementation detail, in the test you don't need to know about the MC nor the deployment process.

This is not true. For observability, we inject the installation name so we need to know about the MC but the main issue with the MC is that this is an invariant of atf and app testing actually boots up a full MC plus deploys apps which is just unbearably slow for testing a helm chart change.

On top of that, we want to support installng our charts (support only models) on non giant swarm cluster and this is not possible with ATF, because MCs is actually needed there

> >you'd still have to make the install layer optional/pluggable and strip the MC dependency from the core

> why? this RFC claims that the goal is to gain speed, so being the MC where the cluster is running is not relevant.

Because the market is mostly using Argo and in Bumblebee we need to support both flux and argo.

> > at which point you've rewritten ATS inside atf so we are trying to not to that at all.

> why? keeping 2 separated toolkits or frameworks makes every new feature and improvement cost twice.

Not specifically as long as both respect the contract. ATS is more generic/flexible in nature that ATF with enforces a giantswarm MC and a giantswarm cluster where kind and so on is not possible at all. Also, you would lose python support which some teams like to have and that would required quite a big migration effort

mcharriere: > This is not true. For observability, we inject the installation name so we need to know about the MC but the main issue with the MC is that this is an invariant of atf and app testing actually boots up a full MC plus deploys apps which is just unbearably slow for testing a helm chart change.

> On top of that, we want to support installng our charts (support only models) on non giant swarm cluster and this is not possible with ATF, because MCs is actually needed there

So, if I read this correctly you have 2 use-cases:
1. Testing mimir/loki as an integrated part of our platform.
2. Testing mimir/loki as a standalone app.

1 is only covered by the `atf` with the long test: creating the ephemeral MC and a WC. This is not solved by `ats`.
2 is what you expect to solve with `ats`. That's also possible in `atf`, as you can install the app in an WC. We do that for envoy-gateway and works fine. You just need design the test that way, which is the default use-case for all apps. This is what I say that it could be speed up adding a kind/vcluster to `atf`.

> Because the market is mostly using Argo and in Bumblebee we need to support both flux and argo.

I don't follow what the deployment method has to do with this RFC.

> Not specifically as long as both respect the contract. 

Contract has nothing to do with features that happen under the hood. Metrics or agent based reports/analysis are custom for each toolkit. If you want to have them in both, you need to double the implementation.

> ATS is more generic/flexible in nature that ATF with enforces a giantswarm MC and a giantswarm cluster where kind and so on is not possible at all. 

in my exp, ats is not more flexible. It has the limitations of running on CircleCI. Again, in `atf`, from the test perspective you don't need to know that the WC you are working with is a giantswarm one (yeah sure, you'll need to comply with some policies, but these are good practices anyway)

> Also, you would lose python support which some teams like to have and that would required quite a big migration effort

I can't comment on team preferences, but from my exp the migration effort is not that big.

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+through environment variables, the same as tests. The hooks are:
+
+| Hook                       | Runs                                                                                               |
+| -------------------------- | -------------------------------------------------------------------------------------------------- |
+| `tests/app/hooks/setup`    | after the cluster is ready, before the app is deployed (for example: install prerequisites, CRDs)  |
+| `tests/app/hooks/pre-run`  | Run before every test suite invocation, for each detected test types                               |
+| `tests/app/hooks/post-run` | Run after every test suite invocation, for each detected test types                                |
+| `tests/app/hooks/teardown` | after all tests, before the harness tears anything down (for example: clean up external resources) |
+
+A missing hook is ignored. A non-zero exit fails the test run.
```
stone-z, line 156: Does a non-zero exit code in `setup` or `pre-run` skip the test invocation altogether?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+- exit code `0` means the test run passed and at least 1 test was executed, exit code `5` means there was no
+  error, but no test was executed at all, and any other non-zero exit code means the test run failed.
+- test developer can deliver hooks that are executed by the test toolkits (see below).
+
+#### The conventional directory
+
+Tests live in one directory: `tests/app/`. Any harness that deploys the app runs that directory the same way.
+An empty `tests/app/` (no module, no project) isn't an opt-in and is skipped.
+
+It's either one Go module or one Python project, not both. The runner picks the executor from what's there:
```
stone-z, line 103: Does this mean we can't have some tests written in Go and some in Python for a given project?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+The test suite is always invoked by the test toolkit (`ats` or `atf`). The toolkit runs the tests using this
+flow:
+
+1. Test toolkit detects that tests are present in `tests/app/`.
+1. Test toolkit installs software dependencies for the test suite (go or python) using the lockfile in
+   `tests/app/`.
+1. Test toolkit prepares the cluster used for testing (installs tools or dependencies it needs to execute
+   tests; might be a no-op, depends on the test toolkit).
+1. If present, the `setup` hook runs after the cluster is ready, before the app is deployed.
+1. The app is deployed using the passed helm chart and the installation is settled (chart install exists
```
stone-z, line 169: Does this contract also specify and guarantee any consideration for nested charts?

For example, the `cluster` chart and bundles are charts that contain HelmReleases -- to what degree are toolkits adopting this contract expected to support the nested resources?

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+   cleanly).
+1. For each test `type` in `smoke`, `functional`:
+   1. `pre-run` hook runs for `type` tests (if present).
+   1. Tests are executed for `type` type, using either `go test` or `pytest`, depending on the detected module
+      type.
+   1. `post-run` hook runs for `type` tests (if present).
+1. If there are `upgrade` tests:
+   1. If the version under test (new) is already installed in the cluster, it is uninstalled.
```
stone-z, line 177: This makes the test non-representative. The cluster must be created using the previous version for upgrade tests.

We commonly need to hand-fix cases where CRDs have changed and resources didn't upgrade cleanly for various reasons. Applying the new version of the application first guarantees that the new CRDs are present when the cluster and resources are created, which is not representative of a real update.

stone-z, line 177: Additional question -
I understand "upgrade" in the `ats` sense to mean "installing a new version of the app, within a cluster which remains at a fixed version". 

How should I understand "upgrade" in an `atf` sense?

Because some of our apps are release defaults and some not, I would like an "upgrade" option that supports "install version 1 of the app into cluster version X, then upgrade the cluster to version Y"

app-testing-contract/README.md
stone-z: I also found this name confusing

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+`KUBECONFIG` stays as-is; it's the standard name, not ours.
+
+#### Capabilities
+
+The cluster intended for PR testing with `ats` is `kind`. As a simple cluster instance, it lacks features like
+cloud identity, real storage, or load balancers. On the other hand, a "full" cluster, like created by `atf`,
+will probably be able to deliver some of these capabilities. To cope with these differences, tests that need
+those capabilities should gate on the `APP_TEST_CAPABILITIES`.
+
+We define the following capabilities, which a toolkit might set in `APP_TEST_CAPABILITIES`:
```
stone-z, line 231: Just a suggestion, not objecting, but I suspect it will be difficult for you to manage capabilities over time like this. People will use and abuse them in unexpected ways, and will need new ones for weird niches.

Rather than try to anticipate what the capability variants might be, an option might be making people be specific about what their test requires (e.g. by detecting a CRD, probing for a service).

Maybe there are KaaS tests where this makes more sense.

---
longer:
For cases I would like to test, capabilities seem like a way of declaring a dependency on an unspecific service, but the test probably requires (directly or indirectly) something specific.

For example, I want to test autoscaling of my app. That doesn't make sense in `kind`, but the autoscaling implementation already depends on the presence of HPA/VPA APIs. I could gate the test based on the CRDs.
Same with metrics. I would want to test that some metrics are collected from my application, but in order to do that I'd need to be able to create the Pod/ServiceMonitor (not in `kind` by default) and then I'd need to know where to get the metrics from. I could gate the test based on whether the CRDs are there and whether you passed me a metrics endpoint to hit (*).

In both cases, an `autoscaling` or `metrics` capability would just be masking what I really needed, and KaaS is going to think `autoscaling` means Karpenter and/or cluster-autoscaler anyway.

\* the metrics example also gets into MC visibility and external "fixtures" which I don't see represented here yet.

app-testing-contract/README.md
```diff
@@ -0,0 +1,307 @@
...
+- environment variables (highest priority)
+
+The reserved toolkit specific names and related overrides are:
+
+- `ats` - `.apptest/config.ats.yaml`
+- `atf` - `.apptest/config.atf.yaml`
+
+The proposed config schema is (including the default values and respective env vars for overrides):
+
+```yaml
```
stone-z, line 278: I suggest versioning the schema

stone-z (commented):
I'd argue `atf` and `ats` serve different purposes and should be decomposed instead of contractually aligned. Because we have unfortunately designed many things around MCs, there must be some awareness of that concept in e2e testing. That's 90% of what `atf` does now, and that won't go away any time soon. However, the actual test execution portion could be extracted entirely from `atf` and made to be reusable in CI. Then `atf` could invoke that thing for WC tests and again for MC tests.

I'm not opposed to trying this to make things better, I just don't think it's solving the right problem.

FWIW I echo Mati and think it makes more sense to just use one tool.

---

Do I understand correctly that the result of this is that I can write `ats` tests, and those tests will be run first in kind on CircleCI and then those same tests will also be run during e2e?

If so, why would I bother writing `atf` tests that only work in one place?

app-testing-contract/README.md
piontec: The idea is that `ats` and `atf` will converge here (and changes in both are needed): the mapping between test types is, I think, `functional-standard`, upgrade is the same, and `smoke` is empty for `atf` right now. As for "where does it run", I'm not sure what you mean: this can run wherever `atf` can start it, it's the toolkit's config and the CI/CD process that should make this decision, not how the tests are executed. Does that make sense?

app-testing-contract/README.md
piontec: It's not limited to envvars, it's "config file first, then env vars overrides". The whole point here is that it's much easier to set env vars in CI environment than it is to dynamically generate a config file.

As for ats: `ats` will no longer manager `kind` cluster lifecycle, the CI process will.

app-testing-contract/README.md
piontec: @mcharriere you touched on the very good point where this whole discussion between Quentin and me actually started. We have opted for "2 tools, 1 way to write tests" because:
- `atf` seems to be very opinionated (for a good reason) about the environment where it runs the tests: it has to be a real GS WC/MC cluster. Adding there "the minimal cluster" option seemed like adding `if (minimal) {a} else {b}` all over the place (no-go) or doing a really heavy refactoring, that will effectively backport `ats` on top of it, as one of its working modes,
- With the above in scope and having in mind that both tools are already used and tested in their current forms, we proposed the current solution as the way to skip the heavy refactoring of both tools and tests and then getting some maturity and stability out of the new tool; starting from scratch today, I would agree this should be a single tool, but the price of unifying them now seems high compared to the benefits.

This for sure isn't a perfect solution, but we preferred this over the alternative.

app-testing-contract/README.md
piontec: Naming is hard... do you have better ideas? Is `APP_TEST_CHART_VALUES_FILES` better? Yes, it's for passing values files (plural) to the installed helm chart.

app-testing-contract/README.md
piontec: I agree, the difference should the capabilities a cluster has, not the toolkit running the tests.

app-testing-contract/README.md
piontec: Scenarios like this ones will be supported by an extension of this spec that we have in mind - but I think it's not worth it to discuss more features before we even approve the general idea. But I totally see your problem and we have already discussed it.

app-testing-contract/README.md
piontec: I think this should be provided by the env, like during the CI setup process. Otherwise, we need a way for the toolkits to install dependencies, potentially system-wide, which means `root` access is needed. And we still probably need to make some assumptions about where the tool is run, because that will depend on the distro and such. CI should have this specific information and can install what's needed without enforcing the same approach on other  CI environments.

app-testing-contract/README.md
piontec: IMHO if the setup fails, the test run is inconclusive, so should fail at that moment.

app-testing-contract/README.md
uvegla: Makes sense. Just maybe it should be stated that double book keeping is required for CI runner and for whatever runs ATF (+ local). It might be an implantation detail how exactly, but a significant point and worth a bit of thoughts how this will be kept up to date.

Its also true however that the specific repos tests might need these so the repos set them up. But should / is there / can we make an interface for this? I am not sure we can, but then for each repo the maintainer must maintain it 3 times (local, ATS, ATF setup) and the tools must be able to call them in a specific manner, no?

I mean for local, devs can call it. For CI devs can put it into the CIrcleCI config maybe (tho then it could be better unified)? For ATS I dont know, depends on how it is ran.

app-testing-contract/README.md
uvegla: +1, good idea

piontec:
> Do I understand correctly that the result of this is that I can write `ats` tests, and those tests will be run first in kind on CircleCI and then those same tests will also be run during e2e?
> 
> If so, why would I bother writing `atf` tests that only work in one place?

The goal here is "write tests once, run in both". There will be no more "atf tests" and "ats tests", as both toolkits will be able to execute them (after applying the contract).

As for making this a single tool, that runs different clusters depending on the requested environment: totally possible, but the idea here is about saving us a major rewrite, and saving us time.

app-testing-contract/README.md
piontec: Yeah, currently we assumed it doesn't make sense to combine python and go tests in a single project, as that makes code sharing between test scenarios impossible. Do you have a use case for that?

app-testing-contract/README.md
piontec: We have an extension in mind that, in short, will allow test developers to provide requirements about what capabilities the cluster-under-test needs to provide, so for example: you need to test a bundle that uses flux resources inside, so you configure your test to request `flux` capabilities. Toolkits will ensure that it's available before your test starts.

app-testing-contract/README.md
piontec: Good point. Still, cluster lifecycle management is on purpose out of scope here. This should be solved in CI, with a flow like:

- create test cluster
- run only smoke anddd functional tests
- recreate test cluster from scratch
- run only the upgrade test

I think this is a common scenrio and as such it should be available and implemented in our CI/CD.

piontec: Again, a pefectly valid scenario, which we find our of scope here, as we try to keep cluster lifecycle separate. On a high level, this can again be a CI pipeline with high level flow like:
- deploy cluster v.X
- start tests with a test toolkit
- upgrade cluster to v.X+1
- start the same tests again using the toolkit

app-testing-contract/README.md
piontec: I see your point, but I'm not sure if gating on CRDs is really better. Let's say I have a chart that is using flux CRs internally: `OCIRepository`, `HelmRelease` and `Kustomization`. Should I check and gate on all 3 of them? What if I need 5? Or should I check just one and assume the rest is available? I would prefer to check that "capabilities" has "flux". But I can also imagine that I might be testing with a dependency that introduced a new CR, like `MutatingPolicy` replacing `ClusterPolicy` in kyverno - in that case, I need to know the cluster has "kyverno" capability with a sufficiently new version - so capabilities also need versions, I think. 
One important note: I think that if the "capabilities" are not specific enough for you, you can just ignore them and fetch the real list of CRDs using just the k8s API. This will give you all the details possible and won't be harder than getting the same list from a toolkit while at the same time not imposing a new feature on the contract.

piontec:
As a general comment: thanks for all the detailed comments, they will become very useful if we accept the overall idea. But the core question right now is "do we find this ok, as it is a compromise of some sort"? `ats` is really close to implementing this contract already and that will unblock team-bumblebee with their testing requirements. For `atf` to catch up, it will definitely take some more time, but that's part of the tradeoff.

app-testing-contract/README.md
stone-z: Well, you're including the hooks and test flow as part of the RFC. So if it's something to be addressed, I think it should be, or if the test sequencing is something that is planned to be flexible and managed separately, then maybe it should be removed?

stone-z: Same comment here - the RFC describes the hooks that are possible, so since cluster lifecycle tests are relevant, it would be good to commit to some support for them in the RFC.

Or, if the RFC is intended to still be higher-level, then the implementation details don't belong here.

app-testing-contract/README.md
stone-z: My point is only the test probably depends on something specific, and people are not going to use these how you expect, so it'll become more headache for you to manage.

Versioning is a good idea in any case if you choose to do capabilities

app-testing-contract/README.md
stone-z: I was just clarifying. I don't necessarily want to have both. I'd prefer to write tests in Go, but it's more important to me that I can run tests everywhere, so if that means Python I'll do that.

Not supporting both in parallel just reinforces my top-level comment.

stone-z (commented):
> The goal here is "write tests once, run in both". There will be no more "atf tests" and "ats tests", as both toolkits will be able to execute them (after applying the contract).

Ok I think I get it now. So, `ats` will be updated to also run Go tests, and `atf` will be updated to also run Python tests.

But only `ats` will be able to run in kind / CircleCI.

So HB and Tenet still need to maintain, and now keep feature parity in, two tools in parallel.

If `ats` can run everywhere, and can now run Go tests, what is the upside of duplicating that in `atf`?

> As for making this a single tool, that runs different clusters depending on the requested environment: totally possible, but the idea here is about saving us a major rewrite, and saving us time.

Is updating two tools with new language support inherently any faster than updating one tool (`atf`) to call another existing tool (`ats`).

---

`atf` and `ats` still seem like different things to me, but I think it's ultimately up to the two teams if they would prefer to maintain parallel tools.

The scope of the RFC is still unclear based on comments. If this is going to be the definitive contract decision, then I would like to see more specific consideration for nested charts and cluster lifecycle tests, and I'd have more input there. If you want to keep this a high-level "we want HB and Tenet to make our tools do the same things, and we will get back to you with a more specific RFC about what you need from the contract" then I'm perfectly happy if you drop the specific API and implementation discussion from this RFC and continue with that plan.

othylmann:
Stepping in as a stakeholder rather than a reviewer, because I think this thread has two arguments tangled into one and that's why it isn't converging.

**Argument 1: how do we author app tests?** One directory, three test types, env-var inputs, JUnit output, tests never provision. Reading back through the thread, nobody has actually objected to any of this. Mati and Zach are raising valid detail questions (capabilities vs. CRD probing, cluster-lifecycle upgrades, schema versioning) but not disputing the authoring model itself.

**Argument 2: how many runners execute it?** This is the real disagreement, and it's an architectural call about `ats` and `atf` that four people are not going to settle in review comments.

The important thing is that these are independent. The contract does not commit us to keeping two runners forever. If anything it's what makes collapsing to one runner affordable later: tests written against the contract survive that migration unchanged, whereas tests written against a harness API do not. Mati and Zach, if your position is that we should end up on one tool, the contract is on your side — it decouples the test corpus from the decision, so we can make it on its own merits and on its own timeline instead of holding app testing hostage to it.

So my proposal:

1. **Narrow this RFC to the authoring contract** and merge it. Zach already offered this: keep it high-level, drop the API and implementation specifics. Concretely that means pulling out `APP_TEST_CAPABILITIES`, the config schema, hook mechanics, and cluster-lifecycle sequencing into follow-ups. Zach's point about capabilities is right — a curated list will be used in ways we didn't anticipate and will rot; that needs its own design pass, not a paragraph here.
2. **Take the runner-count question to a call** with Honeybadger and Tenet, with actual numbers on both sides: what the atf refactor costs, what Bumblebee is blocked on today, and what feature parity across two tools costs per year. Lech, the cost asymmetry may well be real, but right now it's asserted rather than sized, and that's the thing Mati and Zach are pushing back on.

One thing worth saying out loud: we have two overlapping test tools, owned by two teams, with close to zero adoption of either. That's the underlying problem. A contract is a reasonable way to stop the bleeding while we decide, but it isn't the fix, and I don't want merging it to feel like the question is closed.

mcharriere:
> The important thing is that these are independent. The contract does not commit us to keeping two runners forever. If anything it's what makes collapsing to one runner affordable later: tests written against the contract survive that migration unchanged, whereas tests written against a harness API do not. Mati and Zach, if your position is that we should end up on one tool, the contract is on your side — it decouples the test corpus from the decision, so we can make it on its own merits and on its own timeline instead of holding app testing hostage to it.

This is not 100% true, because the current model doesn't cover all the `atf` capabilities. The contract as described here is based on `ats` and limited to its features and model. This means that we can't do what we are currently doing on our (Cabbage) e2e tests with this contract. Plus, the RFC stated earlier than advanced use-cases are out of scope and should be written following the current e2e contract.

With that, I see that we (cabbage) will never migrate to the proposed contract and rely on skipping the ATS validation at all, just to avoid maintaining 2 set of tests. What this is trying to solve is causing the completely contrary effect.

JosephSalisbury:
from sig arch - @mcharriere plz share more thoughts <3

mcharriere:
I'll try to wrap up my feedback in a few points here:

1. `ats` and `atf` have different execution models. I don't think this is a blocker, but we should understand that and map them correctly. If the solution is changing `atf` I'm fine, but we should not loose features.
2. I don't see why maintaining 2 frameworks is an valid option in the long run.
    - if it's not the idea, why moving `atf` testing model to make it compatible with `ats` which would be deprecated afterwards 
    - nstead, why we don't just use `atf` with its current model and contribute to make it faster and solve the problems that it causes.
3. Replacing atf's current go based hook system (integrated in the same pipeline) with bash scripts is a downgrade
4. Splitting the cluster management from the test itself is a good idea that could benefit `atf`, sure. 

And lastly, I don't want to be a blocker: if this is solving your needs go ahead. My concerns are quasi hypotheticals because we won't really see the RFC impact until implementation in `atf`

PR #154, opened 2026-07-16
feat(decision-process): record decisions surfaced during RFC review

mproffitt:
As part of an RFC review, some decisions may come to light that relate to the expected outcome of the RFC. At present these can get lost inside PR comments or Slack threads without a formal record being made.

To improve discoverability, RFCs may now carry an optional `## Decisions` section with dated `### YYYY-MM-DD Title` entries recorded before approval. The format checker validates the section when present and exports the decisions into the rendered JSON so they surface in the handbook.

Existing RFCs are unaffected; the section is optional and not back-ported.

### DIFF
--- a/decision-process/README.md
+++ b/decision-process/README.md
@@ -2,6 +2,7 @@
 creation_date: 2023-10-10
 issues:
 - https://github.com/giantswarm/giantswarm/issues/24661
+last_review_date: 2026-07-16
 owners:
 - https://github.com/orgs/giantswarm/teams/sig-architecture
 - https://github.com/orgs/giantswarm/teams/team-planeteers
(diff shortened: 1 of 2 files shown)


decision-process/README.md
```diff
@@ -110,6 +111,24 @@ Therefore, we will automatically render the merged RFCs into our [handbook](http
...
+### Recording decisions during review
+
+During review, decisions related to the content of the PR may come to light which are pertinent to the approval of the RFC. Record them at the bottom of the RFC **prior to approval** in a section that uses these headings **verbatim**, so decisions stay uniform across all RFCs and can be found and rendered automatically:
+
+```markdown
+## Decisions
+
+### YYYY-MM-DD The decision, stated in the heading
+
+Context and reasoning behind the decision.
```
piontec: Do we want the context here? I think the context should be already given earlier, to start the Comments process, and here we want to have the gist of decisions taken. WDYT?

piontec (commented):
Should we make the Decisions section mandatory? I think we should, as from each RFC there should come some decisions, even when they are of negative type, like "Decision: do not add orange juice to clusters".

By mandatory, I mean validate in the process automatio that it exists.

decision-process/README.md
mproffitt: I think this is just a change of word from `Context` to `Description`

mproffitt:
I left the Decisions block optional because it will not apply to every RFC. Some may be accepted without additional decisions whilst others may have a plethora of them. I think this is the safest way to ensure compatibility across all RFC types.

lyind:
Would it be more clear to write something along the lines of:

> The decision block is optional if the only relevant decision, within the context of the RFC process, is approve or close the PR. As soon as there are turns beyond that, the decision block must be added and list those.

That way it's clear when to add the block. Just declaring it optional may leave too much to chance. (Imagine a contemporary LLM reading and acting on it.)

uvegla:
Where should decision go that come up after it was approved / merged? For example something comes up during implementation or smaller changes later? Should the original RFC be extended?

For example the semver based upgrade RFC was nicely updated / ported back 3 times after it was merged:

- original: https://github.com/giantswarm/rfc/pull/132
- extensions:
  - https://github.com/giantswarm/rfc/pull/148
  - https://github.com/giantswarm/rfc/pull/149
  - https://github.com/giantswarm/rfc/pull/150

This also leads to the question if it is still okay to extend / update the content or should it always go to the Decisions block?

JosephSalisbury:
i reckon this is good enough to get started with. i could make up examples of RFCs that may not need / want a decisions section, so i'm fine w/ optional.

with decisions made after the rfc is approved / merged, i think a follow up PR is fine, we're definitely going to have situations where we "miss" something

JosephSalisbury (approved):

PR #156, opened 2026-07-28
Cluster app values schema RFC: allow `unevaluatedProperties: false` next to `$ref` (supersedes #55)

piontec:
## Why this PR exists

This RFC was proposed in #55 and never reviewed to a conclusion — the stale bot closed it after 100 days of inactivity (2024-06-28). Meanwhile the RFC became a live reference: schemalint's `cluster-app` rule set implements it, and `pkg/lint/rulesets/rulesets.go` sets

```go
referenceURL: "https://github.com/giantswarm/rfc/pull/55", // should be updated when PR is merged
```

so the normative reference our linter points users at is a **closed** PR whose content is not on `main`. This PR reopens the discussion with the RFC brought up to date with what schemalint now actually enforces.

Content is #55's, unchanged, apart from the two commits described below. @marians is the original author.

## What changed relative to #55

Both changes cover one thing: closing an object whose properties come from a `$ref`.

**R15 — carve out the one permitted use of `unevaluatedProperties`** (6668857)

R15 previously said `unevaluatedProperties` and `unevaluatedItems` MUST NOT be used, full stop. That rule made the only correct spelling of a closed `$ref` illegal:

`additionalProperties` only considers the `properties`/`patternProperties` that are its own **siblings** ([2020-12 core §10.3.2](https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.2)) — never properties pulled in through `$ref`. So `additionalProperties: false` next to a bare `$ref` does not tighten the schema; it rejects *every* field the referenced schema defines. Real breakage, in `hello-world-app`:

```
$ helm template test helm/hello-world -f values-ext.yaml
Error: ... - at '/affinity': additional properties 'podAntiAffinity' not allowed
```

`unevaluatedProperties: false` is the only keyword in draft 2020-12 that sees `$ref`-evaluated properties, so R15 now permits exactly that one shape:

```json
"affinity": {
  "$ref": "#/$defs/affinity",
  "type": "object",
  "unevaluatedProperties": false
}
```

Every other use of `unevaluatedProperties` (notably a subschema value rather than `false`, or one with no `$ref` to close) and all use of `unevaluatedItems` remain forbidden, for the original reason: what counts as "unevaluated" depends on annotations produced anywhere else in the schema.

**R3 — point at the exception, at the right strength** (d51f824)

R3 requires `additionalProperties: false` on the root and recommends it (SHOULD) on nested objects. It now names the `$ref` exception: `additionalProperties` MUST NOT be used there, and such an object SHOULD be closed with `unevaluatedProperties: false` instead.

The MUST NOT is the normative half — that form is simply broken. The SHOULD deliberately matches the strength of the nested-object rule it excepts, and matches schemalint: permitted (MAY, per R15), and it satisfies the *recommendation*-severity "object should disable additional properties" rule. It is not required.

## Matching implementation

- **giantswarm/schemalint#331** (open) — `cluster-app` rule set. `isClosedRef` gates the exception on exactly the shape R15 describes: `$ref` present **and** `unevaluatedProperties` compiling to literal `false`. `AvoidUnevaluated` still errors on everything else; `ShouldDisableAdditionalProperties` no longer asks for `additionalProperties: false` on an object that is already closed this way.
- **giantswarm/devctl#2096** (open) — the generated pre-commit config rewrites `additionalProperties: false` → `unevaluatedProperties: false` on `$ref` nodes after schema generation, so charts land in the shape R15 permits.
- **losisin/helm-values-schema-json#317** — upstream generator bug: with `noAdditionalProperties: true` it emits `additionalProperties: false` next to every `$ref`, hence the workaround above.
- **giantswarm/giantswarm#37267** — related pre-commit convergence problem in the same pipeline.

## Follow-up on merge

Repoint schemalint's `referenceURL` from `rfc/pull/55` to the merged path on `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


piontec:
Close as not needed, the original RFC was not merged, so there are no RFCs tp update

